PR #1, opened 2015-08-13
0.1 prototype

jvehent:
This is a first iteration on the tool. It supports editing of yaml and text files. (json support may be broken).

To install the dependencies, first install `libyaml-devel`, `python-devel` and `libffi-devel` via the pkg manager, then `pip install ruamel.yaml cryptography boto3`.

You'll also need a KMS key configured in some IAM account.

@relud: r?


relud:
gave it a quick glance, and it looks good. will read more thoroughly tomorrow

jvehent:
Merging this, will continue improvements in master branch.

PR #7, opened 2015-09-24
Write directly to stdout for decypt/encrypt

relud:
instead of writing a tmpfile, write directly to /dev/stdout


jvehent:
Ha! Of Course!
:+1:

PR #10, opened 2015-09-25
Python26 Python27 and Python34 compat

Natim:
Based on #9.


Natim:
@jvehent According to your tests it should work.

Feel free to try in a real usage that it works with the three of them and add failing tests I will be able to help you.

tests/test_sops.py
```diff
@@ -1,10 +1,10 @@
 import unittest2
 import mock
+import six
 
 import sops
 
-from sys import version_info
-if version_info.major == 2:
+if six.PY2:
```
Natim: if you don't want to install six, you can also use ``sys.version_info[0] == 2``

jvehent:
Thanks! It still breaks on python3, I'll add tests to help fix the issues.

```bash
$ python3 sops/__init__.py -d example.yaml
  File "sops/__init__.py", line 623, in <module>
    main()
  File "sops/__init__.py", line 199, in main
    write_file(tree, path='/dev/stdout', filetype=otype)
  File "sops/__init__.py", line 592, in write_file
    indent=4))
TypeError: 'str' does not support the buffer interface
```

jvehent:
Almost there! I'm merging this to continue working on it in my branch.

PR #12, opened 2015-09-30
0.6: Add support for AWS IAM roles

jvehent:
r? @relud


sops/__init__.py
```diff
@@ -590,17 +595,25 @@ def encrypt_key_with_kms(key, entry):
         print("Unable to find region from ARN '%s' in entry" %
               entry['arn'], file=sys.stderr)
         return entry
```
relud: shouldn't this return None? same for line 591 above

jvehent: Indeed, I'll fix it.

sops/__init__.py
```diff
@@ -590,17 +595,25 @@ def encrypt_key_with_kms(key, entry):
...
-    kms = boto3.client('kms', region_name=region)
+    # if there are no role to assume, return the client directly
+    if not ('role' in entry):
+        return boto3.client('kms', region_name=region)
+    # otherwise, create a client using temporary tokens that assume the role
+    client = boto3.client('sts')
+    role = client.assume_role(RoleArn=entry['role'],
+                              RoleSessionName='sops@'+gethostname())
```
relud: I'd prefer to see 603 and 604 in a try block with an appropriate error message, because i feel like this one will happen fairly often.

sops/__init__.py
```diff
@@ -239,7 +231,7 @@ def detect_filetype(file):
     return 'text'
 
 
-def initialize_tree(path, itype):
+def initialize_tree(path, itype, kms_arns, pgp_fps):
```
relud: any chance I could get this to be `def initialize_tree(path, itype, kms_arns="", pgp_fps=""):` so that api calls from other modules don't have to specify those for decryption?

relud: nevermind

relud:
r+ when `return entry` is fixed (try block on 603/604 is just a preference, feel free to ignore)

sops/__init__.py
jvehent: Fixed in 146b8baea8e3af0d354cba6d515f84111216a371

sops/__init__.py
jvehent: Fixed in 146b8baea8e3af0d354cba6d515f84111216a371

sops/__init__.py
```diff
@@ -578,29 +567,58 @@ def encrypt_key_with_kms(key, entry):
...
+                                  RoleSessionName='sops@'+gethostname())
     except Exception as e:
-        print("failed to encrypt key using kms arn %s: %s, skipping it" %
-              (entry['arn'], e), file=sys.stderr)
-        return entry
-    entry['enc'] = b64encode(
-        kms_response['CiphertextBlob']).decode('utf-8')
-    entry['created_at'] = time.time()
-    return entry
+        print("Unable to switch roles: %s" % e, file=sys.stderr)
```
relud: probably want to `return None` here to prevent it from entering the next try block

jvehent: Fixed in db4346876107d87e1755212faa4d97b7be94ac3d
sorry about that :anguished:

PR #30, opened 2015-11-23
1.0

jvehent:
This PR includes a number of fixes documented in corresponding issues. All changes are backward compatible, and should be easy to review one by one.

@relud : May I please have a review ?


relud:
lgtm if the new flags have been tried and appear to work.

jvehent:
I should probably add some functional tests for that.

jvehent:
I added a functional test for `--rm-pgp` in the Makefile. That doesn't cover everything, but should help a bit. I also added a commit about changing the behavior of `-r`, so it now does inplace rotation of the data key, as opposed to requiring an edit.

jvehent:
Alright, all tests pass, and I'm done moving features around. This is a 1.0, so if @Natim or @almet want to take a look and make sure it's good enough, that'd be fantastic!

PR #33, opened 2015-12-02
Fix typo of env variable name in README

statik:


jvehent:
Woops, nice catch, and thanks for the patch!

Issue #3, opened 2015-09-15
Handle Tree operations in a separate class


jvehent:
Following @ametaireau 's comment, we should create a `Tree` class, which handles its state internally, so we don't have to expose the tree variable to the user of the object.

``` python
tree = Tree() # This loads the tree. In the constructor, do everything you need.

# Then, directly use functions or arguments of the tree. e.g.

tree.key # or tree.get_key() if it takes some time to get the key.
tree.encrypt()
tree.decrypt()
tree.write()
```

jvehent:
From #48 :
> The current implementation of SOPS is functionally based. This is good for programs that rely on transforming data. However, it's beginning to get unwieldy with all of the state passing it is doing (e.g. `digest`, `iv`, `key`, and in #47 `unencrypted_suffix`).
> 
> I suggest considering a refactor to swap out the core with something more class based so state is always located on `self`. I will create a gist to provide examples of what I mean.

twolfson:
I started re-pitching this idea in #48. Here was an example approach for the tree:

https://gist.github.com/twolfson/d00d88af4c0a68749cf3

jvehent:
To be honest, given all the issues we've had deployed sops as a client python app, I'm very tempted to rewrite it in Go and distribute statically compiled binaries. It's not just python 2.7/3 compat' that's the issue, but also dependencies on openssl for the crypto package, and others.
So, before we invest too much time rewriting a lot of the core, maybe we should evaluate the cost of rewriting in Go instead, just for the sake of the discussion.

twolfson:
Can you explain further about OpenSSL/crypto for those unfamiliar with the problem (e.g. myself)?

jvehent:
On El Capitan, Apple stopped shipping the OpenSSL headers, so to install `sops` one needs to first install the cryptography package with the following command: `env LDFLAGS="-L$(brew --prefix openssl)/lib" CFLAGS="-I$(brew --prefix openssl)/include" pip install cryptography`.
which is significantly more complex that just running `pip install sops`, but there's no way around it.

jvehent:
Note that there are several limitations of the YAML support in Go which may make it a bad fit for `sops`, namely that the YAML package doesn't preserve comments through round-trips, and that we would need to walk the tree using reflections mostly, which can be painful and slow. Maybe, for the time being, sticking to Python is a better plan.

twolfson:
[...]

PR #47, opened 2016-02-04
Added support for unencrypted values and branches

### LINKED ISSUE #24: How viable would it be to support unencrypted fields?
I am very intrigued by this project, and I would like to work it into some of the processes that I oversee.

However, I have some use cases where it would be very convenient to only have some values encrypted, while others remain plaintext, in the same YAML file. For example, I may need to store an encrypted username and password or key for an FTP site, however there may be data in the YAML file that would benefit from remaining unencrypted, such as a custom report date range, or the path to a working directory, etc. If there were a top level "no_encryption" key or something of the sort, I assume this could be accomplished rather trivially.

Is this type of support compatible with the purpose of this project?

Looking forward to your reply.

Thanks,
Legomaniac

twolfson:
In #24, we discussed potential solutions for unencrypted values and branches. This PR implements the agreed upon feature (using a key suffix to denote an unencrypted value/branch). In this PR:

- Added tests for encryption and decryption
    - We paid attention to make sure the MAC included unencrypted values to prevent their tampering
- Implemented unencrypted values/branches support on encryption/decryption
- Added CLI support for `--unencrypted-suffix`
- Added unencrypted suffix persistence by storing it under `tree['sops']['unencrypted_suffx']`
- Fixes #24 

**Notes:**

We did our best to keep code style consistent but due to the functional design we wound up passing around `unencrypted`/`unencrypted_suffx` around a lot =(

We weren't sure about when to camelCase kwargs vs snake_case kwargs so we went with the Python convention of snake_case.

### DIFF
--- a/sops/__init__.py
+++ b/sops/__init__.py
@@ -91,9 +91,13 @@ Remove this text and add your content to the file.
 
 """
 
+DEFAULT_UNENCRYPTED_SUFFIX = '_unencrypted'
+
 NOW = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
 
 INPUT_VERSION = VERSION
+
+UNENCRYPTED_SUFFIX = DEFAULT_UNENCRYPTED_SUFFIX
 
 
 def main():
@@ -158,6 +162,10 @@ def main():
     argparser.add_argument('--no-latest-check', action='store_true',
                            dest='nolatestcheck',
                            help="skip check for latest version of sops")
+    argparser.add_argument('--unencrypted-suffix', dest='unencrypted_suffix',
+                           help="override unencrypted key suffix "
+                                "(default: {default})"
+                                .format(default=DEFAULT_UNENCRYPTED_SUFFIX))
     argparser.add_argument('-v', '--version', action='version',
                            version='%(prog)s ' + str(VERSION))
 
@@ -201,6 +209,10 @@ def main():
             panic("cannot operate on non-existent file", error_code=100)
         else:
             print("%s doesn't exist, creating it." % args.file)
+
+    if args.unencrypted_suffix:
+        global UNENCRYPTED_SUFFIX
+        UNENCRYPTED_SUFFIX = args.unencrypted_suffix
 
     if args.encrypt:
         # Encrypt mode: encrypt, display and exit
@@ -360,6 +372,12 @@ def initialize_tree(path, itype, kms_arns=None, pgp_fps=None):
             INPUT_VERSION = tree['sops']['version']
         except:
             None
+        # try to set the unencrypted suffix to the one set in the file
+        try:
+            global UNENCRYPTED_SUFFIX
+            UNENCRYPTED_SUFFIX = tree['sops']['unencrypted_suffix']
+        except:
+            None
     else:
         # load a new tree using template data
         if itype == "yaml":
@@ -429,6 +447,7 @@ def verify_or_create_sops_branch(tree, kms_arns=None, pgp_fps=None):
         tree['sops']['attention'] = 'This section contains key material' + \
             ' that should only be modified with extra care. See `sops -h`.'
         tree['sops']['version'] = VERSION
+        tree['sops']['unencrypted_suffix'] = UNENCRYPTED_SUFFIX
 
     if 'kms' in tree['sops'] and isinstance(tree['sops']['kms'], list):
         # check that we have at least one ARN to work with
@@ -523,6 +542,13 @@ def update_master_keys(tree, key):
     else:
         tree['sops']['version'] = VERSION
 
+    # update unencrypted suffix if it varies
+    if 'unencrypted_suffix' in tree['sops']:
+        if tree['sops']['unencrypted_suffix'] != UNENCRYPTED_SUFFIX:
+            tree['sops']['unencrypted_suffix'] = UNENCRYPTED_SUFFIX
+    else:
+        tree['sops']['unencrypted_suffix'] = UNENCRYPTED_SUFFIX
+
     return tree
 
 
@@ -612,7 +638,7 @@ def remove_master_keys(tree, rm_kms, rm_pgp):
 
 
 def walk_and_decrypt(branch, key, aad=b'', stash=None, digest=None,
-                     isRoot=True, ignoreMac=False):
+                     isRoot=True, ignoreMac=False, unencrypted=False):
     """Walk the branch recursively and decrypt leaves."""
     if isRoot and not ignoreMac:
         digest = hashlib.sha512()
@@ -620,6 +646,7 @@ def walk_and_decrypt(branch, key, aad=b'', stash=None, digest=None,
     for k, v in branch.items():
         if k == 'sops' and isRoot:
             continue    # everything under the `sops` key stays in clear
+        unencrypted_branch = unencrypted or k.endswith(UNENCRYPTED_SUFFIX)
         nstash = dict()
         caad = aad
         if INPUT_VERSION >= 0.9:
@@ -633,15 +660,19 @@ def walk_and_decrypt(branch, key, aad=b'', stash=None, digest=None,
             nstash = stash[k]
         if isinstance(v, dict):
             branch[k] = walk_and_decrypt(v, key, aad=caad, stash=nstash,
-                                         digest=digest, isRoot=False)
+                                         digest=digest, isRoot=False,
+                                         unencrypted=unencrypted_branch)
         elif isinstance(v, list):
             branch[k] = walk_list_and_decrypt(v, key, aad=caad, stash=nstash,
-                                              digest=digest)
+                                              digest=digest,
+                                              unencrypted=unencrypted_branch)
         elif isinstance(v, ruamel.yaml.scalarstring.PreservedScalarString):
-            ev = decrypt(v, key, aad=caad, stash=nstash, digest=digest)
+            ev = decrypt(v, key, aad=caad, stash=nstash, digest=digest,
+                         unencrypted=unencrypted_branch)
             branch[k] = ruamel.yaml.scalarstring.PreservedScalarString(ev)
         else:
-            branch[k] = decrypt(v, key, aad=caad, stash=nstash, digest=digest)
+            branch[k] = decrypt(v, key, aad=caad, stash=nstash, digest=digest,
+                                unencrypted=unencrypted_branch)
 
     if isRoot and not ignoreMac:
         # compute the hash computed on values with the one stored
@@ -660,7 +691,8 @@ def walk_and_decrypt(branch, key, aad=b'', stash=None, digest=None,
     return branch
 
 
-def walk_list_and_decrypt(branch, key, aad=b'', stash=None, digest=None):
+def walk_list_and_decrypt(branch, key, aad=b'', stash=None, digest=None,
+                          unencrypted=False):
     """Walk a list contained in a branch and decrypts its values."""
     nstash = dict()
     kl = []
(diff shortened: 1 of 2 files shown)


sops/__init__.py
```diff
@@ -202,10 +207,17 @@ def main():
         else:
             print("%s doesn't exist, creating it." % args.file)
 
+    if args.unencrypted_suffix:
+        unencrypted_suffix = args.unencrypted_suffix
+    elif 'sops' in tree and 'unencrypted_suffix' in tree['sops']:
+        unencrypted_suffix = tree['sops']['unencrypted_suffix']
```
twolfson: I'm not fully certain if this is the place we want to load this value but I'm not sure of any place better

twolfson:
It looks like Travis CI caught some regressions about value types (e.g. `b'a'` vs `'a'`). I'm not sure what direction we want to take (e.g. still use `iv`, `tag,` and `type` on unencrypted values OR determine if the sounds of this is due to not using something like `six`) so I will wait for feedback.

sops/__init__.py
jvehent: I might fit better into `initialize_tree()` and modify the `UNENCRYPTED_SUFFIX` there.

sops/__init__.py
```diff
@@ -95,6 +95,8 @@
 
 INPUT_VERSION = VERSION
 
+DEFAULT_UNENCRYPTED_SUFFIX = '_unencrypted'
```
jvehent: What about just `UNENCRYPTED_SUFFIX` ? We might modify it in `initialize_tree`, so it may not always be a default.

sops/__init__.py
```diff
@@ -820,6 +840,10 @@ def encrypt(value, key, aad=b'', stash=None, digest=None):
     if digest:
         digest.update(value)
 
+    # if we don't want to encrypt, then return the value
+    if unencrypted:
+        return value
+
```
jvehent: If we know in advance we don't want to encrypt the value, it would be preferable to just not enter `encrypt()`

sops/__init__.py
```diff
@@ -202,10 +207,17 @@ def main():
...
+    elif 'sops' in tree and 'unencrypted_suffix' in tree['sops']:
+        unencrypted_suffix = tree['sops']['unencrypted_suffix']
+    else:
+        unencrypted_suffix = DEFAULT_UNENCRYPTED_SUFFIX
+
     if args.encrypt:
         # Encrypt mode: encrypt, display and exit
         key, tree = get_key(tree, need_key)
-        tree = walk_and_encrypt(tree, key)
+        tree = walk_and_encrypt(tree, key, unencrypted_suffix=unencrypted_suffix)
```
jvehent: Since we can only have one `UNENCRYPTED_SUFFIX` per invocation of sops, I think it would make the code more readable to just rely on a global variable, instead of passing the value around.

tests/test_sops.py
```diff
@@ -175,6 +211,26 @@ def test_bytes_encrypt_and_decrypt(self):
...
+            },
+            "foo": "bar"
+        }""")
+        key = os.urandom(32)
+        tree = OrderedDict()
+        with mock.patch.object(builtins, 'open', m):
+            tree = sops.load_file_into_tree('path', 'json')
+        tree['sops'] = dict()
+        crypttree = sops.walk_and_encrypt(OrderedDict(tree), key, isRoot=True)
+        assert tree['sops']['mac'].startswith("ENC[AES256_GCM,data:")
```
jvehent: would be good to assert that the unencrypted values are still in cleartext here.

sops/__init__.py
```diff
@@ -620,6 +632,7 @@ def walk_and_decrypt(branch, key, aad=b'', stash=None, digest=None,
     for k, v in branch.items():
         if k == 'sops' and isRoot:
             continue    # everything under the `sops` key stays in clear
+        v_unencrypted = unencrypted or k.endswith(unencrypted_suffix)
```
jvehent: what about `must_decrypt = not unencrypted and not k.endswith(UNENCRYPTED_SUFFIX)` ?
and then only:
```python
if must_decrypt:
    decrypt(....)
```

sops/__init__.py
```diff
@@ -670,17 +683,22 @@ def walk_list_and_decrypt(branch, key, aad=b'', stash=None, digest=None):
...
 
 
-def decrypt(value, key, aad=b'', stash=None, digest=None):
+def decrypt(value, key, aad=b'', stash=None, digest=None, unencrypted=False):
     """Return a decrypted value."""
+    if unencrypted:
+        if digest:
+            digest.update(value)
+        return value
+
```
jvehent: humf.... looks like tangling the digest calculation with `decrypt()` wasn't a great idea after all...

sops/__init__.py
```diff
@@ -740,13 +758,14 @@ def decrypt(value, key, aad=b'', stash=None, digest=None):
...
 def walk_and_encrypt(branch, key, aad=b'', stash=None,
-                     isRoot=True, digest=None):
+                     isRoot=True, digest=None, unencrypted=False, unencrypted_suffix=DEFAULT_UNENCRYPTED_SUFFIX):
     """Walk the branch recursively and encrypts its leaves."""
     if isRoot:
         digest = hashlib.sha512()
     for k, v in branch.items():
         if k == 'sops' and isRoot:
             continue    # everything under the `sops` key stays in clear
+        v_unencrypted = unencrypted or k.endswith(unencrypted_suffix)
```
jvehent: same comment as with decrypt (I find `v_unencrypted` to lack expliciteness).

sops/__init__.py
```diff
@@ -754,17 +773,18 @@ def walk_and_encrypt(branch, key, aad=b'', stash=None,
...
             branch[k] = walk_list_and_encrypt(v, key, aad=caad, stash=nstash,
-                                              digest=digest)
+                                              digest=digest, unencrypted=v_unencrypted)
         elif isinstance(v, ruamel.yaml.scalarstring.PreservedScalarString):
-            ev = encrypt(v, key, aad=caad, stash=nstash, digest=digest)
+            ev = encrypt(v, key, aad=caad, stash=nstash, digest=digest, unencrypted=v_unencrypted)
             branch[k] = ruamel.yaml.scalarstring.PreservedScalarString(ev)
         else:
-            branch[k] = encrypt(v, key, aad=caad, stash=nstash, digest=digest)
+            branch[k] = encrypt(v, key, aad=caad, stash=nstash, digest=digest, unencrypted=v_unencrypted)
```
jvehent: here the digest calculation doesn't depend on entering `encrypt()`, so let's not enter it at all.

jvehent:
> It looks like Travis CI caught some regressions about value types (e.g. b'a' vs 'a').
> I'm not sure what direction we want to take (e.g. still use iv, tag, and type on
> unencrypted values OR determine if the sounds of this is due to not using something
> like six) so I will wait for feedback.

This is happening because you enter `encrypt()` even for unencrypted values, and only return the original value after type detection and conversation happens. `encrypt` converts the value to `bytes`, which changes the `JSON` string you're expecting.
```python
def encrypt(value, key, aad=b'', stash=None, digest=None):
.......
     if not isinstance(value, bytes):
         # if not bytes, convert to bytes
         value = str(value).encode('utf-8')
```
If you don't enter `encrypt()` at all for unencrypted values, or return prior to the type convertion, the problem should go away.

jvehent:
This is great work! Thanks a lot for taking on that issue.  The code looks good overall, and modulo addressing the review comments, I think it can be merged fairly rapidly. +:100:

sops/__init__.py
twolfson: Ah, alright. I will try that out

sops/__init__.py
twolfson: But this is a `DEFAULT_` in the sense of `DEFAULT_TEXT` and `DEFAULT_JSON`, isn't it?

sops/__init__.py
twolfson: This was placed in here to handle `digest.update` since it should still be part of the MAC. Did we want to relocate `digest` computation outside of `encrypt` altogether?

We could also consider resolving this by renaming `encrypt` to something like `serialize` or `dumps`

sops/__init__.py
twolfson: That feels pretty dirty, I mentioned this in the initial PR notes but maybe a classical structure for trees would be better (e.g. save `unencrypted_suffix`, `digest`, and `iv` to `self` instead of passing them around)?

tests/test_sops.py
twolfson: The encrypt unencrypted test covers that test case already =/

sops/__init__.py
twolfson: We used the same language (i.e. `unencrypted`) for consistency to make the connection obvious to programmers.

Additionally, we need to pass `v_unencrypted` (or whatever we call it) to nested branch traversals since they need to not encrypt themselves as well (e.g. `{'a_unencrypted': {'b': 'c'}}`). As a result, performing boolean flipping will probably cause more confusion than help =/

sops/__init__.py
twolfson: We can rename this to `unencrypted_branch` if that helps with clarity

sops/__init__.py
twolfson: I don't follow, can you explain further?

twolfson:
Thanks for the quick and thorough review =) I left some counter comments. I also alluded to a classical restructure of `sops`. I will open a new issue to have that discussion with pros/cons listed.

twolfson:
To be clear, I'm waiting on responses to the comments I left before iterating on the PR again (so all updates can be done in 1 fell swoop)

sops/__init__.py
jvehent: Up to you how you want to handle it. You could keep the default global, and create another global to store the current value.

sops/__init__.py
jvehent: Agreed. To be addressed in #3 then.

sops/__init__.py
jvehent: I think a cleaner approach would be to do a first pass on the tree to calculate the digest, then do a second pass to encrypt it. It's beyond the scope of this PR though, so leave it as it is.

sops/__init__.py
jvehent: `unencrypted_branch` does seem a bit clearer

sops/__init__.py
jvehent: Nevermind. For some reason, I misread my own code. This is fine as it is.

jvehent:
Ok, we have a chicken and egg problem: If a value is unencrypted, you want to exit `encrypt()` before it gets converted to bytes. But you still want it added to the digest, which requires converting it to bytes.
`encrypt()` needs to be rewritten like this:
```python
     if not isinstance(value, bytes):
         # if not bytes, convert to bytes
         bvalue = str(value).encode('utf-8')
 
     if digest:
         digest.update(bvalue)

    if unencrypted:
        return value

    value = bvalue
```

sops/__init__.py
twolfson: I think based on how we handle `INPUT_VERSION` and our delay for #3, I will likely move to a global `UNENCRYPTED_SUFFIX` for developer sanity. I might keep a `DEFAULT_` still since it's nice to declare constants too.

sops/__init__.py
twolfson: :+1:

twolfson: I will explore shifting to a global variable

sops/__init__.py
twolfson: This comment has been resolved further down. We will rename it to `unencrypted_branch` for clarity

sops/__init__.py
twolfson: Okie doke

twolfson:
Cool, I think I'm good to revisit this again. I will take another attempt by the end of the weekend.

twolfson:
This PR has been updated with the following changes:

- Renamed `v_unencrypted` to `unencrypted_branch`
- Moved to use `DEFAULT_UNENCRYPTED_SUFFIX` and `UNENCRYPTED_SUFFIX` globals
- Relocated `sops['unencrypted_suffix']` setter to emulate behavior for `INPUT_VERSION`
    - Unfortunately, this means we had to drop an `assert tree['sops']['unencrypted_suffix']` from our tests since there are no tests for `update_master_keys` (and adding one would be verbose/out of scope) =(
- Repaired broken Python 3.4 encryption
    - We deviated a bit from the suggestions; we decided to always perform `if unencrypted: if digest` checks at the head of each `encrypt`/`decrypt` function. This will prevent accidental regressions from developers missing the `unencrypted` edge case
    - We added a `to_bytes` function to simplify reused encoding in `encrypt`/`decrypt` for `unencrypted` content

jvehent:
Code looks good and tests pass, this is good to merge!
I'll test a bit locally and push a couple more unrelated change before publishing 1.7 on pypi.

Thanks a lot for the contribution @twolfson ! +:100:

twolfson:
Awesome, thanks =D

Issue #39, opened 2016-01-29
sops not working for json array of hashes


gwilton:
test.json - array of hashes (test payload)

```
[{
	"example_key": "exaaasasasasple_value",
	"example_array": [
		"example_value1",
		"example_value2"
	],
	"example_number": 1234.56789
}]
```

When using sops for encrypting test.json get the following error.

```
$ sops test.json 
temp file created at /var/folders/k6/660mymls4wbfxxh3j933wp480000gn/T/tmp4bkTf5.json
Traceback (most recent call last):
  File "/usr/local/bin/sops", line 11, in <module>
    sys.exit(main())
  File "/Library/Python/2.7/site-packages/sops/__init__.py", line 292, in main
    restore_sops=stash['sops'])
  File "/Library/Python/2.7/site-packages/sops/__init__.py", line 411, in load_file_into_tree
    tree['sops'] = restore_sops.copy()
TypeError: list indices must be integers, not unicode
```

jvehent:
Indeed, this is not supported at all. Same with YAML arrays. The reason is sops needs a hash to store the `sops` metadata as a top-level key in the resulting document. I guess we could work around that by storing the original json data under a `data` key like we do for binary files, but that's not implemented at the time.

jvehent:
Note that one alternative would be to treat your file as binary, by using an extensions that isn't `.json` (for example `.json.sops`)

twolfson:
Another option might be to store SOPS metadata in a separate file (e.g. `.sopsrc`). It's not too practical if someone renames a file but it allows for the same structure in YAML/JSON.

jvehent:
I think it's important to keep the metadata in the same file.

jvehent:
I added documentation about json and yaml types in the readme. For now, I'll close this as WONTFIX, as it is a known limitation of sops.

twolfson:
I think this is still an error, we prob need to detect that it's not an `instanceof dict` and serialize it as data =/

jvehent:
If the user expects the data to be managed as JSON, we should transparently manage it as BINARY. I agree that it's an error, and returning a useful message to the user, maybe pointing to the doc, would be the right way to go about this.
Then the user can decide to either change the JSON format, or encrypt the file as BINARY.

twolfson:
Yea, a better error message should help (e.g. documenting `--output-type`). What are your thoughts on serializing as binary and then giving a warning instead of a hard failure?

[...]

PR #49, opened 2016-02-10
sops --input-type load exception allows tree to be initialized as list, not dict

bhourigan:
Somewhat related to the issue discussed in https://github.com/mozilla/sops/issues/39. When --input-type is anything other than json/yaml it still tries to load the input as json, which can except. If it does except in some cases `tree` can be initialized as a list instead of a dict.

```
bash-3.2$ cat test.txt
[
]
bash-3.2$ sops --input-type blob --output-type json -e test.txt
  File "/usr/local/bin/sops", line 9, in <module>
    load_entry_point('sops==1.6', 'console_scripts', 'sops')()
  File "/Library/Python/2.7/site-packages/sops/__init__.py", line 194, in main
    pgp_fps=pgp_fps)
  File "/Library/Python/2.7/site-packages/sops/__init__.py", line 353, in initialize_tree
    tree = load_file_into_tree(path, itype)
  File "/Library/Python/2.7/site-packages/sops/__init__.py", line 407, in load_file_into_tree
    tree['data'] = data
TypeError: list indices must be integers, not unicode
bash-3.2$
```

Explicitly initializing `tree` seems to fix this, but I'm not sure if it's the best way. Discussion welcome.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -525,6 +525,24 @@ With this in place, calls to `git diff` will decrypt both previous and current
 versions of the target file prior to displaying the diff. And it even works with
 git client interfaces, because they call git diff under the hood!
 
+Encrypting only parts of a file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Note: this only works on YAML and JSON files, not on BINARY files.
+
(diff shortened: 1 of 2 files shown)


bhourigan:
No problem, changed to `OrderedDict()`

PR #53, opened 2016-02-22
Output format mismatch between format of write_file() and the regex in load_file_into_tree()

bhourigan:
The non-structured input type default expects data to match this regex in `load_file_into_tree()`:

```
                valre = b'(.+)^SOPS=({.+})$'
```

However, the output format in `write_file()` does not add a \n after data, so SOPS is appended to `ENC[]` and causes an error when attempting to decrypt.

Steps to reproduce:
```
bash-3.2$ echo 'This is a textfile' > passwords.txt
bash-3.2$ sops -i -e --output-type blob passwords.txt
please wait while a data encryption key is being generated and stored securely
bash-3.2$ sops -d passwords.txt
PANIC: could not retrieve a key to encrypt/decrypt the tree
bash-3.2$ cat passwords.txt
ENC[AES256_GCM,data:<...>,type:str]SOPS={"attention": "This section contains key material that should only be modified with extra care. See `sops -h`.", "kms": <...>
```

The attached patch explicitly adds a \n after data so the output can now be parsed by `load_file_into_tree()`:

```
ENC[AES256_GCM,data:<...>,type:str]
SOPS={"attention": "This section contains key material that should only be modified with extra care. See `sops -h`.", "kms": <...>
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ with codecs.open(os.path.join(here, 'README.rst'), encoding='utf-8') as f:
 setup(
     name="sops",
     py_modules=['sops'],
-    version="1.6",
+    version="1.7",
     author="Julien Vehent",
     author_email="jvehent@mozilla.com",
     description="Secrets OPerationS (sops) is an editor of encrypted files",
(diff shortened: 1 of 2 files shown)


bhourigan:
Bad PR.

PR #58, opened 2016-02-26
Fixes #57 Add check for stdout before refencing fd

### LINKED ISSUE #57: Extracting a string from a yaml store will result in `UnboundLocalError`
for yaml file foo.yaml

```yaml
foo:
    baz: 'blah'

```
`sops -d foo.yaml --extract '["foo"]["baz"]'`

```
  File "/Library/Frameworks/Python.framework/Versions/2.7/bin/sops", line 9, in <module>
    load_entry_point('sops==1.7', 'console_scripts', 'sops')()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sops/__init__.py", line 241, in main
    write_file(tree, path=dest, filetype=otype)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sops/__init__.py", line 1106, in write_file
    fd.write(tree.encode('utf-8'))
UnboundLocalError: local variable 'fd' referenced before assignment
```

chroto:
fixes #57

### DIFF
--- a/sops/__init__.py
+++ b/sops/__init__.py
@@ -1103,7 +1103,8 @@ def write_file(tree, path=None, filetype=None):
         fd = tempfile.NamedTemporaryFile(suffix="."+filetype, delete=False)
         path = fd.name
 
-    if not isinstance(tree, dict) and not isinstance(tree, list):
+    if path != 'stdout' and not isinstance(tree, dict) and not isinstance(tree, list):
+        # Write the entire tree to file descriptor
         fd.write(tree.encode('utf-8'))
         fd.close()
         return path


marumari:
I think adding a second check for the same thing is not the right way to go.

I would instead have it look like this:

    if path:
        if path != 'stdout':
            fd = open(path, "wb")
        else:
            fd = None
    else:
        fd = tempfile.NamedTemporaryFile(suffix="."+filetype, delete=False)
        path = fd.name

    if fd and not isinstance(tree, dict) and not isinstance(tree, list):
        # Write the entire tree to file descriptor
        fd.write(tree.encode('utf-8'))
        fd.close()
        return path

jvehent:
I do indeed prefer @marumari 's solution. Could you update your PR with it? (also please make sure the tests pass, flake8 is very picky :cop: )

jvehent:
I'm closing this and will push a patch in 1.11

PR #60, opened 2016-03-16
Fix encryption of false booleans

jvehent:
Fixes an annoying bug where booleans of value `false` are lost during encryption.
We already have tests that verify data after a roundtrip of encryption/decryption, so I just added booleans in the test data
r? @relud


relud:
r+

PR #63, opened 2016-03-25
Support for configuration file

jvehent:
This PR adds two things.
First, it ensures that, when creating new files, all master keys requested are available to encrypt the file's data key. `sops` would previously ignore the keys that are requested but not accessible, which is not a deterministic behavior to the user.

Second, it implements a configuration file to be added to sops' managed directory. Similar to `.git`, `sops` will look for a file called `.sops.yaml` backward in the directory tree and use its rules to select master keys to use when creating new files. This is, for now, only used on creation of new files, so we don't pay the cost of recursively looking for the config file for any other call to sops.

@mostlygeek: we discussed this together, maybe you want to comment?

r? @relud

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -292,12 +292,11 @@ Let's take an example:
 
 * file named **something.dev.yaml** should use one set of KMS A
 * file named **something.prod.yaml** should use another set of KMS B
-* all other files use a third set of KMS C
-* All live under **mysecretrepo/something.{dev,prod}.yaml**
+* other files use a third set of KMS C
+* all live under **mysecretrepo/something.{dev,prod}.yaml**
 
-Under those circumstances, a configuration file placed at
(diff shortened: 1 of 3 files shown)


mostlygeek:
I think the design would be better if a `.sops.yaml` file contained only one set of KMS and PGP references. Simpler, no guess work or debugging a pattern match.

jvehent:
> I think the design would be better if a .sops.yaml file contained only one set of KMS and PGP references. Simpler, no guess work or debugging a pattern match.

I would have if our files where organized in directories per environments. Since we mix everything, the only way to make this feature usable is via regexes. That said, the regex is optional (and will catchall if not set), so we could reorganize in the future and just use a single rule per config file, with many config files.

mostlygeek:
R+

relud:
r+

PR #65, opened 2016-04-01
README: change OSX install documentation

oremj:
OSX installation works without fuss if installed using python setup via the method documented at http://docs.python-guide.org/en/latest/starting/install/osx/#doing-it-right


robotblake:
FWIW: I put a PR against Homebrew to add a sops formula and it got accepted.

https://github.com/Homebrew/legacy-homebrew/pull/50587

So now you can just do `brew install sops`.

oremj:
@robotblake I've added a section for brew install sops in d6aa9cc. Thanks!

jvehent:
+:100: :fireworks: :bow:

PR #73, opened 2016-05-13
Split EDITOR parameter, fixes #71

### LINKED ISSUE #71: sops fails to edit file when $EDITOR contains arguments
My $EDITOR variable is currently set to `bbedit --wait --resume`. When I try to edit a file I get the following error:

```
  File "/Users/jbuckley/Library/Python/2.7/bin/sops", line 11, in <module>
    sys.exit(main())
  File "/Users/jbuckley/Library/Python/2.7/lib/python/site-packages/sops/__init__.py", line 298, in main
    run_editor(tmppath)
  File "/Users/jbuckley/Library/Python/2.7/lib/python/site-packages/sops/__init__.py", line 1282, in run_editor
    subprocess.call([editor, path])
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

Looking at the [documentation for subprocess.call](https://docs.python.org/2/library/subprocess.html#subprocess.call) this is occurring because `shell=False` by default. I can work around this by creating a shell script that runs the bbedit command with arguments but I wonder if this should be set to `shell=True`. It's recommended to [not do this with untrusted input](https://docs.python.org/2/library/subprocess.html#frequently-used-arguments) so maybe it should actually remain False. Thoughts?

jvehent:
r? @jbuck

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ with codecs.open(os.path.join(here, 'README.rst'), encoding='utf-8') as f:
 setup(
     name="sops",
     py_modules=['sops'],
-    version="1.12",
+    version="1.13",
     author="Julien Vehent",
     author_email="jvehent@mozilla.com",
     description="Secrets OPerationS (sops) is an editor of encrypted files",
--- a/sops/__init__.py
+++ b/sops/__init__.py
@@ -39,7 +39,7 @@ else:
 if sys.version_info[0] == 3:
     raw_input = input
 
-VERSION = '1.12'
+VERSION = '1.13'
 
 DESC = """
 `sops` supports AWS KMS and PGP encryption:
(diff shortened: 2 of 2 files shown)


jbuck:
@jvehent r+

PR #75, opened 2016-06-01
go-sops: add cli framework

oremj:
@jvehent r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,5 +4,5 @@ A Go decrypter for [sops](https://github.com/mozilla/sops).
 ## Install
 `go get github.com/mozilla-services/go-sops`
 
-## Run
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -4,5 +4,5 @@ A Go decrypter for [sops](https://github.com/mozilla/sops).
 ## Install
 `go get github.com/mozilla-services/go-sops`
 
-## Run
-`go-sops <sops yaml file>`
\ No newline at end of file
+## Decrypt
+`go-sops -d <sops yaml file>`
```
jvehent: can we use `sops` instead of `go-sops` as the binary name? it shouldn't conflict with python sops if the PATH prefers it.

jvehent:
lgtm

PR #82, opened 2016-08-18
Go rewrite

autrilla:


gpgagent/gpgagent.go
```diff
@@ -0,0 +1,179 @@
...
+
+// NewConn connects to the GPG Agent as described in the
+// GPG_AGENT_INFO environment variable.
+func NewConn() (*Conn, error) {
+	sp := strings.SplitN(os.Getenv("GPG_AGENT_INFO"), ":", 3)
+	if len(sp) == 0 || len(sp[0]) == 0 {
+		return nil, ErrNoAgent
+	}
+	addr := &net.UnixAddr{Net: "unix", Name: sp[0]}
+	uc, err := net.DialUnix("unix", nil, addr)
```
oremj: These two lines can be simplified to just:
```
uc, err := net.Dial("unix", sp[0])
```

gpgagent/gpgagent.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return nil, err
+	}
+	br := bufio.NewReader(uc)
+	lineb, err := br.ReadSlice('\n')
```
oremj: Can `ReadString` be used instead here to avoid the cast.

gpgagent/gpgagent.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+	line := string(lineb)
+	if !strings.HasPrefix(line, "OK") {
+		return fmt.Errorf("gpgagent: CLEAR_PASSPHRASE returned %q", line)
+	}
+	return nil
+}
+
+func (c *Conn) GetPassphrase(pr *PassphraseRequest) (passphrase string, outerr error) {
```
oremj: The named return values should be removed here. They aren't really being used.

gpgagent/gpgagent.go
```diff
@@ -0,0 +1,179 @@
...
+			panic("gpgagent: failed to send " + cmd)
+		}
+		line, _, err := c.br.ReadLine()
+		if err != nil {
+			panic("gpgagent: failed to read " + cmd)
+		}
+		if !strings.HasPrefix(string(line), "OK") {
+			panic("gpgagent: response to " + cmd + " was " + string(line))
+		}
+	}
```
oremj: What is the purpose of `panic` here? It looks like those could just be changed to `return` and the `recover` `defer` could be removed.

decryptor/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	case "bytes":
+		return v, nil
+	case "bool":
+		return strconv.ParseBool(v)
+	default:
+		return nil, fmt.Errorf("Unknown datatype: %s", encryptedValue.datatype)
+	}
+}
+
+func Encrypt(value interface{}, key string, additionalAuthData []byte) (string, error) {
```
oremj: If the key is supposed to be `[]byte` might as well change the parameter type in the definition.

json/store.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (store *JSONStore) Load(data, key string) error {
+	err := json.Unmarshal([]byte(data), &store.Data)
+	if err != nil {
+		return fmt.Errorf("Could not unmarshal input data: %s", err)
+	}
+
+	_, err = store.WalkValue(store.Data, "", func(in interface{}, additionalAuthData string) (interface{}, error) {
+		return decryptor.Decrypt(in.(string), key, []byte(additionalAuthData))
```
oremj: Since `string` is required, it might be better to make onLeaves `func(string, string, ...)` And then do the appropriate casting before passing in `in`. Is there a reason to leave it as an `interface{}`?

oremj: Oh, I see below, it is because Encrypt does require `interface{}`

kms/keysource.go
```diff
@@ -0,0 +1,140 @@
...
+	service := kms.New(sess)
+	decrypted, err := service.Decrypt(&kms.DecryptInput{CiphertextBlob: k})
+	if err != nil {
+		return "", fmt.Errorf("Error decrypting key: %v", err)
+	}
+	return string(decrypted.Plaintext), nil
+}
+
+func (key *KMSMasterKey) NeedsRotation() bool {
+	return time.Since(key.CreationDate).Hours() > float64(24*30*6)
```
oremj: `return time.Since(key.CreationDate) > (time.Hour * 24 * 30 * 6)`

json/store.go
autrilla: Of course, we don't know if the leave is an integer, a float, or a string, so we have to use interface{}. Maybe some day unions/enums will be added to go so we can have some sort of StringOrFloatOrInteger type!

kms/keysource.go
autrilla: Fixed by 15b9c0e

Makefile
```diff
@@ -1,94 +0,0 @@
-VIRTUALENV = virtualenv
```
jvehent: We still need a Makefile. See https://github.com/mozilla-services/autograph/blob/master/Makefile for examples

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
+package aes
```
jvehent: Since this package is now AES specific, how about defining an `Encryptor` and `Decryptor` interface in the base sops package and implement it here? That would make adding other algorithms easier.

jvehent: also, add import string here

autrilla: I agree, I thought about this too, but for some reason I veered off and started working on something else.

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+
+type EncryptedValue struct {
+	data     []byte
+	iv       []byte
+	tag      []byte
+	datatype string
+}
+
+func parse(value string) (*EncryptedValue, error) {
+	re := regexp.MustCompile(`^ENC\[AES256_GCM,data:(.+),iv:(.+),tag:(.+),type:(.+)\]`)
```
jvehent: protip: move that out of the function to avoid rebuilding it all the time

```go
var encre = regexp.MustCompile(`^ENC\[AES256_GCM,data:(.+),iv:(.+),tag:(.+),type:(.+)\]`)
func parse(value string) (*EncryptedValue, error) {
```

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	matches := re.FindStringSubmatch(value)
+	if matches == nil {
+		return nil, fmt.Errorf("Input string %s does not match sops' data format", value)
+	}
+	data, err := base64.StdEncoding.DecodeString(matches[1])
+	if err != nil {
+		return nil, errors.New("Error base64-decoding data")
```
jvehent: make sure you return the actual error too: `fmt.Errorf("Error base64-decoding data: %v", err)`

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	iv, err := base64.StdEncoding.DecodeString(matches[2])
+	if err != nil {
+		return nil, errors.New("Error base64-decoding iv")
```
jvehent: same comment about returning `err`

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	tag, err := base64.StdEncoding.DecodeString(matches[3])
+	if err != nil {
+		return nil, errors.New("Error base64-decoding tag")
```
jvehent: same comment about returning `err`

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	datatype := matches[4]
```
jvehent: check `len(matches) == 5` somewhere

autrilla: Doesn't `if matches == nil` do this already?

jvehent: Yep, you're right. https://play.golang.org/p/Th2Zw23tos

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+		return "", err
+	}
+
+	gcm, err := cipher.NewGCMWithNonceSize(aes, len(encryptedValue.iv))
+	if err != nil {
+		return "", err
+	}
+
+	data := append(encryptedValue.data, encryptedValue.tag...)
+	out, err := gcm.Open(nil, encryptedValue.iv, data, additionalAuthData)
```
jvehent: would `decryptedValue` be a better name for `out` ?

jvehent: maybe `decryptedBytes`

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return "", fmt.Errorf("Could not decrypt with AES_GCM: %s", err)
+	}
+	v := string(out)
```
jvehent: s/v/decryptedValue/
we don't get charged extra for good variable names ;)

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	case "bytes":
+		return v, nil
+	case "bool":
+		return strconv.ParseBool(v)
+	default:
+		return nil, fmt.Errorf("Unknown datatype: %s", encryptedValue.datatype)
+	}
+}
+
+func Encrypt(value interface{}, key string, additionalAuthData []byte) (string, error) {
```
jvehent: I'm unsure about `key string`... what encoding should that be in? Is that a base64, an hexadecimal, direct ascii?
I'd propose avoiding all those questions and taking `key []byte` directly. The parsing/convertion would be left to the caller.

jvehent:
comments

Use golint, every exported value, constant, function must have a comment describing what it does and how to use it.

aes/decryptor.go
autrilla: Aren't strings in Go bytes?

autrilla:
This is actually a question I had. Should interface implementers document themselves, or should the interface do it? or both?

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	aes, err := cryptoaes.NewCipher([]byte(key))
```
jvehent: if `aes` is a package name, don't use it as a variable name too, it's confusing.

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return "", fmt.Errorf("Could not create AES Cipher: %s", err)
+	}
+	iv := make([]byte, 32)
```
jvehent: Make 32 a constant and use the constant here and in `NewGCMWithNonceSize`

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	_, err = rand.Read(iv)
+	if err != nil {
+		return "", fmt.Errorf("Could not generate random bytes for IV: %s", err)
+	}
+	gcm, err := cipher.NewGCMWithNonceSize(aes, len(iv))
+	if err != nil {
+		return "", fmt.Errorf("Could not create GCM: %s", err)
+	}
+	var plaintext []byte
+	var t string
```
jvehent: what's `t` ?

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+		return "", fmt.Errorf("Could not create AES Cipher: %s", err)
+	}
+	iv := make([]byte, 32)
+	_, err = rand.Read(iv)
+	if err != nil {
+		return "", fmt.Errorf("Could not generate random bytes for IV: %s", err)
+	}
+	gcm, err := cipher.NewGCMWithNonceSize(aes, len(iv))
+	if err != nil {
+		return "", fmt.Errorf("Could not create GCM: %s", err)
```
jvehent: `Could not initialize encryption ciphers AES GCM: %s`

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+		t = "float"
+		plaintext = []byte(strconv.FormatFloat(value, 'f', 9, 64))
+	case bool:
+		t = "bool"
+		plaintext = []byte(strconv.FormatBool(value))
+	default:
+		return "", fmt.Errorf("Value to encrypt has unsupported type %T", value)
+	}
+
+	out := gcm.Seal(nil, iv, plaintext, additionalAuthData)
```
jvehent: Is `out` the `EncryptedValue` ?

aes/decryptor.go
autrilla: type, but that's a keyword.

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+	ciphertext := out[:len(out)-16]
+	b64ciphertext := base64.StdEncoding.EncodeToString(ciphertext)
+	tag := out[len(out)-16:]
+	b64tag := base64.StdEncoding.EncodeToString(tag)
+	b64iv := base64.StdEncoding.EncodeToString(iv)
+	return fmt.Sprintf("ENC[AES256_GCM,data:%s,iv:%s,tag:%s,type:%s]", b64ciphertext, b64iv, b64tag, t), nil
```
jvehent: Better way to write this code:
```go
return fmt.Sprintf("ENC[AES256_GCM,data:%s,iv:%s,tag:%s,type:%s]",
        base64.StdEncoding.EncodeToString(out[:len(out)-16]),
        base64.StdEncoding.EncodeToString(iv),
        base64.StdEncoding.EncodeToString(out[len(out)-16:]),
        t), nil
```

aes/decryptor.go
```diff
@@ -0,0 +1,122 @@
...
+		plaintext = []byte(strconv.FormatFloat(value, 'f', 9, 64))
+	case bool:
+		t = "bool"
+		plaintext = []byte(strconv.FormatBool(value))
+	default:
+		return "", fmt.Errorf("Value to encrypt has unsupported type %T", value)
+	}
+
+	out := gcm.Seal(nil, iv, plaintext, additionalAuthData)
+	ciphertext := out[:len(out)-16]
```
autrilla: The 16 here should be a constant

aes/decryptor.go
jvehent: I know :) That was a subtle way to suggest a better variable name, like `valueType` for example.

aes/decryptor.go
jvehent: It is: https://golang.org/pkg/crypto/aes/ `aes.BlockSize`

aes/decryptor.go
jvehent: Yeah, a string is a read only slice of bytes, but the user of the API doesn't know that and might think this function takes a base64 string. Setting it to `[]byte` makes it clear the function expects bytes.

aes/decryptor_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"testing/quick"
+)
+
+func TestDecrypt(t *testing.T) {
+	expected := "foo"
+	key := strings.Repeat("f", 32)
```
jvehent: that's `ffffffffffffffffffffffffffffffff`, not `0xffffffffffffffffffffffffffffffff`....

aes/decryptor_test.go
```diff
@@ -0,0 +1,108 @@
...
+		d, err := Decrypt(s, key, []byte(""))
+		if err != nil {
+			return false
+		}
+		return x == d
+	}
+	if err := quick.Check(f, nil); err != nil {
+		t.Error(err)
+	}
+}
```
jvehent: This is neat. I've never used `testing/quick` before. :+1:

aes/decryptor_test.go
```diff
@@ -0,0 +1,108 @@
...
+	f := func(x string) bool {
+		if x == "" {
+			return true
+		}
+		s, err := Encrypt(x, key, []byte(""))
+		if err != nil {
+			fmt.Println(err)
+			return false
+		}
+		d, err := Decrypt(s, key, []byte(""))
```
jvehent: Why not generate some random AAD here?

aes/decryptor_test.go
```diff
@@ -0,0 +1,108 @@
...
+func TestDecryptInvalidAac(t *testing.T) {
+	message := `ENC[AES256_GCM,data:oYyi,iv:MyIDYbT718JRr11QtBkcj3Dwm4k1aCGZBVeZf0EyV8o=,tag:t5z2Z023Up0kxwCgw1gNxg==,type:str]`
+	_, err := Decrypt(message, strings.Repeat("f", 32), []byte(""))
+	if err == nil {
+		t.Errorf("Decrypting with an invalid AAC should fail")
+	}
+}
+
+func TestRoundtripString(t *testing.T) {
+	key := strings.Repeat("f", 32)
```
jvehent: if you're going to do iterations, I think you should get a random key every time.

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	for i, item := range branch {
+		if item.Key == "sops" {
+			branch = append(branch[:i], branch[i+1:]...)
+		}
+	}
+	return branch, nil
+}
+
+func (store JSONStore) Dump(tree sops.TreeBranch) (string, error) {
```
jvehent: `Dump` is inherited from Python. Gophers prefer `Marshal`.

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+	out, err := json.Marshal(tree)
+	if err != nil {
+		return "", fmt.Errorf("Error marshaling to json: %s", err)
+	}
+	return string(out), nil
```
jvehent: I think we should return `[]byte` and not `string` here, because that's what all marshalers do and it would be useful for other clients that want an iowriter.

aes/decryptor_test.go
```diff
@@ -0,0 +1,108 @@
...
+	decryption, err := Decrypt(message, key, []byte("bar:"))
+	if err != nil {
+		t.Errorf("%s", err)
+	}
+	if decryption != expected {
+		t.Errorf("Decrypt(\"%s\", \"%s\") == \"%s\", expected %s", message, key, decryption, expected)
+	}
+}
+
+func TestDecryptInvalidAac(t *testing.T) {
```
autrilla: Yep, finger slipped

aes/decryptor_test.go
autrilla: Sure, but why does this matter? I wasn't going for 0xffffffffffffffffffffffffffffffff

json/test_resources/example.json
```diff
@@ -0,0 +1,33 @@
+{
```
jvehent: Why not reuse the examples that are already at the root?

aes/decryptor_test.go
jvehent: fair enough. I still think you should treat this as an `[]byte`, not a printable string, because others who read the code may get the wrong idea and think this is a passphrase when it's really just bytes.

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (store JSONStore) LoadMetadata(in string) (sops.Metadata, error) {
+	var metadata sops.Metadata
+	data := make(map[interface{}]interface{})
+	encoded := make(map[interface{}]interface{})
+	if err := json.Unmarshal([]byte(in), &encoded); err != nil {
+		return metadata, fmt.Errorf("Error unmarshalling input json: %s", err)
+	}
+
+	sopsJSON, err := json.Marshal(encoded["sops"])
```
jvehent: You need to test for key existence before referencing it.
```go
if _, ok := encoded["sops"]; !ok {
```

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+		return "", fmt.Errorf("Error marshaling to json: %s", err)
+	}
+	return string(out), nil
+}
+
+func (store JSONStore) LoadMetadata(in string) (sops.Metadata, error) {
+	var metadata sops.Metadata
+	data := make(map[interface{}]interface{})
+	encoded := make(map[interface{}]interface{})
+	if err := json.Unmarshal([]byte(in), &encoded); err != nil {
```
jvehent: nitpicking here, but I'd advise against mixing styles between the inline `err != nil` and the `if err != nil` on its own line.

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+
+	sopsJSON, err := json.Marshal(encoded["sops"])
+	if err != nil {
+		return metadata, err
+	}
+
+	err = json.Unmarshal(sopsJSON, &data)
+	if err != nil {
+		return metadata, err
+	}
```
jvehent: I must be missing something... why don't you just assign `data = encoded["sops"]` ?

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+	metadata.MessageAuthenticationCode = data["mac"].(string)
+	lastModified, err := time.Parse(sops.DateFormat, data["lastmodified"].(string))
```
jvehent: is `sops.DateFormat` just `time.RFC3339` ?

aes/decryptor_test.go
autrilla: Fixed 3733347

aes/decryptor_test.go
autrilla: Fixed 3733347

aes/decryptor_test.go
autrilla: Fixed 3733347

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return metadata, fmt.Errorf("Could not parse last modified date: %s", err)
+	}
+	metadata.LastModified = lastModified
+	metadata.UnencryptedSuffix = data["unencrypted_suffix"].(string)
+	metadata.Version = data["version"].(string)
+	if k, ok := data["kms"].([]interface{}); ok {
```
jvehent: Honest question: is this going to fail when `kms` is not present, or when `kms` is not an `[]interface{}`, or both?

json/store.go
```diff
@@ -0,0 +1,126 @@
...
+		ks, err := store.kmsEntries(k)
+		if err == nil {
+			metadata.KeySources = append(metadata.KeySources, ks)
+		}
+
+	}
+
+	if pgp, ok := data["pgp"].([]interface{}); ok {
```
jvehent: same q? as with kms above

main/main.go
```diff
@@ -0,0 +1,298 @@
+package main
```
jvehent: You want this file under the `cmd/sops` directory, not `main`. At the moment, `go get go.mozilla.org/sops/main` would install a program called `main`.

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+			Name:   "kms, k",
+			Usage:  "comma separated list of KMS ARNs",
+			EnvVar: "SOPS_KMS_ARN",
+		},
+		cli.StringFlag{
+			Name:   "pgp, p",
+			Usage:  "comma separated list of PGP fingerprints",
+			EnvVar: "SOPS_PGP_FP",
+		},
+
```
jvehent: extra newline

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		return nil
+	}
+	app.Run(os.Args)
+}
+
+func runEditor(path string) {
+	editor := os.Getenv("EDITOR")
+	var cmd *exec.Cmd
+	if editor == "" {
```
jvehent: You should inverse the logic here: `if os.Getenv("EDITOR") != "" { prepare cmd with that value} else { prepare cmd with vim or nano }`

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+				return key, nil
+			}
+		}
+	}
+	return "", fmt.Errorf("Could not get master key")
+}
+
+func decrypt(c *cli.Context, file string, fileBytes []byte, output io.Writer) error {
+	store := store(file)
+	metadata, err := store.LoadMetadata(string(fileBytes))
```
jvehent: You do a lot of back and forth between `string` and `[]byte`. I think you should just use `[]byte` everything as a default, and convert back to `string` only when printing is needed, or if the type requires it.

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error loading file: %s", err), 3)
+	}
+	key, err := findKey(metadata.KeySources)
```
jvehent: Should this be a method on the `KeySources` type? such as `metadata.KeySources.GetDataKey()`

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return cli.NewExitError(err.Error(), 4)
+	}
+	branch, err := store.Load(string(fileBytes))
```
jvehent: do you mean `root` instead of `branch` ?

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error loading file: %s", err), 6)
+	}
+	tree := sops.Tree{Branch: branch, Metadata: metadata}
+	mac, err := tree.Decrypt(key)
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error decrypting tree: %s", err), 8)
+	}
+	if metadata.MessageAuthenticationCode != mac && !c.Bool("ignore-mac") {
```
jvehent: this `c` variable threw me off, because we also have a `cli` underneath. what's the difference between the two? any better name you could give them?

json/store.go
autrilla: Uh... Good question. That basically grew from https://github.com/mozilla/sops/blob/go-sops/decrypt.go#L12-L49 without much thought on my side. I don't see any reason why we can't just do data = encoded["sops"].(map[interface{}]interface{})

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+		return cli.NewExitError("MAC mismatch.", 9)
+	}
+	out, err := store.Dump(tree.Branch)
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error dumping file: %s", err), 7)
+	}
+	_, err = output.Write([]byte(out))
```
jvehent: did `out` ever need to not be a `[]byte` ? If not, just return it as an `[]byte` right away.

aes/decryptor.go
autrilla: What name do you suggest? `cipher` is taken already.

aes/decryptor.go
autrilla: Fixed

main/main.go
```diff
@@ -0,0 +1,298 @@
...
+		return cli.NewExitError(err.Error(), 4)
+	}
+	branch, err := store.Load(string(fileBytes))
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error loading file: %s", err), 6)
+	}
+	tree := sops.Tree{Branch: branch, Metadata: metadata}
+	mac, err := tree.Decrypt(key)
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error decrypting tree: %s", err), 8)
```
jvehent: Would be nice to define those exit codes as constants, something we didn't do in the python version
```go
const(
    ExitMacFail int = 9
```

json/store.go
autrilla: When `kms` is not present. It should just skip it instead of panicking, and maybe display a warning.

aes/decryptor.go
jvehent: `aescipher` ?

kms/keysource.go
```diff
@@ -0,0 +1,143 @@
...
+	return sess, nil
+}
+
+func (k KMSMasterKey) ToMap() map[string]string {
+	out := make(map[string]string)
+	out["arn"] = k.Arn
+	if k.Role != "" {
+		out["role"] = k.Role
+	}
+	out["created_at"] = k.CreationDate.Format("2006-01-02T15:04:05Z")
```
jvehent: I think a better way to write this would be `k.CreationDate.UTC().Format(time.RFC3339)`

pgp/keysource.go
```diff
@@ -0,0 +1,202 @@
...
+	"golang.org/x/crypto/openpgp/armor"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path"
+	"strings"
+	"time"
+)
+
+const DateFormat = "2006-01-02T15:04:05Z"
```
jvehent: use `time.RFC3339`

pgp/keysource.go
```diff
@@ -0,0 +1,202 @@
...
+
+type GPGMasterKey struct {
```
jvehent: should be `PGPMasterKey`, not `GPG`
I know, it's confusing, but the standard is called OpenPGP and GnuPG is only one implementation.

pgp/keysource.go
```diff
@@ -0,0 +1,202 @@
...
+		return "", fmt.Errorf("Reading PGP message failed: %s", err)
+	}
+	if b, err := ioutil.ReadAll(md.UnverifiedBody); err == nil {
+		return string(b), nil
+	}
+	return "", fmt.Errorf("The key could not be decrypted with any of the GPG entries")
+}
+
+func (key *GPGMasterKey) NeedsRotation() bool {
+	return time.Since(key.CreationDate).Hours() > 24*30*6
```
jvehent: Another way to write this would be `time.Now().Add(-time.Hour*24*30*6).After(key.CreationDate)`
but that's not very readable...

pgp/keysource.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (key *GPGMasterKey) ToString() string {
+	return key.Fingerprint
+}
+
+func (key *GPGMasterKey) gpgHome() string {
+	dir := os.Getenv("GNUPGHOME")
+	if dir == "" {
```
jvehent:
```go
if os.Getenv("GNUPGHOME") != "" {
        return os.Getenv("GNUPGHOME")
}
usr, err := user.Current()
if err != nil {
        return "~/.gnupg"
}
return path.Join(usr.HomeDir, ".gnupg")
```

pgp/keysource.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		return path.Join(usr.HomeDir, ".gnupg")
+	}
+	return dir
+}
+
+func NewGPGMasterKeyFromFingerprint(fingerprint string) GPGMasterKey {
+	return GPGMasterKey{
+		Fingerprint:  strings.Replace(fingerprint, " ", "", -1),
+		CreationDate: time.Now(),
```
jvehent: `time.Now().UTC()`

pgp/keysource.go
```diff
@@ -0,0 +1,202 @@
...
+		return []byte(pass), nil
+	}
+	return nil, fmt.Errorf("No key to unlock")
+}
+
+func (key GPGMasterKey) ToMap() map[string]string {
+	out := make(map[string]string)
+	out["fp"] = key.Fingerprint
+	out["created_at"] = key.CreationDate.Format(DateFormat)
+	out["enc"] = key.EncryptedKey
```
jvehent: I'm putting the comment here but it applies to the entire metadata struct: why does it need to be a map? We need to treat the file data as `interface{}` because we don't know its structure in advance, but the metadata branch is deterministic. It would be much nicer to handle the metadata branch as structs instead. The `sops` package would define the base struct and `pgp` and `kms` packages would define the struct for their own keys.

sops.go
```diff
@@ -0,0 +1,259 @@
+package sops
```
jvehent: I've been thinking about the directory structure some more, and I think it would make sense to put the main package at the root and the sops package under `/sops`.
That way, people who just want to cli, which is the majority anyway, would type `go get go.mozilla.org/sops`. And people who want the lib would import `go.mozilla.org/sops/sops`.
We can discuss it if you think it's a horrible idea.

sops.go
```diff
@@ -0,0 +1,259 @@
...
+	"fmt"
+	"go.mozilla.org/sops/aes"
+	"go.mozilla.org/sops/kms"
+	"go.mozilla.org/sops/pgp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const DateFormat = "2006-01-02T15:04:05Z"
```
jvehent: You don't need to redefine this, just use `time.RFC3339` which is an international standard

pgp/keysource.go
jvehent: After writing this, I remembered you already have `Metadata` defined, so we just need a struct for each keysource then.

sops.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	return fmt.Sprintf("%X", hash.Sum(nil)), nil
+}
+
+func (tree Tree) Decrypt(key string) (string, error) {
+	hash := sha512.New()
+	_, err := tree.Branch.WalkBranch(tree.Branch, make([]string, 0), func(in interface{}, path []string) (interface{}, error) {
+		var v interface{}
+		if !strings.HasSuffix(path[len(path)-1], tree.Metadata.UnencryptedSuffix) {
+			var err error
```
jvehent: It's cleaner to predefine your variables at the top, right under the function name:
```go
func (tree Tree) Decrypt(key string) (string, error) {
var (
        v interface{}
        err error
)
```

yaml/store.go
```diff
@@ -0,0 +1,180 @@
+package yaml
+
+import (
+	"fmt"
+	"github.com/autrilla/yaml"
```
jvehent: It's much rather use the upstream one. Has your patch been accepted yet?

jvehent:
Installation failed with the following error:
```bash
$ go install go.mozilla.org/sops/main
# go.mozilla.org/sops/kms
kms/keysource.go:110: undefined: session.NewSession
kms/keysource.go:124: undefined: session.NewSession
```

json/store.go
oremj: I did this in decrypt to go, so I could convert map[interface{}]interface{} in to a struct, without manually assigning the fields. It was more of a quick hack.

autrilla:
@jvehent yes, the stuff in vendor/ is severely outdated to ease reviewing. Just remove everything under vendor/ and run godep get and it should build.

pgp/keysource.go
autrilla: Actually, I think it should just be `MasterKey`. `pgp.PGPMasterKey` is a bit redundant, but I agree in that it should have been PGPMasterKey.

sops.go
autrilla: I agree, most people will probably want the cli.

yaml/store.go
autrilla: Me too, but it hasn't been reviewed yet. https://github.com/go-yaml/yaml/pull/193

aes/decryptor.go
autrilla: Fixed e885735

pgp/keysource.go
autrilla: Fixed e4c5763

kms/keysource.go
autrilla: Fixed by 762c5cd and e4c5763

aes/decryptor.go
autrilla: Fixed

sops.go
autrilla: Fixed 762c5cd

pgp/keysource.go
autrilla: Fixed 9ffe1c4

aes/decryptor.go
autrilla: Fixed

aes/decryptor.go
autrilla: Fixed

main/main.go
autrilla: The type is `KeySource`, so I'd still have to iterate over each one and call GetDataKey.

main/main.go
autrilla: c is for context. It's what everyone who uses urfave/cli names it, by the looks of it.

sops.go
autrilla: Well, v is only used inside the closure so I don't see why it should belong to the enclosing function. wrt err, it's just a way to get over Go's limitation where you can not mix new and existing names in the same assignment.

jvehent: Bah, I missed the fact it's inside a closure. Same thing apply there though:
```go
func(in interface{}, pa...
var (
        v interface{}
        err error
)
```

main/main.go
autrilla: Fixed d5313c2

Makefile
autrilla: Fixed 49a3ef3. Generating coverage reports for multiple Go packages is a bit complicated so I'll leave that for later.

aes/decryptor.go
autrilla: Fixed 2bc7c0a

aes/decryptor.go
autrilla: Fixed 6ce4b31

json/test_resources/example.json
autrilla: I feel those may change and should probably not be used for tests. In any case I think it's better if the package owns its test resources

json/store.go
autrilla: Fixed e1f2dc1

aes/decryptor.go
autrilla: Fixed 341da25

Issue #98, opened 2016-08-27
export common errors from the sops package


jvehent:
In userplex, I had to define a few custom errors that I think would be better placed in the sops package:
```go
const (
	errNotSopsFormat string = "sops metadata not found in input data"
	errInvalidMac    string = "failed to verify the integrity of the sops file (MAC mismatch)"
)
```
Maybe we need to rethink the error codes from the sops cmd line and export an error type that has both the error string and the code, similar to [os.PathError](https://golang.org/pkg/os/#PathError).

autrilla:
MacMismatch can be found here: https://github.com/mozilla/sops/blob/go-sops/sops.go#L22

I am reluctant to exporting CLI exit codes, and in any case I think it should be the sops package that exports errors, not sops/cmd/sops. I'll expose the sops metadata missing error there too.

jvehent:
> I am reluctant to exporting CLI exit codes

I can agree with that

> it should be the sops package that exports errors, not sops/cmd/sops

Yes, that's what I meant. Sorry if it wasn't clear.

autrilla:
2437827 adds an error for when the sops branch is not found.

Issue #97, opened 2016-08-27
findKey(keysources []sops.KeySource) should be a method on sops.KeySource


jvehent:
Users of the library currently need to reimplement `findKey` in their own code. Making it part of the `sops` package as a method of KeySource would avoid this.

autrilla:
How about as a method of `Metadata`, so the user of the library doesn't have to iterate over all `KeySource`s?

jvehent:
Works for me 👍

autrilla:
Fixed in dd77627

Issue #96, opened 2016-08-27
Rename Dump() method to Marshal() and make it return []byte, not string


jvehent:
The current interface `Dump(TreeBranch) (string, error)` should be `Marshal(TreeBranch) ([]byte, error)`.
`DumpWithMetadata()` should be modified as well.

autrilla:
Fixed in 323db94

Issue #95, opened 2016-08-27
LoadMetadata() should take []byte, not string


jvehent:
`LoadMetadata(in string) (Metadata, error)` converts the `in` string back to `[]byte` right away. It should just accept `[]byte` and avoid the double conversion.

autrilla:
Fixed in 58dec6f

PR #99, opened 2016-08-29
Fixed several issues

autrilla:
#98 
#97 
#96 
#95

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -182,7 +182,7 @@ func store(path string) sops.Store {
 
 func decryptFile(store sops.Store, fileBytes []byte, ignoreMac bool) (sops.Tree, error) {
 	var tree sops.Tree
-	metadata, err := store.LoadMetadata(string(fileBytes))
+	metadata, err := store.LoadMetadata(fileBytes)
 	if err != nil {
 		return tree, cli.NewExitError(fmt.Sprintf("Error loading file: %s", err), exitCouldNotReadInputFile)
 	}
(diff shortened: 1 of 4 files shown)


cmd/sops/main.go
```diff
@@ -243,6 +252,20 @@ func decrypt(c *cli.Context, file string, fileBytes []byte, output io.Writer) er
 	return nil
 }
 
+func generateKey(tree *sops.Tree) ([]byte, error) {
```
jvehent: This should be part of the `sops` or `metadata` package.

jvehent: I think `sops.GenerateDataKey() ([]byte, error)` would be appropriate.

sops.go
```diff
@@ -137,6 +137,19 @@ func (tree Tree) Encrypt(key []byte, cipher DataKeyCipher) (string, error) {
 	return fmt.Sprintf("%X", hash.Sum(nil)), nil
 }
 
+// FindKey retrieves the data key from the first MasterKey in the Metadata's KeySources that's able to return it.
+func (metadata Metadata) FindKey() ([]byte, error) {
```
jvehent: Sorry for being pedantic on the naming here, but I think it should be renamed `GetDataKey`. Find implies we're searching for a location when we're really iterating over encrypted records, decrypting them, and returning a key.

The term `Data Key` is also important because of the distinction between master (kms/pgp) and data (file) keys.

autrilla: Agreed, fixed.

cmd/sops/main.go
autrilla: I think you meant tree.GenerateDataKey()? In any case, that's what I did. That function needs a tree (or Metadata) to save the generated key on all the keysources.

PR #100, opened 2016-08-29
Editor mode

autrilla:
Missing features preventing merge:

- [x] Keep retrying if there's invalid syntax in the edited file, until ^C is captured
- [x] Allow showing the master keys while editing if --show-master keys is passed
- [x] Don't rewrite the file if nothing is changed

### DIFF
--- a/aes/decryptor.go
+++ b/aes/decryptor.go
@@ -19,6 +19,11 @@ type encryptedValue struct {
 
 const nonceSize int = 32
 
+type stashData struct {
+	iv        []byte
+	plaintext interface{}
+}
+
 // Cipher encrypts and decrypts data keys with AES GCM 256
 type Cipher struct{}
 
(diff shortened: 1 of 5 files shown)


autrilla:
This should be mergeable @jvehent

jvehent:
@oremj Any chance you could pick this one up? I'm falling behind on reviews...

oremj:
@jvehent yeah, will do.

autrilla:
Now that I realize, I should also add the example yaml and json files for when a new file is created.

aes/decryptor.go
```diff
@@ -19,6 +19,11 @@ type encryptedValue struct {
 
 const nonceSize int = 32
 
+type stashData struct {
+	iv        []byte
+	plaintext interface{}
+}
+
```
oremj: This should be a public struct if it's being returned from a public function, `Decrypt`.

aes/decryptor.go
```diff
@@ -46,54 +51,67 @@ func parse(value string) (*encryptedValue, error) {
 	return &encryptedValue{data, iv, tag, datatype}, nil
 }
 
-// Decrypt takes a sops-format value string and a key and returns the decrypted value.
-func (c Cipher) Decrypt(value string, key []byte, additionalAuthData []byte) (interface{}, error) {
+// Decrypt takes a sops-format value string and a key and returns the decrypted value and a stash value
+func (c Cipher) Decrypt(value string, key []byte, path string) (plaintext interface{}, stash interface{}, err error) {
```
oremj: Three return values is starting to become a lot. At that point it might be better to wrap them in a struct.

It looks like `stash` is always type `stashData`, does the type need to be `interface{}`?

aes/decryptor.go
```diff
@@ -117,7 +135,7 @@ func (c Cipher) Encrypt(value interface{}, key []byte, additionalAuthData []byte
 	default:
 		return "", fmt.Errorf("Value to encrypt has unsupported type %T", value)
 	}
-	out := gcm.Seal(nil, iv, plaintext, additionalAuthData)
+	out := gcm.Seal(nil, iv, plaintext, []byte(path))
```
oremj: Can the path parameter type be `[]byte` instead, so you don't have a cast here?

cmd/sops/main.go
```diff
@@ -167,7 +171,12 @@ func runEditor(path string) {
 	} else {
 		cmd = exec.Command(editor, path)
 	}
-	cmd.Run()
+
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	return err
```
oremj: Nit pick: you this could be reduced to just `return cmd.Run()`

cmd/sops/main.go
```diff
@@ -282,16 +283,31 @@ func encrypt(c *cli.Context, file string, fileBytes []byte, output io.Writer) er
...
+	return []sops.KeySource{kmsKs, pgpKs}, nil
+}
+
+func encrypt(c *cli.Context, file string, fileBytes []byte, output io.Writer) error {
+	store := store(file)
+	branch, err := store.Unmarshal(fileBytes)
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error loading file: %s", err), exitCouldNotReadInputFile)
+	}
+	var metadata sops.Metadata
```
oremj: Should use struct initialization format here:
```
metadata := sops.Metadata{
   UnencryptedSuffix: ....
}
```

cmd/sops/main.go
```diff
@@ -332,3 +348,164 @@ func rotate(c *cli.Context, file string, fileBytes []byte, output io.Writer) err
...
+	_, err = tmpfile.Write(out)
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Could not write output file: %s", err), exitCouldNotWriteOutputFile)
+	}
+	origHash, err := hashFile(tmpfile.Name())
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Could not hash file: %s", err), exitCouldNotReadInputFile)
+	}
+	invalidSyntax := true
+	for invalidSyntax || tree.Metadata.MasterKeyCount() == 0 {
```
oremj: It looks like invalidSyntax is only false at the very bottom. This variable can be replaced with just `break` at the bottom of the loop. If MasterKeyCount == 0, should this throw some sort of error to the client? It looks like this silently skips over editing the file.

aes/decryptor.go
autrilla: Callers don't need to have access to this struct. The stash is basically an arbitrary structure each cipher can use for bookkeeping, and the sops package (or any other callers) shouldn't touch it at all. They should just save it somewhere when they decrypt, and pass it back when they want to encrypt the same value.

aes/decryptor.go
autrilla: stash is aes.stashData in this cipher, but it would be libsodium.stashData if we added another cipher

aes/decryptor.go
autrilla: Sure, but then I have to cast to []byte elsewhere. `path` comes directly from yaml and json keys, which are of type `string`.

cmd/sops/main.go
autrilla: Fixed

cmd/sops/main.go
autrilla: Fixed

cmd/sops/main.go
autrilla: Fixed

aes/decryptor.go
oremj: That's fine to leave as is then.

oremj:
This LGTM!

Issue #101, opened 2016-08-30
Working with the Qubes GPG client wrapper.


BorisKourt:
I have a Qubes OS setup that follows this GPG practice: https://www.qubes-os.org/doc/split-gpg/

The solution for cross VM communication relies on their `qubes-gpg-client-wrapper`. Which works appropriately with git and enigmail. 

Is it possible for me to configure which GPG sops should point to? I have not been able to find this in the documentation provided.

If not, is there another way for me to make these work together? So far I tested using aliases, but it didn't seem work.

Thank you!

autrilla:
In short, no. Currently sops simply calls whichever `gpg` binary it finds in `PATH`, so that will be GNUPG, most likely. Aliases won't work as sops (correctly) doesn't use a shell to run pgp. There are a few possible solutions:

- Link `qubes-gpg-client-wrapper` to some `gpg` and add the directory where that binary lives to `PATH` so that it's found before GNU's one.
- Slightly more desirable, file a PR so that instead of using `gpg`, we can use any other binary name, so that anything with `gpg`'s interface works.

BorisKourt:
Question about implementation:

I set the default GPG binary and pulled the value from an environment variable (if one exists). Now I am left with the decision of how to pass this to the functions that use GPG. 

I was about to pass through as a parameter. But that is introducing quite a bit of clutter, as I have to 'thread' it through parent functions. 

I am not seeing you use much global state so I assume that is not a good direction.

I am also seeing the `tree` structure that I could use to pass the GPG path. But I am not sure if that is part of the final text output or not. It feels like the cleanest direction, but I am 100% unfamiliar with the code-base so just want to check if that's the case.

Thanks

autrilla:
`tree` is indeed part of the final output. I would personally just have a GPG_BINARY_NAME global variable that's set to os.environ.get('GPG_BINARY_NAME', 'gpg') or something like that. There's currently no abstraction for configurations at all, so that's as good as we can get without doing major structural changes (which are coming with the new go version).

BorisKourt:
Please take a look and tell me if there is anything you need changed.

PR #102, opened 2016-08-30
Add GPG exec options for #101

BorisKourt:
This appears to work properly. Do you need me to do anything else?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -378,6 +378,19 @@ Creating a new file with the right keys is now as simple as
 Note that the configuration file is ignored when KMS or PGP parameters are
 passed on the sops command line or in environment variables.
 
+Specify a different GPG executable
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+`sops` checks for the `SOPS_GPG_EXEC` environment variable. If specified, 
+it will attempt to use the executable set there instead of the default 
+of `gpg`.
+
+Example: place the following in your `~/.bashrc`
+
+.. code:: bash
+
+	SOPS_GPG_EXEC = 'your_gpg_client_wrapper'
+	
 Important information on types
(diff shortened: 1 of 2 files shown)


BorisKourt:
Please check the current changes. 

I set `GPG_EXEC = None` to initialize it, not sure if this is a convention that works for you. 

I am new to the format of the README, but here is my edit in the branch: https://github.com/BorisKourt/sops/blob/gpg-exec-option/README.rst#specify-a-different-gpg-executable

autrilla:
That change to the readme is fine. Keep in mind that you've made GPG_EXEC a global variable, and doing `GPG_EXEC = os.environ.get('SOPS_GPG_EXEC', 'gpg')` on `main` will just create a new variable in `main`'s scope instead of altering the global variable. You either want to use the global keyword, like this:
```
global UNENCRYPTED_SUFFIX
UNENCRYPTED_SUFFIX = args.unencrypted_suffix
```
, or just set `GPG_EXEC = os.environ.get('SOPS_GPG_EXEC', 'gpg')` on the global scope instead of inside main().

BorisKourt:
Ah oops. I guess when I had the if earlier it simply passed the env check and then defaulted back to regular GPG. :/ 

I think I will go down the global route? It seems more appropriate to do the check with the other env checks.

BorisKourt:
Updated

autrilla:
Please fix the flake8 errors here https://travis-ci.org/mozilla/sops/jobs/156325475#L460 and then I can merge this.

BorisKourt:
Alright, looks like `flake8` passed!

autrilla:
Merged, thanks for the contribution!

BorisKourt:
No problem, thanks for looking it over!

PR #109, opened 2016-09-06
Added JSON example file

autrilla:


jvehent:
Silly question: could we simply has this defined once as an anonymous struct and marshal it into whatever format needed?

autrilla:
We could have a default tree, yes, but that would only work for json and yaml, we'd need to special-case binary files.

jvehent:
Idea: add a "description" key to the struct that says `Welcome to SOPS! Remove this text and add your content to the file.`. When in BINARY mode, only print the value of `description`. When in other modes, just marshal the whole struct.

Benefit: if/when we add more formats, they'll just get the example data right away.

autrilla:
I've made this change on #110 because I needed some of the changes there.

PR #112, opened 2016-09-07
Format floats and booleans like the Python version does

autrilla:
Floats should not have padding zeroes after the decimal point, and booleans should be Titlecase.


jvehent:
Nice catch!
This looks like it should have a unit test to take an input like `2.90000` and convert it correct.

jvehent:
Please add a comment to explain why this is necessary.

PR #120, opened 2016-09-10
Repaired trailing ellipses when outputting primitive values on stdout

### LINKED ISSUE #119: Outputting value from YAML file to stdout has trailing ellipsis
We are using SOPS to dump specific values in some scripts. Unfortunately, SOPS seems to be adding a `...` to the end of the output. Here are steps to reproduce:

``` bash
$ # Generate a temporary PGP backed file
$ sops tmp.yml --pgp 'XXXX'
INFO: tmp.yml doesn't exist, creating it.
INFO: generating and storing data encryption key
INFO: temp file created at /tmp/tmpJC0MRr.yaml
INFO: file written to tmp.yml
$ # Output content in plaintext
$ sops tmp.yml --decrypt
gpg: encrypted with 2048-bit RSA key, ID XXXX, created 2016-09-09
      "XXXX"
foo: bar
$ # Output value on leaf of YAML tree
$ sops tmp.yml --decrypt --extract '["foo"]'
gpg: encrypted with 2048-bit RSA key, ID XXXX, created 2016-09-09
      "XXXX"
bar
...
```

After doing some digging, it looks like this is due to us doing an `fd` check along side `isinstance(tree, list)/isinstance(tree, dict)` when we shouldn't. As a result, we are dumping a value as YAML when it's not:

https://github.com/mozilla/sops/blob/1.14/sops/__init__.py#L1281-L1285

https://github.com/mozilla/sops/blob/1.14/sops/__init__.py#L1287-L1292

twolfson:
As described in #119, when we output primitives (i.e. non-dicts/lists) to `stdout` we are receiving a trailing ellipses due to the content being treated as a YAML tree. In this PR:
- Repaired non-dict/list logic to not treat stdout much differently
- Fixes #119

### DIFF
--- a/sops/__init__.py
+++ b/sops/__init__.py
@@ -1278,10 +1278,13 @@ def write_file(tree, path=None, filetype=None):
         fd = tempfile.NamedTemporaryFile(suffix="."+filetype, delete=False)
         path = fd.name
 
-    if fd and not isinstance(tree, dict) and not isinstance(tree, list):
-        # Write the entire tree to file descriptor
-        fd.write(tree.encode('utf-8'))
-        fd.close()
+    if not isinstance(tree, dict) and not isinstance(tree, list):
+        if path == 'stdout':
+            sys.stdout.write(tree.encode('utf-8'))
+        else:
+            # Write the entire tree to file descriptor
+            fd.write(tree.encode('utf-8'))
+            fd.close()
         return path
 
     if filetype == "yaml":


twolfson:
I should note that I didn't add any tests for this as I don't recall there being any easy harness nor does it seem like it due to us writing directly to `sys.stdout`

sops/__init__.py
```diff
@@ -1278,10 +1278,13 @@ def write_file(tree, path=None, filetype=None):
         fd = tempfile.NamedTemporaryFile(suffix="."+filetype, delete=False)
         path = fd.name
 
-    if fd and not isinstance(tree, dict) and not isinstance(tree, list):
```
autrilla: Wouldn't it be simpler to just change line 1276 from `fd = None` to `fd = sys.stdout`?

twolfson: Probably but I was being consistent with the rest of this file. If we want to go the simpler route, I think it would be better suited in another PR as it might break something unexpectedly whereas this definitely only patches what is broken

jvehent:
Assigning to @autrilla for review and delegating him merging powers 😈 
The Go version behaves correctly, let's make sure the two have the same behavior.

autrilla:
@twolfson sorry for the delay. If you can rebase your change onto master, we can merge this.

twolfson:
GitHub supports rebasing when landing so you should be good without intervention from me =/

https://github.com/blog/2243-rebase-and-merge-pull-requests

autrilla:
@twolfson it does not support it when there's merge conflicts

twolfson:
There aren't any merge conflicts that I can see =/ This PR only touches `sops/__init__.py` and the only new commits on `master` are to the README:

https://github.com/mozilla/sops/commits/master

https://github.com/mozilla/sops/commit/cf78a11fc356c1f091418a791978c73ac14f8c29

My commit and its parent:

https://github.com/twolfson/sops/commit/b8ce8fb53f193d7c888151f18b084a454f380c10

https://github.com/twolfson/sops/commit/fc5a78ee83cdd2ecb46f7946fde20125d7d8154b

Maybe it's something in this repo's configuration?

jvehent:
@twolfson : Do you still want to merge this? If so, it should go into the `python-sops` branch. The new master is now in Go and solve the issue.

autrilla:
@jvehent I don't think this issue exists in the Go version.

twolfson:
@jvehent What's the support plan for `python-sops`? Will it still be maintained/regularly released? If so, then I'll redirect the branch (going to do pre-emptively). If not, then we should probably close this PR =/

jvehent:
My plan is to maintain bug and security fixes for at least a year.

twolfson:
Cool, I'm going to do a squash/rebase since it looks like we never got that working via GitHub =/

twolfson:
Alright, rebase/squash onto `python-sops` complete

PR #121, opened 2016-09-16
fix README to use the correct extract subpath flag

imsky:


jvehent:
Good catch, thanks for the patch!

PR #123, opened 2016-09-19
Fixed invalid go get in makefile

autrilla:


Makefile
```diff
@@ -10,7 +10,7 @@ GOLINT 		:= golint
 all: test vet generate install
 
 install:
-	$(GO) install github.com/mozilla-services/autograph
+	$(GO) install go.mozilla.org/sops
```
jvehent: Should be `go.mozilla.org/sops/cmd/sops`

autrilla: Fixed. Obviously it makes more sense to actually install the binary, yes!

PR #128, opened 2016-10-22
Comment support and 4 spaces per indent

autrilla:


jvehent:
Got a little vendoring problem that's breaking the travis run: `sops.go:9:2: cannot find package "go.mozilla.org/yaml.v2" in any of:`.

sops.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"go.mozilla.org/sops/kms"
 	"go.mozilla.org/sops/pgp"
+	"go.mozilla.org/yaml.v2"
```
jvehent: We don't have it namespaced under `go.mozilla.org` (and I don't think we should if the end goal is to merge upstream). The correct import path is `github.com/mozilla-services/go-yaml`, I believe.

sops.go
```diff
@@ -99,6 +100,9 @@ func (tree TreeBranch) walkValue(in interface{}, path []string, onLeaves func(in
 
 func (tree TreeBranch) walkSlice(in []interface{}, path []string, onLeaves func(in interface{}, path []string) (interface{}, error)) ([]interface{}, error) {
 	for i, v := range in {
+		if _, ok := v.(yaml.Comment); ok {
+			continue
+		}
```
jvehent: Would there be a way to handle this that doesn't break the boundary between sops and the data type? In other words: can sops be unaware of the `yaml.Comment` type?

sops.go
```diff
@@ -110,7 +114,10 @@ func (tree TreeBranch) walkSlice(in []interface{}, path []string, onLeaves func(
 
 func (tree TreeBranch) walkBranch(in TreeBranch, path []string, onLeaves func(in interface{}, path []string) (interface{}, error)) (TreeBranch, error) {
 	for i, item := range in {
-		newV, err := tree.walkValue(item.Value, append(path, item.Key), onLeaves)
+		if _, ok := item.Key.(yaml.Comment); ok {
```
jvehent: Same comment about sops being storage type agnostic.

yaml/config.go
```diff
@@ -7,7 +7,7 @@ import (
 	"path"
 	"regexp"
 
-	"github.com/mozilla-services/yaml"
+	"go.mozilla.org/yaml.v2"
```
jvehent: revert import path

yaml/store.go
```diff
@@ -2,13 +2,12 @@ package yaml
...
 import (
 	"fmt"
-	"strconv"
-	"time"
-
-	"github.com/mozilla-services/yaml"
 	"go.mozilla.org/sops"
 	"go.mozilla.org/sops/kms"
 	"go.mozilla.org/sops/pgp"
+	"go.mozilla.org/yaml.v2"
```
jvehent: revert import path

PR #129, opened 2016-10-26
GUI editors in README

autrilla:
Document that sops will wait for the editor process to exit, and that this can cause problems with some GUI editors.


jvehent:
r+ 👍

PR #132, opened 2016-10-27
Update version when editing a file

### LINKED ISSUE #131: go-sops should update document version to 2.0
Currently go-sops does not write a version to encrypted documents.

autrilla:
Fixes #131

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -41,6 +41,8 @@ const (
 	exitNoEditorFound                       int = 201
 )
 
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -41,6 +41,8 @@ const (
 	exitNoEditorFound                       int = 201
 )
 
+const version = "2.0.0"
```
jvehent: We only use 2 digits currently. let's stick to that for now.

PR #139, opened 2016-10-28
Allow decryption of empty values

### LINKED ISSUE #138: go-sops fails to decrypt JSON file with empty value
`example.json` contains an empty value that go-sops fails to decrypt with the error:
```
Could not load file: Error decrypting tree: Error walking tree: Could not decrypt value: Input string  does not match sops' data format
```

autrilla:
Fixes #138

### DIFF
--- a/aes/decryptor.go
+++ b/aes/decryptor.go
@@ -54,6 +54,9 @@ func parse(value string) (*encryptedValue, error) {
 
 // Decrypt takes a sops-format value string and a key and returns the decrypted value and a stash value
 func (c Cipher) Decrypt(value string, key []byte, path string) (plaintext interface{}, stash interface{}, err error) {
+	if value == "" {
+		return "", nil, nil
+	}
 	encryptedValue, err := parse(value)
 	if err != nil {
 		return "", nil, err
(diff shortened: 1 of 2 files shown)


aes/decryptor.go
```diff
@@ -28,7 +28,7 @@ type stashData struct {
 // Cipher encrypts and decrypts data keys with AES GCM 256
 type Cipher struct{}
 
-var encre = regexp.MustCompile(`^ENC\[AES256_GCM,data:(.+),iv:(.+),tag:(.+),type:(.+)\]`)
+var encre = regexp.MustCompile(`^ENC\[AES256_GCM,data:(.*),iv:(.+),tag:(.+),type:(.+)\]`)
```
autrilla: sops-python stores an empty value, so that's the route we have to go.

jvehent: Fair. I'm ok to merge this.

jvehent:
Looks good. Might I ask for a test? :)

autrilla:
@jvehent there's already one. Look at the change in aes/decryptor_test.go

PR #143, opened 2016-10-31
Add KMS encryption context support

autrilla:


jvehent:
> @jvehent if the format is wrong, the map returned will be nil

It doesn't look like you're testing for it though, or did I miss it? If the cmdline flag is set, but the map is nil, we should exit with an invalid format error to tell the user, not continue silently.

PR #148, opened 2016-11-01
Exit if there's errors encrypting the data key

### LINKED ISSUE #146: "Could not get master key" during rotation results in deleting the file
When running `sops -r -i <file>` without having the necessary AWS permissions, sops errors with `Could not get master key` and write an empty file to disk, effectively destroying data. sops should exit and leave the file untouched.

autrilla:
Instead of showing a warning, if there's errors encrypting any of the data keys, sops will show the errors and exit.

Fixes #146


cmd/sops/main.go
```diff
@@ -362,9 +362,9 @@ func encrypt(c *cli.Context, file string, fileBytes []byte, output io.Writer) er
 		KeySources:        ks,
 	}
 	tree := sops.Tree{Branch: branch, Metadata: metadata}
-	key, err := tree.GenerateDataKey()
-	if err != nil {
-		return cli.NewExitError(err.Error(), exitCouldNotRetrieveKey)
+	key, errs := tree.GenerateDataKey()
+	if len(errs) > 0 {
+		return cli.NewExitError(fmt.Sprintf("Error encrypting the data key with one or more master keys: %s", errs), exitCouldNotRetrieveKey)
```
jvehent: maybe `strings.Join(errs, "\n")` ?

cmd/sops/main.go
```diff
@@ -407,7 +407,11 @@ func rotate(c *cli.Context, file string, fileBytes []byte, output io.Writer) err
 	tree.Metadata.AddPGPMasterKeys(c.String("add-pgp"))
 	tree.Metadata.RemoveKMSMasterKeys(c.String("rm-kms"))
 	tree.Metadata.RemovePGPMasterKeys(c.String("rm-pgp"))
-	tree.Metadata.UpdateMasterKeys(newKey)
+	errs = tree.Metadata.UpdateMasterKeysIfNeeded(newKey)
+	fmt.Println(errs)
```
jvehent: debug leftover

sops.go
```diff
@@ -207,21 +207,13 @@ func (tree Tree) Decrypt(key []byte, cipher DataKeyCipher, stash map[string][]in
 }
 
 // GenerateDataKey generates a new random data key and encrypts it with all MasterKeys.
-func (tree Tree) GenerateDataKey() ([]byte, error) {
+func (tree Tree) GenerateDataKey() ([]byte, []error) {
```
jvehent: Another option would be to return a single error and fail on the first error. I'm not sure which is best.

PR #156, opened 2016-11-02
main.go improvements

autrilla:
Extracted input and output handling from each "command" (edit, rotate, encrypt, decrypt).
Output files are only opened if the command was executed successfully.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -48,6 +48,40 @@ const (
 
 const version = "2.0-beta"
 
+func loadPlainFile(c *cli.Context, store sops.Store, fileName string, fileBytes []byte) (tree sops.Tree, err error) {
+	branch, err := store.Unmarshal(fileBytes)
+	if err != nil {
+		return tree, cli.NewExitError(fmt.Sprintf("Error loading file: %s", err), exitCouldNotReadInputFile)
+	}
+	tree.Branch = branch
+	ks, err := getKeySources(c, fileName)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


autrilla:
As a side note, I've switched back from os.OpenFile to os.Create, because now we only open the file when we're sure we want to overwrite the contents, so truncating the file is desired.

cmd/sops/main.go
```diff
@@ -254,7 +254,7 @@ func main() {
 		if err != nil {
 			return err
 		}
-
+		// We open the file *after* the operations on the tree have been executed to avoid truncating it when there's errors
```
jvehent: I'm not against long lines, but that's a bit much :)

cmd/sops/main.go
```diff
@@ -207,32 +207,45 @@ func main() {
...
-			return err
-		}
 
 		inputStore := inputStore(c, fileName)
 		outputStore := outputStore(c, fileName)
 
-		tree, err := loadPlainFile(c, inputStore, fileName, fileBytes)
+		fileBytes, readFileErr := ioutil.ReadFile(fileName)
+		plainTree, loadPlainFileErr := loadPlainFile(c, inputStore, fileName, fileBytes)
+		encryptedTree, loadEncryptedFileErr := loadEncryptedFile(c, inputStore, fileBytes)
```
jvehent: This is weird. I think I understand why you did it this way, but it's a very unusual flow. At the very least, add a comment to explain why it's done this way.

PR #160, opened 2016-11-03
Add --set flag

autrilla:
The behavior is a bit different from the Python version. Instead of `--set '["app2"]["key"]' '"app2keystringvalue"'`, you have to do `--set '["app2"]["key"] "app2keystringvalue"'`

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -15,6 +15,9 @@ import (
 	"strings"
 	"time"
 
+	encodingjson "encoding/json"
+	"reflect"
+
 	"go.mozilla.org/sops/aes"
 	"go.mozilla.org/sops/json"
 	"go.mozilla.org/sops/kms"
@@ -29,6 +32,7 @@ const (
 	exitErrorDumpingTree                       int = 4
 	exitErrorReadingConfig                     int = 5
 	exitErrorInvalidKMSEncryptionContextFormat int = 6
+	exitErrorInvalidSetFormat                  int = 7
 	exitErrorEncryptingMac                     int = 21
 	exitErrorEncryptingTree                    int = 23
 	exitErrorDecryptingMac                     int = 24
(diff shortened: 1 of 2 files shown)


cmd/sops/main.go
```diff
@@ -192,6 +195,10 @@ func main() {
 			Name:  "encryption-context",
 			Usage: "comma separated list of KMS encryption context key:value pairs",
 		},
+		cli.StringFlag{
+			Name:  "set",
+			Usage: `set a specific key or branch in the input JSON or YAML document. value must be a json encoded string. (edit mode only). eg. --set '["somekey"][0] {"somevalue":true}'`,
```
jvehent: Does `--set` open an editor? I thought it would just set the value and exit.

cmd/sops/main.go
```diff
@@ -557,77 +564,115 @@ func edit(c *cli.Context, file string, fileBytes []byte) ([]byte, error) {
...
 		}
-		if bytes.Equal(newHash, origHash) {
-			return nil, cli.NewExitError("File has not changed, exiting.", exitFileHasNotBeenModified)
+		branch := parent.(sops.TreeBranch)
+		for i, kv := range branch {
+			if kv.Key == key {
+				branch[i].Value = valueTree
+				break
+			}
 		}
```
jvehent: Would it be cleaner to move the above chunk of code to a function outside of `main()` ?

cmd/sops/main.go
```diff
@@ -557,77 +564,115 @@ func edit(c *cli.Context, file string, fileBytes []byte) ([]byte, error) {
...
-		return nil, cli.NewExitError(fmt.Sprintf("Could not write output file: %s", err), exitCouldNotWriteOutputFile)
-	}
-	origHash, err := hashFile(tmpfile.Name())
-	if err != nil {
-		return nil, cli.NewExitError(fmt.Sprintf("Could not hash file: %s", err), exitCouldNotReadInputFile)
-	}
-	for {
-		err = runEditor(tmpfile.Name())
+	if c.String("set") != "" {
+		pathValuePair := strings.SplitAfterN(c.String("set"), "] ", 2)
```
jvehent: Need a comment that describes the algorithm we're following here, this is too convoluted to just follow the code.

cmd/sops/main.go
autrilla: It doesn't, why are you asking?

jvehent: Before the `Usage` string says `(edit mode only)`.

autrilla: Oh, I just copied that from the python version https://github.com/mozilla/sops/blob/python-sops/sops/__init__.py#L148-L153

PR #162, opened 2016-11-04
Add version comparison to sops cmd, fixes #161

### LINKED ISSUE #161: Implement version lookup
go-sops does not support version lookup. We could do it by storing the version in a separate file `version.go` and retrieving it as an `http.Get()` to the github raw url to compare with the current version.

jvehent:
r? @autrilla


cmd/sops/version.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if outdated {
+		out += fmt.Sprintf("[info] sops %s is available, update with `go get -u go.mozilla.org/sops/cmd/sops`\n", upstreamVersion)
+	}
+	fmt.Fprintf(c.App.Writer, "%s", out)
+}
+
+// A_is_newer_than_B takes 2 semver strings are returns true
+// is the A is newer than B, false otherwise
+func A_is_newer_than_B(A, B string) (bool, error) {
```
autrilla: nit: why snake_case?

cmd/sops/version.go
```diff
@@ -0,0 +1,76 @@
...
+func retrieveLatestVersionFromUpstream() (string, error) {
+	resp, err := http.Get("https://raw.githubusercontent.com/mozilla/sops/master/cmd/sops/version.go")
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	scanner := bufio.NewScanner(resp.Body)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if len(line) > 20 && line[0:17] == `const version = "` {
```
autrilla: Maybe it would be a better idea to load the version from a VERSION file? this looks way too complicated to me

cmd/sops/version.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if vA.Compare(vB) > 0 {
+		// vA is newer than vB
+		return true, nil
+	}
+	return false, nil
+}
+
+// retrieveLatestVersionFromUpstream gets the latest version from the source code at Github
+func retrieveLatestVersionFromUpstream() (string, error) {
```
autrilla: If we're sticking to semver, I think this should return a semver struct and not a string, that way we can get rid of `A_is_newer_than_B` and just have a static `var version, _ = semver.Make("whatever-our-version-is")` on the top-level of `main.go`.

cmd/sops/version.go
jvehent: Using a non Go file is impractical because the compiled binary would not know its own version without reading that file, which may or may not exist on the file system.

cmd/sops/version.go
jvehent: We'll still need to parse the semver found in documents and retrieved from upstream, so I'm not sure what this buys us.

cmd/sops/version.go
jvehent: because I reused the python syntax :p will change it to cAmElCaSe

cmd/sops/version.go
autrilla: Yes, it won't exist on the file system, but we aren't reading it from the file system anyway, we're reading it from GitHub. To get the local version we can just have a variable on main.go, like I mention in the comment above.

cmd/sops/version.go
autrilla: It would result in not needing `A_is_newer_than_B` (`latestVersion.Compare(currentVersion)` reads equally well) and making the function to parse the latest version much, much simpler

cmd/sops/version.go
jvehent: But that would require storing the version in 2 places instead of just one.

autrilla: That's a good point. If we go this route, at least I'd have `strings.StartsWith(`const version = "`)` instead of `len(line) > 20 && line[0:17] == `const version = "``.

cmd/sops/version.go
jvehent: I think I like it the way it is. 😈

jvehent:
@autrilla Could you give this another pass and merge it?

autrilla:
(I reran the travis job due to that intermittent test we have. I have to open an issue for that test.)

PR #167, opened 2016-11-05
when file extension ends is yml assume yaml format

jessechahal:


cmd/sops/main.go
```diff
@@ -328,6 +328,8 @@ func outputStore(context *cli.Context, path string) sops.Store {
 func defaultStore(path string) sops.Store {
 	if strings.HasSuffix(path, ".yaml") {
 		return &yaml.Store{}
+	} else if strings.HasSuffix(path, ".yml") {
+		return &yaml.Store{}
```
autrilla: nit: I think this makes it look as if .yaml and .yml were actually different things. I would prefer:

```
if strings.HasSuffix(path, ".yaml") || strings.HasSuffix(path, ".yml") {
 	return &yaml.Store{}
} ...
```
Instead

jessechahal: updated

Issue #163, opened 2016-11-04
Fix intermittent test failure


autrilla:
[This test](https://github.com/mozilla/sops/blob/520eab6b9a83172a6f25a802f70264ad4ff974e2/kms/keysource_test.go#L87-L107) is failing intermittently due to the fact that maps are unordered in Go, and sometimes the output is `key2:value2,key1:value1` instead of the expected `key1:value1,key2:value2`

autrilla:
Fixed by #169

PR #169, opened 2016-11-06
When converting master key to map, store context keys alphabetically sorted

alvarogzp:
This fixes issue #163, and make toMap() more stable across different runs.


autrilla:
Also, for the future, if you include "Fixes #163" in your commit message, GitHub will automatically close the issue for you.

PR #177, opened 2016-11-14
Functional tests

autrilla:
This should lay the foundations towards having tests for the sops binary itself.

r? @g-k


g-k:
r+

curious why you chose rust for the functional tests

autrilla:
I wanted to avoid Go to prevent anyone from calling Go functions to work around stuff in the future. I suppose I could have picked Python or any other language other than Rust.

g-k:
Cool!  That makes sense.

PR #179, opened 2016-11-18
Marshal extracted values with the corresponding store

### LINKED ISSUE #174: --extract should marshal as the document type
Currently `--extract` manually serializes values instead of just marshaling as YAML, like the Python version does. This makes go-sops unable to extract maps.

autrilla:
Fixes #174

### DIFF
--- a/json/store.go
+++ b/json/store.go
@@ -39,7 +39,7 @@ func (store BinaryStore) MarshalWithMetadata(tree sops.TreeBranch, metadata sops
 
 // MarshalValue is unusable for BinaryStore
 func (store BinaryStore) MarshalValue(v interface{}) ([]byte, error) {
-	panic("BinaryStore can not marshal a single value")
+	return nil, fmt.Errorf("Binary files are not structured and extracting a single value is not possible")
 }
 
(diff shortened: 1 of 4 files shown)


json/store.go
```diff
@@ -37,6 +37,11 @@ func (store BinaryStore) MarshalWithMetadata(tree sops.TreeBranch, metadata sops
 	return store.store.MarshalWithMetadata(tree, metadata)
 }
 
+// MarshalValue is unusable for BinaryStore
+func (store BinaryStore) MarshalValue(v interface{}) ([]byte, error) {
+	panic("BinaryStore can not marshal a single value")
```
jvehent: Why `panic` when you can just return an `error` ?

jvehent: I'm concerned that the stack trace returned by a panic will prevent users from understanding they made a mistake in their command line flags, whereas an explicit error help them.

I think the error message should be something like `error: binary files are not structured and extracting a single value is not possible.`

autrilla: Yeah, that's more intuitive for users. Fixed.

PR #180, opened 2016-11-18
Allow --set to add keys to maps

### LINKED ISSUE #172: Allow --set to add keys to maps
Currently using `--set` to set an inexistent key will throw an error. We should instead add that key and set its value to whatever is passed, like the Python version does.

autrilla:
Fixes #172


jvehent:
Would be nice to have a test for this one, the behavior is a bit tricky.

PR #183, opened 2016-11-19
Actually look for vim or nano when no $EDITOR is provided

autrilla:


jvehent:
teh version, you need to bump eet.

PR #186, opened 2016-12-30
Make sure GPG_EXEC is set before running GPG

dann7387:
GPG_EXEC was only being set when run as a CLI tool.
It was not being set when used as a Python library.

### DIFF
--- a/sops/__init__.py
+++ b/sops/__init__.py
@@ -382,6 +382,7 @@ def main():
     print("INFO: file written to %s" % (path), file=sys.stderr)
     sys.exit(0)
 
+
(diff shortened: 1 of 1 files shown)


dann7387:
@jvehent Looks like that Travis CI tests fail to install Python on OSX. Is this a known problem?

jvehent:
If you rebase against `python-sops`, the osx issue should be fixed now.

jvehent:
Looks good. I'll need to release a new version on pip. Will handle that tomorrow.

jvehent:
Sops 1.16 has been released to pypi. Thanks for the patch!

PR #196, opened 2017-02-04
Allow decrypting files with no MAC

### LINKED ISSUE #195: panic when YAML has no MAC
Similar to #194 except it crashed on the shared git files.

panic: interface conversion: interface is nil, not string

goroutine 1 [running]:
panic(0x49a160, 0xc420338140)
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/panic.go:500 +0x1a1
go.mozilla.org/sops/yaml.(*Store).UnmarshalMetadata(0x785f68, 0xc42015c000, 0xb58, 0xd58, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/jbuckley/Documents/go/src/go.mozilla.org/sops/yaml/store.go:159 +0x3c1
main.loadEncryptedFile(0xc420095180, 0x749620, 0x785f68, 0xc42015c000, 0xb58, 0xd58, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/jbuckley/Documents/go/src/go.mozilla.org/sops/cmd/sops/main.go:74 +0xa1
main.main.func1(0xc420095180, 0x0, 0x0)
	/Users/jbuckley/Documents/go/src/go.mozilla.org/sops/cmd/sops/main.go:227 +0x383
go.mozilla.org/sops/vendor/gopkg.in/urfave/cli%2ev1.HandleAction(0x47aae0, 0x55bb90, 0xc420095180, 0x0, 0x0)
	/Users/jbuckley/Documents/go/src/go.mozilla.org/sops/vendor/gopkg.in/urfave/cli.v1/app.go:485 +0xd4
go.mozilla.org/sops/vendor/gopkg.in/urfave/cli%2ev1.(*App).Run(0xc420105d40, 0xc42000c1e0, 0x3, 0x3, 0x0, 0x0)
	/Users/jbuckley/Documents/go/src/go.mozilla.org/sops/vendor/gopkg.in/urfave/cli.v1/app.go:259 +0x74f
main.main()
	/Users/jbuckley/Documents/go/src/go.mozilla.org/sops/cmd/sops/main.go:284 +0xe14

autrilla:
Includes functional test that checks that sops returns with a successful exit code when you give it a file without a MAC and --ignore-mac.

Fixes #195

cc @jbuck 

@jvehent: While doing this I noticed there's a lot of duplication on the JSON and YAML stores to parse the metadata. What do you think of refactoring it so that each store just returns its metadata as a `map[interface{}]interface{}`, and then another method (the same for all trust stores) converts that map into the proper struct?


jvehent:
Refactoring is always an excellent idea!

sops.go
```diff
@@ -468,3 +468,109 @@ func ToBytes(in interface{}) ([]byte, error) {
 		return nil, fmt.Errorf("Could not convert unknown type %T to bytes", in)
 	}
 }
+
+// MapToMetadata tries to convert a map[string]interface{} obtained from an encrypted file into a Metadata struct.
+func MapToMetadata(data map[string]interface{}) (Metadata, error) {
+	var metadata Metadata
+	metadata.MessageAuthenticationCode, _ = data["mac"].(string)
```
jvehent: Should there be a friendly log message here to indicate that the MAC is ignored because none was found?

autrilla: I think it would be nicer, yes. Otherwise you'll just get something like:

`MAC mismatch failed, got , expected AAAAAAAAAAAAAAAAAA`

jvehent:
Now that I'm looking at this again, would it allow an attacker to delete the MAC from a document and modify it, without `sops` catching it?

If a MAC is missing, we shouldn't blindly ignore the issue, unless the `ignore-mac` flag is set by the user.

autrilla:
No. If the MAC is missing, and ignore-mac is not passed in, it will show a warning telling the user the file has no MAC, and fail. The file's MAC is set to an empty string if it doesn't exist, and that's never going to match the computed MAC.

autrilla:
Maybe we could improve the message a bit by saying "MAC mismatch. File has no MAC, computed {}", instead of "MAC mismatch. File has , computed {}"

jvehent:
To you want to modify it before I merge it?

autrilla:
Yes, I'm working on it right now. I'm also adding a test that checks
decryption fails when the file has no MAC and --ignore-mac is not passed in.

PR #209, opened 2017-05-25
Shamir Secret Sharing support

### LINKED ISSUE #208: Implement Shamir's Secret Sharing for data key retrieval
Suggested by @jvehent.

Currently, every single master key has the capacity to fully decrypt the data key, and therefore the whole file. It would be desirable to have a way to require access to more than one master key in order to retrieve the data key. Using [Shamir's Secret Sharing](https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing), if one master key is compromised, the attacker can't get the data key, as they will need a quorum of keys to retrieve it. We can use [Vault's implementation](https://github.com/hashicorp/vault/tree/master/shamir) of the algorithm.

This should be relatively easy to implement by modifying the way we retrieve the data key from the encrypted file, `GetDataKey` in `sops.go`; and the way we write the data key to the encrypted file, `UpdateMasterKeys` in `sops.go`. 

The number of master keys required to form a quorum should probably be configurable, defaulting to half the number of keys plus one.

autrilla:
Adds Shamir Secret Sharing support to the library. 

When, ready, will fix #208
Blocking merge:
- [x] Integration with the command line tool
- [x] Documentation for the feature in the `README`
- [x] Functional tests


sops.go
```diff
@@ -43,6 +43,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/vault/shamir"
```
jvehent: it's a very small library inside a gigantic project. I think we should duplicate it.

cmd/sops/main.go
```diff
@@ -200,6 +206,14 @@ func main() {
 			Name:  "set",
 			Usage: `set a specific key or branch in the input JSON or YAML document. value must be a json encoded string. (edit mode only). eg. --set '["somekey"][0] {"somevalue":true}'`,
 		},
+		cli.BoolFlag{
+			Name: "shamir",
```
jvehent: "sss" might be a better flag, standing for "shamir secret splitting"

README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
 .. code:: bash
 
 	SOPS_GPG_EXEC = 'your_gpg_client_wrapper'
+
+Shamir Secret Sharing
+~~~~~~~~~~~~~~~~~~~~~
+
+By default, `sops` encrypts the data key with all master keys, such that if any
```
jvehent: s/with all master keys/with each master key/

README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
...
+of the master keys is available, the whole file can be decrypted. Sometimes, it
```
jvehent: s/keys/key/

README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
...
+is desirable to require access to several keys in order to be able to decrypt
```
jvehent: several **master** keys

README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
...
+files. This can be achieved with Shamir's Secret Sharing. With this method,
+the data key is split into several parts, one for each master key, and `quorum`
+parts are required in order to retrieve the data key and decrypt the file.
+
+You can enable this mode by passing `--shamir` to the encrypt mode or by passing
+it to the edit mode for new files. You can set `quorum` with
+`--shamir-quorum number`.
```
jvehent: The quorum must be lower than or equal to the number of master keys.

README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
...
+
+`quorum` defaults to half the available master keys plus one.
```
jvehent: Why plus one? Is it rounded to the upper or lower value?

README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
...
+
+For example:
+
+```
+sops --shamir --shamir-quorum 5 example.json
+```
+
+This will require at least 5 master keys are available in order to decrypt
```
jvehent: remove "are available"

cmd/sops/version.go
```diff
@@ -10,7 +10,7 @@ import (
 	"gopkg.in/urfave/cli.v1"
 )
 
-const version = "2.0.9"
+const version = "2.0.10"
```
jvehent: You shouldn't bump the version. It gets painful when other pull requests are merged prior to the one bumping the version, so I do it in the master branch directly, or in a separate PR.

shamir/README.md
```diff
@@ -0,0 +1 @@
+Forked from [Vault](https://github.com/hashicorp/vault/tree/master/shamir)
```
jvehent: also put a copy of the original LICENSE file in this directory, even if it's the same as the one `sops` uses.

sops.go
```diff
@@ -105,10 +108,10 @@ type Tree struct {
 
 // TrimTreePathComponent trimps a tree path component so that it's a valid tree key
 func TrimTreePathComponent(component string) (string, error) {
-	if component[len(component)-1] != ']' {
+	if component[len(component) - 1] != ']' {
 		return "", fmt.Errorf("Invalid component")
 	}
-	component = component[:len(component)-1]
+	component = component[:len(component) - 1]
```
jvehent: did something change in gofmt?

sops.go
```diff
@@ -288,6 +291,12 @@ type Metadata struct {
 	MessageAuthenticationCode string
 	Version                   string
 	KeySources                []KeySource
+	// Shamir specifies if the data key this file is encrypted with was
+	// split between all key sources using Shamir's Secret Sharing.
+	Shamir                    bool
```
jvehent: nit on the comment: "Shamir is `true` when the data key is split across multiple master keys according to the shamir's secret splitting algorithm"

sops.go
```diff
@@ -288,6 +291,12 @@ type Metadata struct {
...
+	// ShamirQuorum is the number of parts required to recover the original
```
jvehent: s/parts/master key/

sops.go
```diff
@@ -357,8 +371,47 @@ func (m *Metadata) UpdateMasterKeysIfNeeded(dataKey []byte) (errs []error) {
...
+// updateMasterKeysShamir splits the data key into parts using Shamir's Secret
+// Sharing algorithm and encrypts each part with a master key
+func (m *Metadata) updateMasterKeysShamir(dataKey []byte) (errs []error) {
+	keyCount := 0
+	for _, ks := range m.KeySources {
+		for range ks.Keys {
+			keyCount++
+		}
+	}
+	defaultQuorum := (keyCount / 2) + 1
```
jvehent: how about this?
```go
defaultQuorum := math.Floor(float64(keyCount)/2.0 + 0.5)
```

sops.go
```diff
@@ -357,8 +371,47 @@ func (m *Metadata) UpdateMasterKeysIfNeeded(dataKey []byte) (errs []error) {
...
+	// If the quorum wasn't set, default to half the keys plus one
+	if m.ShamirQuorum == 0 {
+		m.ShamirQuorum = defaultQuorum
+	}
```
jvehent: nit: rewrite this as
```go
if m.ShamirQuorum == 0 {
    m.ShamirQuorum := math.Floor(float64(keyCount)/2.0 + 0.5)
}
```
and get rid of defaultQuorum entirely

autrilla:
There are unit tests in this PR already (in `shamir/shamir_test.go`), did you miss them or are you thinking of testing something else?

I think integration tests for this are a good idea, I'll get on that.

I don't have an opinion on what the quorum should default to. I think a default of 2 is alright because it's still an improvement over not using SSS, and it does have the advantage you mention of being easier to explain to users, so I'll change it to default to 2.

sops.go
autrilla: No, but I switched from gofmt to goimports, and it seems they differ in a few small things. I don't know whether it was you or oremj that told me to use goimports because it sorts imports 😜

jvehent:
By unit tests, I meant tests in [sops_test.go](https://github.com/mozilla/sops/blob/master/sops_test.go), but integration tests might be enough.

cmd/sops/main.go
autrilla: I think `shamir` is clearer. Maybe `shamir-secret-sharing` is even better, although I feel it's a bit too verbose.

jvehent: You don't call RSA `-rivest`, do you? 😉

autrilla:
Unfortunately urfave/cli doesn't allow multiple flags for the same flag, so I think we'll have to settle with `--shamir-secret-sharing`.

Anyway, there's now two functional tests for this feature:

- One encrypts a file with 10 master keys (they're actually all the same PGP key, but this doesn't matter to sops), and tries decrypting it, and asserts that it succeeds.
- The other one encrypts a file with 10 master keys, corrupts the first PGP message (the way I do this is very hacky), tries decrypting again, and asserts that it fails.

jvehent:
Now being implemented in #238

PR #211, opened 2017-05-26
Use shlex to parse $EDITOR

### LINKED ISSUE #210: Editor arguments are not supported properly
When you pass arguments to your editor sops doesn't understand them. For example:

```
export EDITOR="/usr/local/bin/emacs -nw"
```

results in the following:

```
sops file
Could not run editor: fork/exec /usr/local/bin/emacs -nw: no such file or directory
```

Removing the `-nw` makes this problem go away.

autrilla:
Fixes #210 by parsing $EDITOR like a shell would so sops can pass arguments to the editor.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -18,6 +18,8 @@ import (
 	encodingjson "encoding/json"
 	"reflect"
 
+	"github.com/google/shlex"
+
 	"go.mozilla.org/sops/aes"
 	"go.mozilla.org/sops/json"
 	"go.mozilla.org/sops/kms"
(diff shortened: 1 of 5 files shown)


cmd/sops/main.go
```diff
@@ -295,7 +297,11 @@ func runEditor(path string) error {
 		}
 		cmd = exec.Command(strings.Split(string(out), "\n")[0], path)
 	} else {
-		cmd = exec.Command(editor, path)
+		parts, err := shlex.Split(editor)
```
autrilla: It uses bourne shell rules to split the string. We could achieve something that would work in most cases just splitting on spaces, but that would break if, for example, someone had something like:

```
$EDITOR=editor "foo bar"
```

I expect there are many more edge cases where simply splitting on spaces wouldn't work.

PR #217, opened 2017-07-26
[PYTHON] Unpin cryptography requirement

rouge8:
Doesn't seem like there's a reason not to use newer versions of cryptography and it has the advantage of being much easier to install with wheels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     install_requires=[
         'ruamel.yaml==0.11.7',
         'boto3>=1.1.3',
-        'cryptography==1.4',
+        'cryptography>=1.4',
         'setuptools>=18.8.1',
         'ordereddict>=1.1',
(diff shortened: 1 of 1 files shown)


jvehent:
Anndd of course CI on OSX is busted again. I'll have to debug it before we can merge this.
The reason cryptography is pinned is mostly due to OSX's package management fragility, so I want to be careful about merging this.

rouge8:
@jvehent I think I fixed Travis for OS X in #218

autrilla:
I don't think we should use `>=` in our dependency requirements. It's prone to breakage due to dependencies updating with breaking changes. I'm okay with updating this dependency in particular, but I think longer-term it'd be better to use something like pyup.io to keep dependencies updated while keeping builds reproducible.

autrilla:
Closing as this has made no progress, feel free to reopen

Issue #214, opened 2017-07-18
no love for the gov


mrgleeco:
i've great success with running sops in a typical aws account.  However, when i run in us-gov-west-1, there is always an error:

```
Could not load file: Error encrypting the data key with one or more master keys: [Failed to encrypt new data key with master key "<redacted  keyname>": Failed to create session: No valid ARN found in "<redacted keyname>"]
```

it looks as if the `keysource.go` is hardcoded regex for `^arn:aws:kms:`.
But the arn string for govcloud has a pattern like `arn:aws-us-gov:kms:us-gov-west-1:1234567890:key/...`

autrilla:
Thanks for the bug report.

According to [Amazon](http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-arns) the ARN formats are:

```
arn:partition:service:region:account-id:resource
arn:partition:service:region:account-id:resourcetype/resource
arn:partition:service:region:account-id:resourcetype:resource
```

With `partition` being `aws-partitionname` or `aws`.

So I suppose we could just change the regex to `^arn:aws[\w-]+:kms:(.+):([0-9]+):key/(.+)$`. Ideally we'd find some sort of function in the AWS SDK that lets us extract the region and role from the ARN.

mrgleeco:
sdk looks to define (& export) at https://github.com/aws/aws-sdk-go/blob/master/aws/endpoints/defaults.go#L11-L13

possibly a regex with a test to string match one of these?  Feels brittle, but also likely not too much changes here.

mrgleeco:
ping -
and thx @autrilla for looking

i've a patch in branch at https://github.com/gleeco/sops/commit/b715e29f6b821920953d3d7c1c5bdef10e7aab53

i've not a testing story;  there looks like special hooks wrt `createSession` which may mean insane mockery;  any guidance from the team?

thx in advance!

autrilla:
This has been fixed

mrgleeco:
👍  - is there a release schedule?

jvehent:
I just released 2.0.10 https://github.com/mozilla/sops/releases/tag/2.0.10

PR #228, opened 2017-08-14
change: check valid arn partition

gleeco:
[issue 214](https://github.com/mozilla/sops/issues/214)

Change to make `createSession` handle the current valid AWS partitions:

* `aws`
* `aws-cn`
* `aws-us-gov`

Previously only `aws` had been hard-coded in the arn check

### DIFF
--- a/kms/keysource.go
+++ b/kms/keysource.go
@@ -140,13 +140,12 @@ func (key MasterKey) createStsSession(config aws.Config, sess *session.Session) 
 }
 
 func (key MasterKey) createSession() (*session.Session, error) {
-	// possible partitions in $1: aws, aws-cn, aws-us-gov
-	re := regexp.MustCompile(`^arn:(aws[\w-]*):kms:(.+):([0-9]+):key/(.+)$`)
+	re := regexp.MustCompile(`^arn:aws[\w-]*:kms:(.+):[0-9]+:key/.+$`)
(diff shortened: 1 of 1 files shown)


kms/keysource.go
```diff
@@ -140,12 +140,22 @@ func (key MasterKey) createStsSession(config aws.Config, sess *session.Session)
...
+		"aws-cn": 2,
+		"aws-us-gov": 3,
+	}
+	re := regexp.MustCompile(`^arn:(aws[\w-]*):kms:(.+):([0-9]+):key/(.+)$`)
 	matches := re.FindStringSubmatch(key.Arn)
 	if matches == nil {
 		return nil, fmt.Errorf("No valid ARN found in %q", key.Arn)
 	}
-	config := aws.Config{Region: aws.String(matches[1])}
+	_, exists := validPartition[matches[1]]
```
autrilla: I don't think this check is needed. We don't currently check that the region is valid, so I don't see why we should check whether the partition is valid or not. In my opinion we should just expand the regex to

```
re := regexp.MustCompile(`^arn:aws[\w-]*:kms:(.+):([0-9]+):key/(.+)$`)
```

gleeco:
well - i saw the earlier comment & not the latter - and pushed up the simplified version :-/
lmk if this works, otherwise quick force-push to restore

PR #233, opened 2017-08-21
Basic backwards and forward compatibility tests

autrilla:
What the tests do:

- Encrypt a file with all old versions of SOPS
- Check that all old files can be decrypted with the latest version of SOPS
- Encrypt a file with the latest version of SOPS
- Check that the file can be decrypted with all old versions of SOPS


functional-tests/make-binaries
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+go build -o ./sops go.mozilla.org/sops/cmd/sops
+BRANCH=`git rev-parse --abbrev-ref HEAD`
+git tag | grep "2.*" | xargs -I {} bash -c "git checkout {} && go build -o /tmp/sops{} go.mozilla.org/sops/cmd/sops"
```
jvehent: why only build `2.x` tags?

functional-tests/resources/file.yaml
```diff
@@ -0,0 +1,13 @@
+first_name: John
+last_name: Doe
+age: 45
+phone_numbers:
+  - type: work
+    number: '1234567890'
```
jvehent: that's not a number, that's a string

functional-tests/make-binaries
autrilla: Because we can't build 1.x, but yeah, building everything but 1.x is the right approach.

functional-tests/resources/file.yaml
autrilla: Phone numbers are sometimes not just numbers, see 17) and the following here https://github.com/googlei18n/libphonenumber/blob/master/FALSEHOODS.md

PR #238, opened 2017-08-25
Remote keys with Shamir Secret Sharing

autrilla:
This is a big PR. The documentation is not ready yet. We could probably use some more tests.

The tests are failing due to dependency vendoring.

Hopefully this description of the changes will aid code review.

# Key groups and Shamir

This PR introduces the concept of key groups. A SOPS file has at least one key group. A key group is a set of master keys. Each key group is assigned a part of the AES data key. If there is only one key group, the group is assigned the whole data key. If there is more than one key group, the data key is split up into as many parts as there are groups using [Shamir's Secret Sharing](https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing). Optionally, the user can provide the number of groups required to recover the data key. As a default, all groups are required in order to recover the data key. All keys inside one group encrypt the same information, so access to only one key per group is required for decryption.

## Compatibility

With this change, there is now a difference between the internal representation SOPS has of files, and the actual files stored on the disk. Old files have no concept of key groups, but new versions of SOPS do. SOPS will load files with the old file format, and internally convert from the old representation to the new internal representation, where there will only be one key group with all master keys the old file had. This means we preserve full backwards-compatibility in this respect.

Files with only one key group will be saved in the same binary format previous versions of SOPS use, so they will be forwards-compatible as well.

Old versions of SOPS will not be able to decrypt files with more than one key group.

# Remote keys

The other big new feature in this PR are remote keys. SOPS now uses a client-server approach to encrypting and decrypting the data key. We call this the key service. By default, SOPS runs a local key service in-process, so most users won't even notice this change. Previously, whenever we needed to encrypt or decrypt a data key, we'd call `Encrypt` or `Decrypt` on a `MasterKey` interface implementer. From this version onward, we use a key service client to send an encrypt or decrypt request to a key service, which then performs the operation for us. Currently, the key server simply creates a new `MasterKey` interface implementer every time and then calls `Encrypt` or `Decrypt` on it, but in future versions we could expect the `Encrypt` and `Decrypt` methods to be fully removed from the `MasterKey` interface.

The requests are sent using [gRPC](https://grpc.io/) and [Protocol Buffers](https://github.com/google/protobuf). There are currently two possible requests: an `EncryptRequest` and a `DecryptRequest`. Both of these requests contain an identifier for the key they should perform the operation with, and the plaintext or encrypted data key, respectively. The requests *do not* contain any cryptographic keys, public or private.

The key service supports all key types SOPS supports, so both KMS and PGP keys. Whenever we try to encrypt or decrypt a data key, SOPS will try to do so first with the local key service (unless it's disabled), and if that fails, it will try all other remote key services until one succeeds.

## Compatibility

This new feature does not affect compatibility with older SOPS versions.

# Store refactor

As mentioned under the key groups section, this PR introduces features that have to change how SOPS works with files internally. For this reason, I refactored the store logic to separate the internal representation SOPS has of files from the external representation used in JSON and YAML files.

The existing code that mapped YAML/JSON files to the internal Go structs and back was messy, verbose, and error prone, so I think a change like this would have been good regardless.

## Compatibility

This breaks backward compatibility for JSON files created with SOPS 1.x. This is due to SOPS 1.x encoding the `version` field (which represents the SOPS version the file was created with) in JSON files as a number, instead of a string. This change uses `encoding/json` to decode the SOPS file directly into a Go struct. Unfortunately, `encoding/json` refuses to decode a JSON number into a Go string.

There is an upgrade path, however, as any time users use versions of SOPS from 2.0.0 to 2.0.10 on the file, the `version` field will be reencoded as a JSON string. Alternatively users can manually convert the field to a string.

We can try to provide a good error message that guides users towards a solution for this problem before this PR is merged.

# `cmd/sops` refactor

I have split up the CLI into 4 "commands": edit, encrypt, decrypt, and rotate. Each of the commands lives in its own file, and has a defined set of options that modify it. Previously, it was unclear what flags applied when, so this refactor should make that more transparent. This would also make it easier to redesign the CLI to a git-like command-based interface many popular CLIs use (docker and kubernetes come to mind) if we ever wanted to.

## Compatibility

The CLI is still the same as in older versions of SOPS, with a few extra flags added to it.


README.rst
```diff
@@ -392,6 +392,36 @@ Example: place the following in your `~/.bashrc`
 .. code:: bash
 
 	SOPS_GPG_EXEC = 'your_gpg_client_wrapper'
+
```
autrilla: This has to be rewritten

example.json
```diff
@@ -20,7 +20,7 @@
     ],
     "anEmptyValue": "",
     "sops": {
-        "version": 1.6,
+        "version": "1.6",
```
autrilla: I've changed the example to use a string version field, because otherwise it doesn't work with new SOPS versions

keyservice/keyservice.pb.go
```diff
@@ -0,0 +1,432 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
autrilla: This is auto-generated and does not need to be reviewed.

pgp/sops_functional_tests_key.asc
```diff
@@ -32,3 +32,94 @@ sZlaMHbqJX6TpOiFU8FvMU8syhBgWNYYq/3JNKw9GVDabO5Ggbluup/q1/uJHO1y
 v8bK3sq5s580DSaWamLgSLwl3UfCnL3Qa0XDT6SoY1L/Wzr1iW+9
 =zqxY
 -----END PGP PRIVATE KEY BLOCK-----
+-----BEGIN PGP PRIVATE KEY BLOCK-----
```
autrilla: I have added a new public/private key pair and a public key to this file, so that tests can use key groups.

shamir/shamir.go
```diff
@@ -0,0 +1,260 @@
+package shamir
```
autrilla: This is forked from [Vault](https://github.com/hashicorp/vault/tree/master/shamir) and does not need to be reviewed.

shamir/shamir_test.go
```diff
@@ -0,0 +1,198 @@
+package shamir
```
autrilla: This is forked from [Vault](https://github.com/hashicorp/vault/tree/master/shamir) and does not need to be reviewed.

shamir/tables.go
```diff
@@ -0,0 +1,77 @@
+package shamir
```
autrilla: This is forked from [Vault](https://github.com/hashicorp/vault/tree/master/shamir) and does not need to be reviewed.

shamir/tables_test.go
```diff
@@ -0,0 +1,13 @@
+package shamir
```
autrilla: This is forked from [Vault](https://github.com/hashicorp/vault/tree/master/shamir) and does not need to be reviewed.

sops.go
```diff
@@ -103,40 +111,32 @@ type Tree struct {
...
-	component = strings.Replace(component, `"`, "", 2)
-	component = strings.Replace(component, `'`, "", 2)
-	return component, nil
-}
-
-// Truncate truncates the tree following Python dictionary access syntax, for example, ["foo"][2].
-func (tree TreeBranch) Truncate(path string) (interface{}, error) {
-	components := strings.Split(path, "[")
+// Truncate truncates the tree to the path specified
+func (tree TreeBranch) Truncate(path []interface{}) (interface{}, error) {
```
autrilla: Reasoning behind this change: parsing of the component string should take place in the `cmd` package, not in the `sops` package.

sops.go
```diff
@@ -318,142 +320,128 @@ type Store interface {
...
+// GetDataKeyWithKeyServices retrieves the data key, asking KeyServices to decrypt it with each
+// MasterKey in the Metadata's KeySources until one of them succeeds.
+func (m Metadata) GetDataKeyWithKeyServices(svcs []keyservice.KeyServiceClient) ([]byte, error) {
+	if m.DataKey != nil {
+		return m.DataKey, nil
 	}
-}
-
-// AddKMSMasterKeys parses the input comma separated string of AWS KMS ARNs, generates a KMS MasterKey for each ARN, and then adds the keys to the KMS KeySource
-func (m *Metadata) AddKMSMasterKeys(kmsArns string, context map[string]*string) {
```
autrilla: This is a niche operation and I think it belongs in the `cmd` package, not in `sops`.

sops.go
```diff
@@ -318,142 +320,128 @@ type Store interface {
...
+				break
 			}
-			ks.Keys = append(ks.Keys, keys...)
-			m.KeySources[i] = ks
 		}
 	}
-}
-
-// RemovePGPMasterKeys takes a comma separated string of PGP fingerprints and removes the keys corresponding to those fingerprints from the metadata's KeySources
-func (m *Metadata) RemovePGPMasterKeys(pgpFps string) {
```
autrilla: This is a niche operation and I think it belongs in the `cmd` package, not in `sops`.

sops.go
```diff
@@ -318,142 +320,128 @@ type Store interface {
...
-		}
-	}
-	return
+	return m.UpdateMasterKeysWithKeyServices(dataKey, []keyservice.KeyServiceClient{
+		keyservice.NewLocalClient(),
+	})
 }
 
-// AddPGPMasterKeys parses the input comma separated string of GPG fingerprints, generates a PGP MasterKey for each fingerprint, and adds the keys to the PGP KeySource
-func (m *Metadata) AddPGPMasterKeys(pgpFps string) {
```
autrilla: This is a niche operation and I think it belongs in the `cmd` package, not in `sops`.

sops.go
```diff
@@ -318,142 +320,128 @@ type Store interface {
...
-func (m *Metadata) RemovePGPMasterKeys(pgpFps string) {
-	var keys []MasterKey
-	for _, k := range pgp.MasterKeysFromFingerprintString(pgpFps) {
-		keys = append(keys, k)
-	}
-	m.RemoveMasterKeys(keys)
-}
-
-// RemoveKMSMasterKeys takes a comma separated string of AWS KMS ARNs and removes the keys corresponding to those ARNs from the metadata's KeySources
-func (m *Metadata) RemoveKMSMasterKeys(arns string) {
```
autrilla: This is a niche operation and I think it belongs in the `cmd` package, not in `sops`.

yaml/config.go
```diff
@@ -62,27 +71,50 @@ func (f *configFile) load(bytes []byte) error {
 	return nil
 }
 
-// MasterKeyStringsForFile returns a comma separated string of KMS ARNs and a comma separated list of PGP fingerprints. If the config bytes are left empty, the function will look for the config file by itself.
-func MasterKeyStringsForFile(filepath string, confBytes []byte) (kms, pgp string, err error) {
+func KeyGroupsForFile(filepath string, confBytes []byte, kmsEncryptionContext map[string]*string) ([]sops.KeyGroup, error) {
```
autrilla: Instead of returning a bunch of strings, the configuration now returns key groups instead.

keyservice/server.go
```diff
@@ -0,0 +1,113 @@
...
+	var ctx map[string]*string
+	for k, v := range key.Context {
+		ctx[k] = &v
+	}
+	kmsKey := kms.MasterKey{
+		Arn:               key.Arn,
+		Role:              key.Role,
+		EncryptionContext: ctx,
+	}
+	err := kmsKey.Encrypt(plaintext)
```
milescrabill: this is that object-oriented code smell I'm talking about - we're using the MasterKey (interface) here to encrypt the data key, but we're not returning the encrypted data key via this method - instead we're setting a field of the MasterKey (struct) itself to the encrypted data key and then later using that value.

keys/keys.go
```diff
@@ -0,0 +1,13 @@
+package keys
+
+// MasterKey provides a way of securing the key used to encrypt the Tree by encrypting and decrypting said key.
+type MasterKey interface {
```
milescrabill: I would like to see this interface moved to be more functional. Encrypt in particular - it could take []byte (unencrypted data key) and return []byte (encrypted data key) rather than having implementing structs (kms.MasterKey and pgp.MasterKey) have fields for the encrypted data key.

sops.go
```diff
@@ -318,142 +320,128 @@ type Store interface {
...
-					break
-				}
-			}
-			if !matchFound {
-				newKeys = append(newKeys, k)
-			}
+func (m *Metadata) UpdateMasterKeysWithKeyServices(dataKey []byte, svcs []keyservice.KeyServiceClient) (errs []error) {
+	if len(svcs) == 0 {
+		return []error{
+			fmt.Errorf("No key services provided, can not update master keys."),
```
milescrabill: s/can not/cannot

README.rst
```diff
@@ -392,6 +392,76 @@ Example: place the following in your `~/.bashrc`
 .. code:: bash
```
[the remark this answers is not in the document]
autrilla: @jvehent

cmd/sops/edit.go
```diff
@@ -0,0 +1,271 @@
...
+	branch, err := opts.InputStore.Unmarshal(fileBytes)
+	if err != nil {
+		return nil, cli.NewExitError(fmt.Sprintf("Error unmarshalling file: %s", err), exitCouldNotReadInputFile)
+	}
+	tree.Branch = branch
+	tree.Metadata = sops.Metadata{
+		KeyGroups:         opts.KeyGroups,
+		UnencryptedSuffix: opts.UnencryptedSuffix,
+		Version:           version,
+		ShamirQuorum:      int(opts.GroupQuorum),
```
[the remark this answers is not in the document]
autrilla: I think the real question is why is ShamirQuorum a int, because negative values don't make sense

sops.go
```diff
@@ -184,7 +184,7 @@ func (tree TreeBranch) walkBranch(in TreeBranch, path []string, onLeaves func(in
 		}
 		key, ok := item.Key.(string)
 		if !ok {
-			return nil, fmt.Errorf("Tree contains a non-string key (type %T): %s. Only string keys are" +
+			return nil, fmt.Errorf("Tree contains a non-string key (type %T): %s. Only string keys are"+
```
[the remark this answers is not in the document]
autrilla: goimports

keys/keys.go
autrilla: The encrypted data key eventually has to be passed to the stores, because it has to be on the output file. Currently this is simple because MasterKeys store the encrypted data key themselves. I'm not sure how this could be done cleanly.

sops.go
```diff
@@ -281,35 +282,36 @@ func (tree Tree) GenerateDataKey() ([]byte, []error) {
...
-	ToString() string
-	ToMap() map[string]interface{}
-}
+type KeyGroup []keys.MasterKey
 
 // Store provides a way to load and save the sops tree along with metadata
 type Store interface {
 	Unmarshal(in []byte) (TreeBranch, error)
-	UnmarshalMetadata(in []byte) (Metadata, error)
+	UnmarshalMetadata(in []byte) (*Metadata, error)
```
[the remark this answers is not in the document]
autrilla: Reverted this change

pgp/keysource.go
```diff
@@ -24,8 +28,48 @@ type MasterKey struct {
 	CreationDate time.Time
 }
 
-// Encrypt encrypts the data key with the PGP key with the same fingerprint as the MasterKey. It looks for PGP public keys in $PGPHOME/pubring.gpg.
-func (key *MasterKey) Encrypt(dataKey []byte) error {
+func (key *MasterKey) EncryptedDataKey() []byte {
+	return []byte(key.EncryptedKey)
```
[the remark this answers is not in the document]
autrilla: EncryptedKey ends up in YAML/JSON files and therefore should probably be unicode, not arbitrary bytes

cmd/sops/edit.go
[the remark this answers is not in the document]
milescrabill: Looks like the shamir code expects an int. No reason to convert back and forth when the whole point is to pass it their.

sops.go
autrilla: Fixed

cmd/sops/edit.go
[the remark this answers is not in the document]
autrilla: Fixed

cmd/sops/edit.go
```diff
@@ -0,0 +1,271 @@
...
+		newBranch, err := opts.InputStore.Unmarshal(edited)
+		if err != nil {
+			fmt.Printf("Could not load tree: %s\nProbably invalid syntax. Press a key to return to the editor, or Ctrl+C to exit.", err)
+			bufio.NewReader(os.Stdin).ReadByte()
+			continue
+		}
+		if opts.ShowMasterKeys {
+			metadata, err := opts.InputStore.UnmarshalMetadata(edited)
+			if err != nil {
+				fmt.Printf("sops branch is invalid: %s.\nPress a key to return to the editor, or Ctrl+C to exit.", err)
```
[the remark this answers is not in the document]
autrilla: Fixed

cmd/sops/encrypt.go
```diff
@@ -0,0 +1,60 @@
...
+		ShamirQuorum:      int(opts.GroupQuorum),
+	}
+	dataKey, errs := tree.GenerateDataKeyWithKeyServices(opts.KeyServices)
+	if len(errs) > 0 {
+		err = fmt.Errorf("Could not generate data key: %s", errs)
+		return nil, err
+	}
+
+	err = encryptTree(encryptTreeOpts{
+		Stash: make(map[string][]interface{}), DataKey: dataKey, Tree: &tree, Cipher: opts.Cipher,
```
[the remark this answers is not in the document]
autrilla: Fixed

cmd/sops/main.go
```diff
@@ -286,35 +372,40 @@ func main() {
...
 		}
-		cmd = exec.Command(strings.Split(string(out), "\n")[0], path)
-	} else {
-		parts, err := shlex.Split(editor)
+		addr := url.Host
+		if url.Scheme == "unix" {
+			addr = url.Path
+		}
+		var opts []grpc.DialOption
+		opts = append(opts, grpc.WithInsecure())
```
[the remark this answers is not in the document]
autrilla: Fixed

cmd/sops/main.go
```diff
@@ -334,65 +425,43 @@ func outputStore(context *cli.Context, path string) sops.Store {
...
-			return nil, cli.NewExitError(err.Error(), exitInvalidTreePathFormat)
+		if component[len(component)-1] != ']' {
+			return nil, fmt.Errorf("component %s doesn't end with ]", component)
 		}
-		if newBranch, ok := v.(sops.TreeBranch); ok {
-			tree.Branch = newBranch
+		component = component[:len(component)-1]
+		if component[0] == byte('"') || component[0] == byte('\'') {
+			// The component is a string
+			component = component[1 : len(component)-1]
```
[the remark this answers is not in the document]
autrilla: I don't think this is done elsewhere, or at least I can't find it.

cmd/sops/rotate.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Add new master keys
+	for _, key := range opts.AddMasterKeys {
+		tree.Metadata.KeyGroups[0] = append(tree.Metadata.KeyGroups[0], key)
+	}
+	// Remove master keys
```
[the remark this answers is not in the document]
autrilla: Fixed

pgp/keysource.go
```diff
@@ -24,8 +28,48 @@ type MasterKey struct {
...
+	cmd.Stderr = &stderr
+	err := cmd.Run()
+	if err != nil {
+		return err
+	}
+	key.EncryptedKey = stdout.String()
+	return nil
+}
+
+func (key *MasterKey) encryptWithCryptoOpenpgp(dataKey []byte) error {
```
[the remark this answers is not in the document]
autrilla: Fixed

keyservice/keyservice.proto
```diff
@@ -0,0 +1,41 @@
...
+}
+
+message DecryptResponse {
+	bytes plaintext = 1;
+}
+
+service KeyService {
+	rpc Encrypt (EncryptRequest) returns (EncryptResponse) {}
+	rpc Decrypt (DecryptRequest) returns (DecryptResponse) {}
+}
```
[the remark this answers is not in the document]
autrilla: Fixed

jvehent:
I am using this version of Sops on my workstation, and access to KMS encrypted files (no PGP) no longer works. KMS decryption works, but Sops continues to the next decryption method, as if it was trying to reach a shamir quorum, but those files don't use key groups.

autrilla:
Fixed in 3de6531eba920f34fe514262e3d5e3997b7d3b65. Includes functional test for the bug.

cmd/sops/main.go
```diff
@@ -286,35 +443,41 @@ func main() {
...
+		}
+		addr := url.Host
+		if url.Scheme == "unix" {
+			addr = url.Path
 		}
-		cmd = exec.Command(strings.Split(string(out), "\n")[0], path)
-	} else {
-		parts, err := shlex.Split(editor)
+		opts := []grpc.DialOption{
+			grpc.WithInsecure(),
```
milescrabill: is there a reason we're always using WithInsecure?

cmd/sops/rotate.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, rmKey := range opts.RemoveMasterKeys {
+		for i := range tree.Metadata.KeyGroups {
+			for j, groupKey := range tree.Metadata.KeyGroups[0] {
```
milescrabill: this doesn't appear to be fixed per above comment, need to iterate over tree.Metadata.Keygroups[i] not [0]

sops.go
```diff
@@ -318,142 +320,132 @@ type Store interface {
...
+		// If there's only one key group, we can't do Shamir. All keys
+		// in the group encrypt the whole data key.
+		parts = append(parts, dataKey)
+	} else {
+		var err error
+		if m.ShamirQuorum == 0 {
+			m.ShamirQuorum = len(m.KeyGroups)
+		}
+		log.Printf("Multiple KeyGroups found, proceeding with Shamir with quorum %d", m.ShamirQuorum)
+		parts, err = shamir.Split(dataKey, len(m.KeyGroups), int(m.ShamirQuorum))
```
milescrabill: shouldn't this be an int already?

cmd/sops/main.go
autrilla: I think adding support for loading TLS client and server certificates for this would end up requiring significant changes to the config. Currently we just suggest using an SSH tunnel to secure the connection. This is mostly a case of us not needing TLS on the connection yet.

sops.go
autrilla: Yup. I ran [unconvert](https://github.com/mdempsky/unconvert) to make sure there aren't any other cases of this, and this is the only one in the codebase.

cmd/sops/rotate.go
autrilla: Fixed, nice catch

cmd/sops/main.go
milescrabill: OK, as long as it's thoroughly documented that this is the case it's ok.

autrilla: I've added a warning for this to the readme: https://github.com/mozilla/sops/pull/238/files#diff-88b99bb28683bd5b7e3a204826ead112R455

cmd/sops/common/common.go
```diff
@@ -0,0 +1,84 @@
...
+		return cli.NewExitError(fmt.Sprintf("Could not encrypt MAC: %s", err), codes.ErrorEncryptingMac)
+	}
+	opts.Tree.Metadata.MessageAuthenticationCode = mac
+	return nil
+}
+
+func LoadEncryptedFile(inputStore sops.Store, inputPath string) (*sops.Tree, error) {
+	fileBytes, err := ioutil.ReadFile(inputPath)
+	if err != nil {
+		return nil, cli.NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
```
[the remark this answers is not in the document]
jvehent: Something along those lines https://github.com/upspin/upspin/blob/master/errors/errors.go#L85

cmd/sops/decrypt.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+
+	"go.mozilla.org/sops"
+	"go.mozilla.org/sops/cmd/sops/codes"
+	"go.mozilla.org/sops/cmd/sops/common"
+	"go.mozilla.org/sops/keyservice"
+	"gopkg.in/urfave/cli.v1"
+)
+
+type DecryptOpts struct {
```
jvehent: I'm assuming we don't need to export those, since this is a `main` package. Keeping them private might be cleaner.

shamir/shamir.go
jvehent: I don't think we can because we don't want to vendor the entire `vault` codebase, which is fairly massive.

cmd/sops/edit.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/google/shlex"
+	"go.mozilla.org/sops"
+	"go.mozilla.org/sops/cmd/sops/codes"
+	"go.mozilla.org/sops/cmd/sops/common"
+	"go.mozilla.org/sops/keyservice"
+	"go.mozilla.org/sops/stores/json"
+	"gopkg.in/urfave/cli.v1"
+)
+
+type EditOpts struct {
```
jvehent: Again, keep you private structs private, for clarity.

jvehent:
The entire `sops` codebase needs to go through GoLint, as there are many missing comments on public/exported functions and structs. Generally speaking, the comment-to-code in this patch (and, I'm assuming, the entire codebase) is way too low for anyone else to efficiently contribute to the project.

cmd/sops/common/common.go
```diff
@@ -0,0 +1,84 @@
+package common
```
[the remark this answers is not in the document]
autrilla: This had to be moved to a different package to prevent import cycles. I'd want this to be in `cmd/sops`, but it can't be, because for example `cmd/sops/subcommand/edit` uses this, and `cmd/sops` uses `cmd/sops/subcommand/edit`.

cmd/sops/common/common.go
```diff
@@ -0,0 +1,84 @@
...
+	Stash       map[string][]interface{}
+	KeyServices []keyservice.KeyServiceClient
+	IgnoreMac   bool
+	Cipher      sops.DataKeyCipher
+}
+
+func DecryptTree(opts DecryptTreeOpts) ([]byte, error) {
+	dataKey, err := opts.Tree.Metadata.GetDataKeyWithKeyServices(opts.KeyServices)
+	if err != nil {
+		return nil, cli.NewExitError(err.Error(), codes.CouldNotRetrieveKey)
```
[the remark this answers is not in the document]
autrilla: This package is part of the CLI, it's just splitting up code that was previously in one file in `cmd/sops/main.go`.

cmd/sops/codes/codes.go
```diff
@@ -0,0 +1,26 @@
...
+	ConfigFileNotFound                     int = 61
+	KeyboardInterrupt                      int = 85
+	InvalidTreePathFormat                  int = 91
+	NoFileSpecified                        int = 100
+	CouldNotRetrieveKey                    int = 128
+	NoEncryptionKeyFound                   int = 111
+	FileHasNotBeenModified                 int = 200
+	NoEditorFound                          int = 201
+	FailedToCompareVersions                int = 202
+)
```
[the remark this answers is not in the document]
autrilla: These are command line specific codes, this is still under `cmd/`. I don't think it belongs in the library part of the code.

README.rst
```diff
@@ -392,6 +392,76 @@ Example: place the following in your `~/.bashrc`
...
+that if any of the master keys is available, the file can be decrypted.
+However, it is sometimes desirable to require access to several master keys in
+order to be able to decrypt files. This can be achieved with key groups. With
+key groups, the data key is split into several parts, one for each key group.
+Each key group contains one or more master keys, and all the keys in each
+group encrypt the data key. For decryption, `quorum` parts are required in
+order to retrieve the data key, so at least one key in `quorum` groups has to
+be available.
+
+`quorum` defaults to the number of key groups.
```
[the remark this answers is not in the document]
autrilla: `By default, the quorum is set to the number of key groups. If you have three key groups configured in your Sops file, then one master key from each group will be required to decrypt the file.`

Added `and you don't override the default quorum,`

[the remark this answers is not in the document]
autrilla: How about this?

`By default, the quorum is set to the number of key groups. For example, if you
have three key groups configured in your SOPS file and you don't override the
default quorum, then one master key from each of the three groups will be
required to decrypt the file.`

README.rst
```diff
@@ -392,6 +392,76 @@ Example: place the following in your `~/.bashrc`
...
+      # First key group
+      - pgp: fingerprint1,fingerprint2
+        kms: arn1,arn2
+      # Second key group
+      - pgp: fingerprint3,fingerprint4
+        kms: arn3,arn4
+      # Third key group
+      - pgp: fingerprint5,fingerprint6
+        kms: arn5,arn6
+```
```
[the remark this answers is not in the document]
autrilla: It's not supported, it has to be passed as a command line argument (if not, it will just use the default). I assume this is something we want to support?

cmd/sops/common/common.go
```diff
@@ -0,0 +1,84 @@
...
+			if fileMac == "" {
+				fileMac = "no MAC"
+			}
+			return nil, cli.NewExitError(fmt.Sprintf("MAC mismatch. File has %s, computed %s", fileMac, computedMac), codes.MacMismatch)
+		}
+	}
+	return dataKey, nil
+}
+
+type EncryptTreeOpts struct {
```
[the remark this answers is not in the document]
autrilla: Encrypt and decrypt have some options in common, but they also have some differences (e.g. `IgnoreMac`, so I don't think we should do that. Additionally these options may diverge even more in the future.

cmd/sops/common/common.go
```diff
@@ -0,0 +1,84 @@
...
+	DataKey []byte
+}
+
+func EncryptTree(opts EncryptTreeOpts) error {
+	mac, err := opts.Tree.Encrypt(opts.DataKey, opts.Cipher, opts.Stash)
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Error encrypting tree: %s", err), codes.ErrorEncryptingTree)
+	}
+	opts.Tree.Metadata.LastModified = time.Now().UTC()
+	mac, err = opts.Cipher.Encrypt(mac, opts.DataKey, opts.Tree.Metadata.LastModified.Format(time.RFC3339), opts.Stash)
```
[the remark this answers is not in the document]
autrilla: It is right, but the argument name for the function shouldn't be path, but `additionalData`.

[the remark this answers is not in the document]
autrilla: If we want lines to be wrapped, I think we should set a line length limit, because otherwise it's hard to tell when to wrap. That line fits on my screen, so I didn't wrap it.

cmd/sops/common/common.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Could not encrypt MAC: %s", err), codes.ErrorEncryptingMac)
+	}
+	opts.Tree.Metadata.MessageAuthenticationCode = mac
```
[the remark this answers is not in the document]
autrilla: 1. The MAC returned by `Tree.Encrypt` is unencrypted. This value is encrypted, and the unencrypted value is thrown away, as it's not needed.
2. We could do this instead, but I don't see a big difference

```go
func EncryptTree(opts EncryptTreeOpts) error {
	var err error
	opts.Tree.Metadata.MessageAuthenticationCode, err = opts.Tree.Encrypt(opts.DataKey, opts.Cipher, opts.Stash)
	if err != nil {
		return cli.NewExitError(fmt.Sprintf("Error encrypting tree: %s", err), codes.ErrorEncryptingTree)
	}
	opts.Tree.Metadata.LastModified = time.Now().UTC()
	opts.Tree.Metadata.MessageAuthenticationCode, err = opts.Cipher.Encrypt(
		opts.Tree.Metadata.MessageAuthenticationCode, opts.DataKey, opts.Tree.Metadata.LastModified.Format(time.RFC3339), opts.Stash)
	if err != nil {
		return cli.NewExitError(fmt.Sprintf("Could not encrypt MAC: %s", err), codes.ErrorEncryptingMac)
	}
	return nil
}
```

autrilla:
@jvehent would you be okay with adding key group names in another PR instead of this one? I feel like adding more code on top of this one is a bad idea.

cmd/sops/common/common.go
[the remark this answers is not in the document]
autrilla: This package contains functions meant to be used by the `cmd/sops` binary. This package does not have a stable API guarantee. Because some subcommands are split under `subcommand/` packages, these functions can't be in `cmd/sops` because of import cycles. For instance, `cmd/sops/subcommand/groups` uses `LoadEncryptedFile`, and `cmd/sops` imports `cmd/sops/subcommand/groups`.

cmd/sops/codes/codes.go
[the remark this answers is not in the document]
jvehent: Then, do they even belong in their own package?

cmd/sops/common/common.go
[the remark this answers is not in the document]
jvehent: should they be under `sops/cmd/internal` ?

cmd/sops/common/common.go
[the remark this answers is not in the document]
jvehent: What I'm trying to say is that `cli` is an implementation detail of the cmd, and the sub-packaged should throw generic errors instead of going through the `cli` package

cmd/sops/common/common.go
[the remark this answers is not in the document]
jvehent: If it doesn't fit in a github review page and forces me to scroll horizontally, it should be wrapped :)

cmd/sops/common/common.go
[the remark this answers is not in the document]
jvehent: don't use variables for stuff you throw away, just use `_`.
There's isn't a huge difference with your code sample, but I think this kind of minor change actually makes it easier for future contributors to follow the code.

cmd/sops/common/common.go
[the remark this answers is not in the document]
jvehent: agreed

cmd/sops/common/common.go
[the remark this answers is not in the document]
autrilla: But it is used, it's passed to `opts.Cipher.Encrypt(mac, ...)`

README.rst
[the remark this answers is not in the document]
jvehent: I think it makes sense to have it in the configuration. It is likely we would want to set the quorum requirement per repo.

cmd/sops/common/common.go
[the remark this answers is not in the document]
jvehent: Ok, I totally missed that... Can the variables be named differently then? maybe "unencrypted_mac" for the first one?

cmd/sops/edit.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func EditExample(opts editExampleOpts) ([]byte, error) {
+	// Load the example file
+	var fileBytes []byte
+	if _, ok := opts.InputStore.(*json.BinaryStore); ok {
+		// Get the value under the first key
+		fileBytes = []byte(exampleTree[0].Value.(string))
+	} else {
+		var err error
```
jvehent: maybe declare `fileBytes` and `err` in the function signature?

cmd/sops/edit.go
```diff
@@ -0,0 +1,274 @@
...
+	dataKey, errs := tree.GenerateDataKeyWithKeyServices(opts.KeyServices)
+	if len(errs) > 0 {
+		return nil, cli.NewExitError(fmt.Sprintf("Error encrypting the data key with one or more master keys: %s", errs), codes.CouldNotRetrieveKey)
+	}
+	stash := make(map[string][]interface{})
+
+	return edit(opts.editOpts, &tree, dataKey, stash)
+}
+
+func Edit(opts editOpts) ([]byte, error) {
```
jvehent: These are newly exported functions, and I'm not clear on the value of exporting them. Care to clarify?

cmd/sops/rotate.go
```diff
@@ -0,0 +1,74 @@
...
+	// Remove master keys
+	for _, rmKey := range opts.RemoveMasterKeys {
+		for i := range tree.Metadata.KeyGroups {
+			for j, groupKey := range tree.Metadata.KeyGroups[i] {
+				if rmKey.ToString() == groupKey.ToString() {
+					tree.Metadata.KeyGroups[i] = append(tree.Metadata.KeyGroups[i][:j], tree.Metadata.KeyGroups[i][j+1:]...)
+				}
+			}
+		}
+	}
```
jvehent: Should there be a check to warn the user in case remove keys makes it impossible to satisfy the quorum? A sort of `footgun prevention` feature.

cmd/sops/subcommand/groups/add.go
```diff
@@ -0,0 +1,51 @@
+package groups
```
jvehent: nit: do we need a `subcommand` directory vs putting `groups` directly under `cmd/sops` ?

cmd/sops/subcommand/groups/delete.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return err
+	}
+	tree.Metadata.KeyGroups = append(tree.Metadata.KeyGroups[:opts.Group], tree.Metadata.KeyGroups[opts.Group+1:]...)
+
+	if opts.GroupQuorum != 0 {
+		tree.Metadata.ShamirQuorum = opts.GroupQuorum
+	}
+	// The quorum should always be smaller or equal to the number of key groups
+	tree.Metadata.ShamirQuorum = min(tree.Metadata.ShamirQuorum, len(tree.Metadata.KeyGroups))
```
jvehent: Will this silently change the quorum of a file should the user delete a keygroup that previously satisfied that quorum?
If so, I wonder if that's a good idea. I think we should make sure quorum modifications are intentional so users don't inadvertently reduce the security of their secrets files.

I would propose:
1. if deleting a key group results in a quorum that can no longer be satisfied, refuse the change and display an error
2. require users to update the quorum at the same time as deleting a key group for the operation to succeed: `sops groups --delete group1 --update-quorum 2`, or something along those lines

functional-tests/.sops.yaml
```diff
@@ -1,2 +1,10 @@
 creation_rules:
+    - filename_regex: test_roundtrip_keygroups.yaml
+      key_groups:
+      - pgp: 1022470DE3F0BC54BC6AB62DE05550BC07FB1A0A
+      - pgp: 729D26A79482B5A20DEAD0A76945978B930DD7A2
+    - filename_regex: test_roundtrip_keygroups_missing_decryption_key.yaml
+      key_groups:
+      - pgp: 1022470DE3F0BC54BC6AB62DE05550BC07FB1A0A
+      - pgp: 620B9A4C96230B91E7473D20113D2B26EA0890C7
```
jvehent: per this file structure, are we able to have a mix of pgp and kms keys in a single key group?

jvehent: Going further, since this is a new format, we can avoid repeating the mistakes of the past. I would suggest:

```yaml
creation_rules:
   - filename_regex: *\.yaml$
     key_groups:
        - name: group1
          kms:
            - arn: arn:foo:bar
              role: somerole1
            - arn: arn:foo:baz
              role: somerole2
          pgp:
            - 1022470DE3F0BC54BC6AB62DE05550BC07FB1A0A
            - 620B9A4C96230B91E7473D20113D2B26EA0890C7
        - name: group2
           pgp:
             - 729D26A79482B5A20DEAD0A76945978B930DD7A2
```

keys/keys.go
jvehent: That's a lot of interfaces to satisfy. Are they all required? Can some of them be moved to optional interfaces to make it easier to implement new master keys without requiring all the functionalities?

keyservice/keyservice.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+
+message Key {
+	oneof key_type {
+		KmsKey kms_key = 1;
+		GpgKey gpg_key = 2;
```
jvehent: nit: gpg is an implementation of pgp. I prefer to use the term `pgp` everywhere, not `gpg`. (I know, it's confusing)

keyservice/keyservice.proto
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+message GpgKey {
```
jvehent: s/gpg/pgp/

keyservice/server.go
jvehent: Is there a reason why it's implemented this way, versus returning `EncryptedKey` as the output of `kmsKey.Encrypt(plaintext)` ?

keyservice/server.go
```diff
@@ -0,0 +1,113 @@
...
+		Role:              key.Role,
+		EncryptionContext: ctx,
+	}
+	kmsKey.EncryptedKey = string(ciphertext)
+	plaintext, err := kmsKey.Decrypt()
+	return []byte(plaintext), err
+}
+
+func (ks Server) Encrypt(ctx context.Context,
+	req *EncryptRequest) (*EncryptResponse, error) {
```
jvehent: nit: why wrap function declarations lines that aren't even that long?

keyservice/server.go
```diff
@@ -0,0 +1,113 @@
...
+	key := *req.Key
+	switch k := key.KeyType.(type) {
+	case *Key_GpgKey:
```
jvehent: linters don't like underscores

jvehent:
Here's the output from golint
```
go list ./... | grep -v /vendor/ | xargs -L1 golint
sops.go:1:1: package comment should be of the form "Package sops ..."
sops.go:115:1: receiver name tree should be consistent with previous receiver name branch for TreeBranch
sops.go:145:1: receiver name tree should be consistent with previous receiver name branch for TreeBranch
sops.go:166:1: receiver name tree should be consistent with previous receiver name branch for TreeBranch
sops.go:180:1: receiver name tree should be consistent with previous receiver name branch for TreeBranch
sops.go:285:1: comment on exported method Tree.GenerateDataKeyWithKeyServices should be of the form "GenerateDataKeyWithKeyServices ..."
sops.go:309:6: exported type KeyGroup should have comment or be unexported
sops.go:329:1: exported method Metadata.UpdateMasterKeysWithKeyServices should have comment or be unexported
sops.go:332:15: error strings should not be capitalized or end with punctuation or a newline
sops.go:368:43: error strings should not be capitalized or end with punctuation or a newline
sops.go:426:27: error strings should not be capitalized or end with punctuation or a newline
/home/ulfr/src/go.mozilla.org/sops/cmd/sops/rotate.go:46:1: expected statement, found '<<' (and 3 more errors)
cmd/sops/codes/codes.go:4:2: exported const CouldNotReadInputFile should have comment (or a comment on this block) or be unexported
cmd/sops/common/common.go:15:6: exported type DecryptTreeOpts should have comment or be unexported
cmd/sops/common/common.go:23:1: exported function DecryptTree should have comment or be unexported
cmd/sops/common/common.go:45:6: exported type EncryptTreeOpts should have comment or be unexported
cmd/sops/common/common.go:52:1: exported function EncryptTree should have comment or be unexported
cmd/sops/common/common.go:66:1: exported function LoadEncryptedFile should have comment or be unexported
cmd/sops/subcommand/groups/add.go:11:6: exported type AddOpts should have comment or be unexported
cmd/sops/subcommand/groups/add.go:21:1: exported function Add should have comment or be unexported
cmd/sops/subcommand/groups/add.go:40:17: should omit type *os.File from declaration of var outputFile; it will be inferred from the right-hand side
cmd/sops/subcommand/groups/delete.go:11:6: exported type DeleteOpts should have comment or be unexported
cmd/sops/subcommand/groups/delete.go:28:1: exported function Delete should have comment or be unexported
cmd/sops/subcommand/groups/delete.go:50:17: should omit type *os.File from declaration of var outputFile; it will be inferred from the right-hand side
cmd/sops/subcommand/keyservice/keyservice.go:15:6: exported type Opts should have comment or be unexported
cmd/sops/subcommand/keyservice/keyservice.go:20:1: exported function Run should have comment or be unexported
keyservice/client.go:9:6: exported type LocalClient should have comment or be unexported
keyservice/client.go:13:1: exported function NewLocalClient should have comment or be unexported
keyservice/client.go:17:1: exported method LocalClient.Decrypt should have comment or be unexported
keyservice/client.go:22:1: exported method LocalClient.Encrypt should have comment or be unexported
keyservice/keyservice.go:11:1: exported function KeyFromMasterKey should have comment or be unexported
keyservice/server.go:11:6: exported type Server should have comment or be unexported
keyservice/server.go:61:1: exported method Server.Encrypt should have comment or be unexported
keyservice/server.go:88:1: exported method Server.Decrypt should have comment or be unexported
kms/keysource.go:36:1: exported method MasterKey.EncryptedDataKey should have comment or be unexported
kms/keysource.go:40:1: exported method MasterKey.SetEncryptedDataKey should have comment or be unexported
pgp/keysource.go:31:1: exported method MasterKey.EncryptedDataKey should have comment or be unexported
pgp/keysource.go:35:1: exported method MasterKey.SetEncryptedDataKey should have comment or be unexported
pgp/keysource.go:80:21: error strings should not be capitalized or end with punctuation or a newline
shamir/shamir.go:61:28: don't use underscores in Go names; func parameter x_samples should be xSamples
shamir/shamir.go:61:39: don't use underscores in Go names; func parameter y_samples should be ySamples
shamir/shamir.go:90:2: don't use underscores in Go names; var log_a should be logA
shamir/shamir.go:91:2: don't use underscores in Go names; var log_b should be logB
shamir/shamir.go:114:2: don't use underscores in Go names; var log_a should be logA
shamir/shamir.go:115:2: don't use underscores in Go names; var log_b should be logB
shamir/shamir.go:231:2: don't use underscores in Go names; var x_samples should be xSamples
shamir/shamir.go:232:2: don't use underscores in Go names; var y_samples should be ySamples
shamir/shamir_test.go:191:3: don't use underscores in Go names; var x_vals should be xVals
shamir/shamir_test.go:192:3: don't use underscores in Go names; var y_vals should be yVals
stores/stores.go:13:6: exported type SopsFile should have comment or be unexported
stores/stores.go:18:1: comment on exported type Metadata should be of the form "Metadata ..." (with optional leading article)
stores/stores.go:52:1: exported function MetadataFromInternal should have comment or be unexported
stores/stores.go:104:1: exported method Metadata.ToInternal should have comment or be unexported
stores/yaml/store.go:137:1: comment on exported method Store.MarshalValue should be of the form "MarshalValue ..."
stores/yaml/store.go:144:1: receiver name s should be consistent with previous receiver name store for Store
yaml/config.go:74:1: exported function KeyGroupsForFile should have comment or be unexported
```

keyservice/server.go
```diff
@@ -0,0 +1,113 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return &EncryptResponse{
+			Ciphertext: ciphertext,
+		}, nil
+	case nil:
+		return nil, grpc.Errorf(codes.NotFound, "Must provide a key")
+	default:
+		return nil, grpc.Errorf(codes.NotFound, "Unknown key type")
```
jvehent: Shouldn't these be two different codes?

keyservice/server.go
```diff
@@ -0,0 +1,113 @@
...
+	case *Key_GpgKey:
+		plaintext, err := ks.decryptWithPgp(k.GpgKey, req.Ciphertext)
+		if err != nil {
+			return nil, err
+		}
+		return &DecryptResponse{
+			Plaintext: plaintext,
+		}, nil
+	case *Key_KmsKey:
+		plaintext, err := ks.decryptWithKms(k.KmsKey, req.Ciphertext)
```
jvehent: Instead of having all those `encryptwithX` and `decryptwithY`, could we have two interfaces that master keys have to satisfy that could simply be used in a generic way to encrypt and decrypt a data key? That would prevent having to add master-key specific code everytime we add a new key type.

kms/keysource.go
```diff
@@ -60,23 +74,28 @@ func (key *MasterKey) EncryptIfNeeded(dataKey []byte) error {
 
 // Decrypt decrypts the EncryptedKey field with AWS KMS and returns the result.
 func (key *MasterKey) Decrypt() ([]byte, error) {
+	log.Printf("Attempting decryption of KMS MasterKey with ARN %s", key.Arn)
```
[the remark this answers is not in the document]
jvehent: +1
I think the default logging mode should be fairly silent, and users would use `-v` and `-vv` to increase log verbosity, similar to most unix tools.

kms/keysource.go
```diff
@@ -220,7 +239,7 @@ func ParseKMSContext(in interface{}) map[string]*string {
 		for _, kv := range strings.Split(in, ",") {
 			kv := strings.Split(kv, ":")
 			if len(kv) != 2 {
-				fmt.Printf("[WARNING]: KMS Encryption Context could not be parsed, context will not be used")
+				log.Printf("[WARNING]: KMS Encryption Context could not be parsed, context will not be used")
```
jvehent: nit: don't put a colon after brackets, it's redundant

pgp/keysource.go
```diff
@@ -64,6 +108,26 @@ func (key *MasterKey) Encrypt(dataKey []byte) error {
...
+		log.Printf("Encryption of GPG MasterKey with fingerprint %s succeeded", key.Fingerprint)
+		return nil
+	}
+	log.Print("Encryption with golang's openpgp package failed, falling back to the GPG binary")
+	binaryErr := key.encryptWithGPGBinary(dataKey)
+	if binaryErr == nil {
+		log.Printf("Encryption of GPG MasterKey with fingerprint %s succeeded", key.Fingerprint)
+		return nil
+	}
+	log.Printf("Encryption of GPG MasterKey with fingerprint %s failed", key.Fingerprint)
```
jvehent: s/GPG/PGP/

pgp/keysource.go
```diff
@@ -64,6 +108,26 @@ func (key *MasterKey) Encrypt(dataKey []byte) error {
...
+	return fmt.Errorf(`could not encrypt data key with PGP key.
+	\tgolang.org/x/crypto/openpgp error: %s
+	\tGPG binary error: %s`, openpgpErr, binaryErr)
```
jvehent: should we return multiple error strings?

pgp/keysource.go
```diff
@@ -87,11 +167,32 @@ func (key *MasterKey) Decrypt() ([]byte, error) {
...
+		log.Printf("Decryption of GPG MasterKey with fingerprint %s succeeded", key.Fingerprint)
+		return dataKey, nil
+	}
+	log.Print("Decryption with golang's openpgp package failed, falling back to the GPG binary")
+	dataKey, binaryErr := key.decryptWithGPGBinary()
+	if binaryErr == nil {
+		log.Printf("Decryption of GPG MasterKey with fingerprint %s succeeded", key.Fingerprint)
+		return dataKey, nil
+	}
+	log.Printf("Decryption of GPG MasterKey with fingerprint %s failed", key.Fingerprint)
```
jvehent: s/GPG/PGP/

pgp/keysource.go
```diff
@@ -169,8 +270,8 @@ func (key *MasterKey) fingerprintMap(ring openpgp.EntityList) map[string]openpgp
 func (key *MasterKey) passphrasePrompt(keys []openpgp.Key, symmetric bool) ([]byte, error) {
 	conn, err := gpgagent.NewConn()
 	if err == gpgagent.ErrNoAgent {
-		fmt.Println("gpg-agent not found, continuing with manual passphrase input...")
-		fmt.Print("Enter PGP key passphrase: ")
+		log.Printf("gpg-agent not found, continuing with manual passphrase input...")
+		log.Print("Enter PGP key passphrase: ")
```
jvehent: this isn't strictly speaking a log, it's a prompt

yaml/config.go
jvehent: See my previous comment on the structure of the keygroup configuration

jvehent:
Reading a bit more about `sss`, it seems the term `quorum` is never used but `threshold` is used instead. I wonder if we want to use `threshold` as well and be consistent with crypto terminology.

cmd/sops/edit.go
```diff
@@ -0,0 +1,274 @@
...
+		Key:   "example_number",
+		Value: 1234.56789,
+	},
+	sops.TreeItem{
+		Key:   "example_booleans",
+		Value: []interface{}{true, false},
+	},
+}
+
+type runEditorUntilOkOpts struct {
```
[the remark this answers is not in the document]
autrilla: Why? This isn't really the state

functional-tests/.sops.yaml
autrilla: Yes, a mix of pgp and kms keys is allowed.

We can do what you suggest for the syntax inside kms and pgp entries in each key group, but it might be a bit confusing to have two different syntaxes, one for outside key groups and one for inside.

keyservice/server.go
autrilla: We don't control the underscore in this name, protobuf does this

keyservice/server.go
autrilla: I don't know. Both are a "not found" to me, one is "no key provided" and the other is "key type no found in our records". Do you think there's a better code for this?

keyservice/server.go
autrilla: This is not really possible because we don't control the implementation of KmsKey and PgpKey, protobuf does.

kms/keysource.go
autrilla: There's a different branch were I'm working on logging improvements. This branch uses logrus and log levels. Should I merge that branch into this one?

pgp/keysource.go
autrilla: Yes, I fixed that in my logging branch.

pgp/keysource.go
autrilla: I suppose it'd be even better to define our own error type containing the two possible errors with an `Error() string` function.

keyservice/keyservice.proto
autrilla: I've replaced all usages of GPG to PGP, except when it specifically refers to the GPG binary

autrilla:
Yes, I think threshold is better. I'll replace the name.

keyservice/server.go
autrilla: Yes. The encrypted data key has to be in the MasterKey, because we have to keep track of what MasterKey encrypted the data key. The encrypted data key has to end up in the output file, so we need to keep that information around, somehow. I don't disagree that this being part of the MasterKey interface is bad, but this kind of was the way it worked before with `ToMap` (which we could remove now). This should probably go in another PR and not in this one.

cmd/sops/codes/codes.go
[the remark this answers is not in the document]
autrilla: They have to be, otherwise there's import cycles

cmd/sops/subcommand/groups/delete.go
autrilla: Yeah, I think this might be better security-wise. The actual command is `sops groups delete --file file.yaml --shamir-secret-sharing-threshold 2 1`

functional-tests/.sops.yaml
autrilla: Currently we specify the role inside the ARN. I assume we want to keep doing that?

autrilla: Yeah, from the README:

You can specify a role in the --kms flag and SOPS_KMS_ARN variable by appending it to the ARN of the master key, separated by a + sign:
```
<KMS ARN>+<ROLE ARN>
arn:aws:kms:us-west-2:927034868273:key/fe86dd69-4132-404c-ab86-4269956b4500+arn:aws:iam::927034868273:role/sops-dev-xyz
```

I've made a change so it can be specified like you said. I've also updated the README, and the example in it looks like this:

```
creation_rules:
    - filename_regex: .*keygroups.*
      key_groups:
      # First key group
      - pgp:
        - fingerprint1
        - fingerprint2
        kms:
        - arn: arn1
          role: role1
          context:
            foo: bar
        - arn: arn2
      # Second key group
      - pgp:
        - fingerprint3
        - fingerprint4
        kms:
        - arn: arn3
        - arn: arn4
      # Third key group
      - pgp:
        - fingerprint5
```

cmd/sops/rotate.go
autrilla: Rotate doesn't add or remove groups, only keys, and in the case where the user would remove all keys from a group, SOPS would error anyway. I don't think this is an issue in rotate, but it certainly is in `sops groups delete`, as you pointed out.

PR #250, opened 2017-09-16
Sops3

jvehent:
To be merge when ready to release.


CHANGELOG.rst
```diff
@@ -1,6 +1,25 @@
...
+  keys to access a data key and decrypt a file. See `sops groups -help`.
+
+* Keyservice to forward access to a local master key on a socket, similar to
+  gpg-agent. See `sops keyservice --help`.
+
+* Encrypt comments by default
+
+* Support for Google Compute Platform KMS
+
+* Refactor of the store logic to separate the internal representation SOPS
```
autrilla: Do users care about this?

CHANGELOG.rst
```diff
@@ -1,6 +1,25 @@
...
+  has of files from the external representation used in JSON and YAML files
+
+* Reencoding of versions as string on sops 1.X files, may break backward
+  compatibility but will be handled automatically.
```
autrilla: It's not handled automatically, but the following error message is shown: `Error loading file metadata: Error unmarshalling input json: json: cannot unmarshal number into Go struct field Metadata.version of type string`

CHANGELOG.rst
jvehent: users, no, but they also don't read changelogs. developers may care, and they are the target audience.

CHANGELOG.rst
jvehent: What should users who see this message do?

autrilla: They have two options:

- Rotate the file with any version of SOPS `2.x.x`, which will make `Metadata.version` a string.
- Manually edit the JSON file to ensure the entry under `sops -> version` is a string and not a number.

autrilla:
Yes, I've opened #251 for it.

PR #251, opened 2017-09-16
Show error message when decrypting SOPS1 JSON files

autrilla:
This shows the following:

```
❯ sops -d example.json
Error loading file metadata: SOPS versions higher than 2.0.10 can not automatically decrypt JSON files created with SOPS 1.x. In order to be able to decrypt this file, you can either edit it manually and make sure the JSON value under `sops -> version` is a string and not a number, or you can rotate the file with any version of SOPS between 2.0 and 2.0.10
```

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -252,6 +252,16 @@ func (store Store) UnmarshalMetadata(in []byte) (sops.Metadata, error) {
 	file := stores.SopsFile{}
 	err := json.Unmarshal(in, &file)
 	if err != nil {
+		if err, ok := err.(*json.UnmarshalTypeError); ok {
+			if err.Value == "number" && err.Struct == "Metadata" && err.Field == "version" {
+				return sops.Metadata{},
+					fmt.Errorf("SOPS versions higher than 2.0.10 can not automatically decrypt JSON files " +
+						"created with SOPS 1.x. In order to be able to decrypt this file, you can either edit it " +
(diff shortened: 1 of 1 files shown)


stores/json/store.go
```diff
@@ -252,6 +252,15 @@ func (store Store) UnmarshalMetadata(in []byte) (sops.Metadata, error) {
 	file := stores.SopsFile{}
 	err := json.Unmarshal(in, &file)
 	if err != nil {
+		if err, ok := err.(*json.UnmarshalTypeError); ok {
+			if err.Value == "number" && err.Struct == "Metadata" && err.Field == "version" {
```
jvehent: Nice!

stores/json/store.go
```diff
@@ -252,6 +252,15 @@ func (store Store) UnmarshalMetadata(in []byte) (sops.Metadata, error) {
...
+				return sops.Metadata{},
+					fmt.Errorf("SOPS versions higher than 2.0.10 can not automatically decrypt JSON files " +
+						"created with SOPS 1.x. In order to be able to decrypt this file, you can either edit it " +
+						"manually and make sure the JSON value under `sops -> version` is a string and not a " +
+						"number, or you can rotate the file with any version of SOPS between 2.0 and 2.0.10")
```
jvehent: `you can rotate the key with "-r"`

PR #252, opened 2017-09-16
Fixed bug when encrypting nil values

### LINKED ISSUE #171: YAML store is unable to handle nil values
Passing the following YAML document to sops makes it crash:

```
foo: ~
```

JCepedaVillamayor:
Fixes #171 

The null value is not encrypted since it does not make sense to encrypt it.
Python sops converts the nil value into an empty string e.g:
Original YAML file:
`foo: ~`
Result of a round-trip with Python sops:
`foo: ''`
Result of a round-trip using Go sops:
`foo: null`
This PR includes a test for this behaviour


sops_test.go
```diff
@@ -116,6 +116,10 @@ func TestEncrypt(t *testing.T) {
 			Key:   "foobar",
 			Value: 2.12,
 		},
+		TreeItem{
```
autrilla: Could you add the same thing to `TestDecrypt`? We should test that it works both ways.

JCepedaVillamayor: Added! Now it tests the value when is decrypting

PR #253, opened 2017-09-17
Improve logging

autrilla:
With log levels, logger names, and structured logging.


autrilla:
There's no way in this patch for the user to select the log level from the command line, and I'm not sure if we'd even want this -- would we just set all loggers to the level the user provides? Do we want a way for users to configure a level for each logger?

PR #255, opened 2017-09-25
Add command to reconfigure keys

autrilla:
Currently there's no easy way to add or remove keys from files en masse. This adds a way to update the keys in all files in a directory (or also for a single file) using the `.sops.yaml` config file. It prompts the user to confirm the change to every file (unless the `-y` flag is provided), and shows a diff so the user can see what changes will be made.


cmd/sops/main.go
```diff
@@ -211,6 +213,33 @@ func main() {
 				},
 			},
 		},
+		{
+			Name:  "updatekeys",
+			Usage: "update the keys of a SOPS file using the config file",
+			Flags: append([]cli.Flag{
+				cli.StringFlag{
+					Name:  "file, f",
+					Usage: `the file or directory that should be updated with the new keys. If providing a directory, SOPS will walk it and update all files in all subdirectories. '.git' directories and '.sops.yaml' files are ignored.`,
```
jvehent: Using a flag is different from the regular sops command line that just takes the file from the first non-flag argument. I think we should preserve it. 

As far as passing a whole directory, I wonder if that's needed, because ops can always use something along the line of `find . -name '*.yaml' -exec sops updatekeys {} \;`

cmd/sops/main.go
```diff
@@ -211,6 +213,33 @@ func main() {
...
+				},
+				cli.BoolFlag{
+					Name:  "yes, y",
+					Usage: `automatic yes to prompts. Assume "yes" as answer to all prompts and run non-interactively.`,
```
jvehent: `pre-approve all changes and run non-interactively`

cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -0,0 +1,198 @@
+package updatekeys
+
+import (
+	"fmt"
+
```
jvehent: nit: gofmt

cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+	if info.IsDir() {
+		filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+			if info.IsDir() {
+				return nil
+			}
+			ignoredPaths := []string{
+				"/.git/",
```
jvehent: As a general rule, I don't like hardcoding exceptions. I'm also not sure that's needed because `.git` store binary blobs, not sops file, and `.sops.yaml` doesn't have sops metadata.

cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -0,0 +1,198 @@
...
+				".sops.yaml",
+			}
+			for _, ignoredPath := range ignoredPaths {
+				if strings.Contains(path, ignoredPath) {
```
jvehent: this will trigger on `foo.sops.yaml` which isn't a sops configuration file

cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+			return nil
+		})
+	} else {
+		syncFile(opts)
+	}
+	return nil
+}
+
+func syncFile(opts Opts) error {
```
jvehent: `sync` defines a specific operation (move a state from a source to a destination) that is different from what you're doing here. I'd suggest `update` or `modify`.

cmd/sops/subcommand/updatekeys/updatekeys.go
autrilla: The reasoning for this was to eliminate hundreds of error messages when running it, but if we go the route of just accepting a single file, it won't be needed at all.

cmd/sops/subcommand/updatekeys/updatekeys.go
autrilla: Hmm, I don't know how this happens, but `goimports` doesn't seem to care about it. I'll fix it manually.

autrilla:
All comments should be addressed now.

- Removed the "recursiveness".
- Updated the usage description of `-y`
- Use arguments instead of flags for the file to update
- Don't ignore files containing `.git` or `.sops.yaml`.

PR #256, opened 2017-09-28
Auditing support

autrilla:
An example audit log entry:

```
sops=# SELECT * FROM decrypt_event;
 id |         timestamp          | username |                               file                               
----+----------------------------+----------+------------------------------------------------------------------
  1 | 2017-09-28 17:56:42.842362 | autrilla | /Users/autrilla/Projects/go/src/go.mozilla.org/sops/example.json
(1 row)
```

More info about this can be found in the docs in the README


autrilla:
Tests fail due to missing deps

audit/schema.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE decrypt_event (
```
jvehent: would we need different tables for encrypt and decrypt, or just put everything into a single table named "events" ?
from a querying perspective, I think a single table would be fine, and just store the `action-{encrypt|decrypt}` as a column.

audit/schema.sql
```diff
@@ -0,0 +1,13 @@
...
+  username TEXT,
+  file TEXT
+);
+
+CREATE RULE decrypt_event_delete_protection AS ON DELETE TO decrypt_event DO INSTEAD NOTHING;
+
+CREATE ROLE sops;
+ALTER ROLE sops WITH NOSUPERUSER INHERIT NOCREATEROLE NOCREATEDB LOGIN PASSWORD 'sops';
+GRANT INSERT ON decrypt_event TO sops;
+GRANT USAGE ON decrypt_event_id_seq TO sops;
```
jvehent: naice!

audit/audit.go
```diff
@@ -0,0 +1,81 @@
...
+
+func Register(auditor Auditor) {
+	auditors = append(auditors, auditor)
+}
+
+type Auditor interface {
+	Handle(event interface{})
+}
+
+type DecryptEvent struct {
```
jvehent: should the event struct be tied to `Decrypt` or generic?

audit/audit.go
```diff
@@ -0,0 +1,81 @@
...
+
+func init() {
+	log = logging.NewLogger("AUDIT")
+}
+
+var auditors []Auditor
+
+func SubmitEvent(event interface{}) {
+	for _, auditor := range auditors {
+		auditor.Handle(event)
```
jvehent: this is nice, it'll let us broadcast events to other destinations later on, like email or irc

README.rst
```diff
@@ -583,6 +583,33 @@ in `/tmp/sops.sock` and not the local key service, you can run:
...
+~~~~~~~~
+
+Sometimes, users want to be able to tell what files were accessed by whom in an
+environment they control. For this reason, SOPS includes a way to log the
+decryption of files to PostgreSQL. The log includes a timestamp, the username
+SOPS is running as, and the file that was decrypted.
+
+Because we don't want users of SOPS to be able to control auditing, the audit
+configuration is baked in statically in the binary at compile time using Go's
+linker -X ldflag. Currently, there's two configurable options:
```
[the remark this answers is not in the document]
autrilla: But then we can't allow users to override the path for that config file. I don't have a strong opinion about this, both approaches have advantages and disadvantages.

audit/schema.sql
autrilla: I'm not sure if there's any practical difference. I think one table per event type is better because different events might include other data. If you need to query everything, you can always just join several tables.

audit/audit.go
autrilla: Same as the database, this is just future proofing. Other events might need other fields.

audit/audit.go
```diff
@@ -0,0 +1,81 @@
...
+		DB: db,
+	}
+}
+
+func (p *PostgresAuditor) Handle(event interface{}) {
+	switch event := event.(type) {
+	case DecryptEvent:
+		// Save the event to the database
+		log.WithField("file", event.File).
+			Info("Saving decrypt event to database")
```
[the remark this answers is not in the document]
autrilla: I don't see why not. Maybe this should be debug level and not info, though.

autrilla:
I've addressed the comments, except the ones I disagreed with (having shared tables/structs for all events).

README.rst
```diff
@@ -583,6 +583,46 @@ in `/tmp/sops.sock` and not the local key service, you can run:
...
+
+For example, to enable auditing to a PostgreSQL database named :code:`sops`
+running on localhost, using the user :code:`sops` and the password :code:`sops`,
+:code:`/etc/sops/audit.yaml` should have the following contents:
+
+.. code:: yaml
+
+    backends:
+        postgres:
+            - connection_string: "postgres://sops:sops@localhost/sops?sslmode=disable"
```
jvehent: I don't think we should have `sslmode=disable` in an example config. We should also link to whatever upstream doc explains the postgres URI format.

README.rst
```diff
@@ -583,6 +583,46 @@ in `/tmp/sops.sock` and not the local key service, you can run:
...
+            - connection_string: "postgres://sops:sops@remotehost/sops?sslmode=verify-full"
+
+The schema your database should have can be found at :code:`audit/schema.sql`.
+This schema defines the tables that store the audit events and a role named
+:code:`sops` that only has permission to add entries to the audit event tables.
```
jvehent: Maybe rephrase as a tutorial?
1. create the database and credentials using the schema
2. configure `audit.yaml` to allow sops to write into the db

README.rst
autrilla: Yeah, fixed.

README.rst
autrilla: Let me know what you think of the new phrasing.

audit/audit.go
```diff
@@ -0,0 +1,111 @@
...
+	// restructuring SOPS to use dependency injection instead of just using
+	// globals everywhere.
+	if flag.Lookup("test.v") == nil {
+		for _, pgConf := range conf.Backends.Postgres {
+			auditors = append(auditors, NewPostgresAuditor(pgConf.ConnStr))
+		}
+	}
+}
+
+const configFile = "/etc/sops/audit.yaml"
```
jvehent: This could probably be platform specific for folks running in windows? unsure if that's worth the effort...

PR #259, opened 2017-10-03
Create slim dockerfile

autrilla:
Useful as a simple way to get a SOPS binary. Users can then mount whichever secrets they want into the container (and potentially a key service socket) and decrypt/encrypt from inside.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
       - rpm
       - ruby
       - python3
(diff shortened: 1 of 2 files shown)


autrilla:
@jvehent can you configure Mozilla's Dockerhub credentials on Travis-CI? They need to be in the `DOCKER_USERNAME` and `DOCKER_PASSWORD` env vars.

PR #260, opened 2017-10-03
Fix metadata not found returning the wrong error

autrilla:
Contains tests

### DIFF
--- a/stores/stores.go
+++ b/stores/stores.go
@@ -22,7 +22,11 @@ import (
 
 // SopsFile is a struct used by the stores as a helper to unmarshal the SOPS metadata
 type SopsFile struct {
-	Metadata Metadata `yaml:"sops" json:"sops"`
+	// Metadata is a pointer so we can easily tell when the field is not present
+	// in the SOPS file by checking for nil. This way we can show the user a
+	// helpful error message indicating that the metadata wasn't found, instead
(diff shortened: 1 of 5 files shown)


autrilla:
Not sure why codecov is saying this decreases coverage, will investigate.

stores/stores.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // SopsFile is a struct used by the stores as a helper to unmarshal the SOPS metadata
 type SopsFile struct {
-	Metadata Metadata `yaml:"sops" json:"sops"`
+	Metadata *Metadata `yaml:"sops" json:"sops"`
```
jvehent: uh?

autrilla: So the JSON and YAML parsers will default the field to the zero value if it's not in the YAML/JSON document. Using a pointer is cleaner imo because it will leave it nil, whereas if we don't use a pointer it'll leave it to the zero value of `Metadata`.

But we could also leave it as-is and then do `if file.Metadata == Metadata{} { ... }`.

jvehent: I'm fine with the change, but it raised an eyebrow so I think you should put that explanation in a comment above, so we don't wonder why it's done this way next time around.

jvehent:
travisci is slow and this looks like a basic change so
yolo

PR #263, opened 2017-10-05
Fix some errors not being shown and erroring when no config file is found

### LINKED ISSUE #262: sops encrypt not returning anything
I installed sops on MacOS using

`$ go get -u go.mozilla.org/sops/cmd/sops`

I exported the  **SOPS_KMS_ARN** env variable and my AWS credentials are present in ~/.aws/credentials.

When I run `sops -e my-file.yml` nothing happens and I get no feedback.

I cloned the repo and ran the same command using sops.go and this time it worked.

Any idea of what might be happening ?

autrilla:
First of all, urfave/cli doesn't print errors unless they're a `cli.ExitError`, which causes some errors to not be shown to the user. I've created a `toExitError` function that converts the error to a `cli.ExitError` if it isn't one.

Secondly, there were some cases of the `err` variable being overwritten by branches in main.go, which caused the error to also not be printed.

Third, the `--config` flag was not being used, and getting the Shamir threshold errored if no `.sops.yaml` config file was found.

Fixes #262

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -367,13 +367,15 @@ func main() {
 		var output []byte
 		var err error
 		if c.Bool("encrypt") {
-			keyGroups, err := keyGroups(c, fileName)
+			var groups []sops.KeyGroup
+			groups, err = keyGroups(c, fileName)
 			if err != nil {
-				return err
+				return toExitError(err)
 			}
-			shamirThreshold, err := shamirThreshold(c, fileName)
+			var threshold int
+			threshold, err = shamirThreshold(c, fileName)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -367,13 +367,15 @@ func main() {
 		var output []byte
 		var err error
 		if c.Bool("encrypt") {
-			keyGroups, err := keyGroups(c, fileName)
+			var groups []sops.KeyGroup
+			groups, err = keyGroups(c, fileName)
```
jvehent: why not use
```go
groups, err := keyGroups(c, fileName)
```

PR #268, opened 2017-10-12
Circleci

jvehent:


.circleci/config.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  build:
+    working_directory: /go/src/github.com/mozilla/sops
+    docker:
+      - image: circleci/golang:1.8
+    steps:
+      - checkout
+      - run:
+          name: Build Sops
+          command: make
```
autrilla: sops is built inside the container so we don't really need this, but it doesn't hurt

PR #269, opened 2017-10-13
Use go.mozilla.org package name in CircleCI config

autrilla:
Also updates the Docker image to include the AWS cli and use a newer version of Go

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,20 +1,20 @@
 version: 2
 jobs:
   build:
-    working_directory: /go/src/github.com/mozilla/sops
+    working_directory: /go/src/go.mozilla.org/sops
     docker:
       - image: circleci/golang:1.8
     steps:
       - checkout
+      - setup_remote_docker
       - run:
           name: Build Sops
-          command: make 
-      - setup_remote_docker                                                                                                                                                                                                                                    
-      - run:                                                                                                                                                                                                                                                   
-          name: Build containers                                                                                                                                                                                                                               
-          command: |                                                                                                                                                                                                                                           
-            docker build -t mozilla/sops .                                                                                                                                                                                                          
-            docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"                                                                                                                                                                          
+          command: make
+      - run:
+          name: Build containers
+          command: |
+            docker build -t mozilla/sops .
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,25 +1,21 @@
...
+          name: Build containers
+          command: |
+            docker build -t mozilla/sops .
+            docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"
       - run:
           name: Push containers
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
                 ./bin/ci/deploy_dockerhub.sh "latest"
                 ./bin/ci/deploy_dockerhub.sh "$CIRCLE_SHA1"
```
milescrabill: This is going to give you a newly tagged image on dockerhub for _every commit_ to master. I suspect you do not want this for something like sops.

Dockerfile
```diff
@@ -1,7 +1,10 @@
...
 WORKDIR /go/src/go.mozilla.org/sops
-RUN go get -d -v
-RUN go test ./...
-RUN go install -v ./...
+
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o /bin/sops ./cmd/sops
+RUN apt-get update
+RUN apt-get install -y vim python-pip emacs
+RUN pip install awscli
+ENV EDITOR vim
```
milescrabill: newline at eof

.circleci/config.yml
```diff
@@ -1,25 +1,21 @@
...
-          name: Build containers                                                                                                                                                                                                                               
-          command: |                                                                                                                                                                                                                                           
-            docker build -t mozilla/sops .                                                                                                                                                                                                          
-            docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"                                                                                                                                                                          
+          name: Build containers
+          command: |
+            docker build -t mozilla/sops .
+            docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"
```
milescrabill: See below. Is this what you want?

.circleci/config.yml
autrilla: I think that's fine. We could also tag images with git tags but I don't think pushing on every commit to master would be terrible.

PR #270, opened 2017-10-13
Explain how Shamir's Secret Sharing works

autrilla:


shamir/shamir.go.commented
```diff
@@ -0,0 +1,237 @@
+package shamir
```
jvehent: Why the separate `.commented` file? Can't this be merged directly with shamir.go?

jvehent: @autrilla ^^^

autrilla: Oops, missed this. I'm not sure if we're allowed to modify the shamir file, and I removed some of their comments anyway. We could theoretically merge this, but idk if having an algorithm explanation in the source code alongside implementation details is a lot of noise. Your call.

jvehent: I don't think we should have a separate .commented file. I'd prefer to merge it alongside the source code.

jvehent:
ping @autrilla

autrilla:
done @jvehent

PR #282, opened 2018-01-01
Add parameter to optionally validate all matching keys on decrypt

itwasntandy:
Normally on decryption, SOPS will try each key service in turn, returning as soon
as one has succeeded.
There are times when it may be useful to validate that all keys from a particular
key service are working. For example prior to or following significant infrastructure
changes, or when cycling upstream keys.

You can enable this behavior by passing in `--validate-key <keyservice>`.

For example to ensure all KMS keys are validated you can run:

`sops --decrypt --validate-key kms file.yml`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -590,6 +590,22 @@ in `/tmp/sops.sock` and not the local key service, you can run:
 
 `sops --enable-local-keyservice=false --keyservice unix:///tmp/sops.sock -d file.yaml`
 
+Validate Keys
+~~~~~~~~~~~~~
+
+Normally on decryption, SOPS will try each key service in turn, returning as soon
+as one has succeeded.
+There are times when it may be useful to validate that all keys from a particular
+key service are working. For example prior to or following significant infrastructure
+changes, or when cycling upstream keys.
+
+You can enable this behavior by passing in `--validate-key <keyservice>`.
+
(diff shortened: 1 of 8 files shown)


jvehent:
Thanks for the patch. I think the use case is interesting. I'm not sure about the termonology "validate-key", as this is primarily testing accessibility for decryption. Maybe a better way to approach this would be to implement a `with-key` option that takes a fingerprint or an arn, and performs the action with the given key? It could also support some minimal globbing logic, for example to test all kms keys with `with-key arn:*`, or something along those lines.

itwasntandy:
@jvehent sure, that makes more sense. I'll rework the PR accordingly.

autrilla:
@itwasntandy did you ever have the chance to rework this?

jvehent:
@itwasntandy Ping. If you could rebase this PR, we can try to merge it into the next release. Otherwise we'll have to close it for inactivity.

itwasntandy:
@jvehent hey, sorry, since my son was born at the start of the year, I didn't yet get the chance to follow up on this. 
I should have time in the next week, to rebase it, and rework per the feedback.

Thank you for your patience

jvehent:
this PR has been stale for a year, should we close it?

autrilla:
I'll close since this has gone stale, but feel free to reopen.

PR #284, opened 2018-01-14
fix position of table of contents

0xmohit:


autrilla:
I'm not sure if this was intended. @jvehent ?

jvehent:
Yep, that's on purpose.

PR #296, opened 2018-02-09
Auditing support (cont)

ajvb:
Opening this PR to wrap up @autrilla's auditing support (https://github.com/mozilla/sops/pull/256)

This PR addresses these CR comments:
- adding EncryptEvent support
- adding RotateEvent support
- code review comments on `audit.go`'s `init()` func


audit/audit.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func (p *PostgresAuditor) Handle(event interface{}) {
+	switch event := event.(type) {
+	case DecryptEvent:
+		// Save the event to the database
+		log.WithField("file", event.File).
+			Debug("Saving decrypt event to database")
+		u, err := user.Current()
```
autrilla: Since all current event types need the current user and getting the current user is a cheap operation, I think we should get it regardless of the event type so we don't have the same code on all event types

README.rst
```diff
@@ -602,6 +602,52 @@ in `/tmp/sops.sock` and not the local key service, you can run:
...
+In order to enable auditing, you must first create the database and credentials
+using the schema found in :code:`audit/schema.sql`. This schema defines the
+tables that store the audit events and a role named :code:`sops` that only has
+permission to add entries to the audit event tables. The default password for
+the role :code:`sops` is :code:`sops`. You should change this password.
+
+Once you have created the database, you have to tell SOPS how to connect to it.
+Because we don't want users of SOPS to be able to control auditing, the audit
+configuration file location is not configurable, and must be at
+:code:`/etc/sops/audit.yaml`
```
jvehent: Add: `This file should have strict permissions such that only the root user can modify it.`

audit/audit.go
```diff
@@ -0,0 +1,136 @@
...
+	if flag.Lookup("test.v") != nil {
+		return
+	}
+	for _, pgConf := range conf.Backends.Postgres {
+		auditors = append(auditors, NewPostgresAuditor(pgConf.ConnStr))
+	}
+}
+
+// TODO: Make platform agnostic
+const configFile = "/etc/sops/audit.yaml"
```
jvehent: that's actually pretty easy to do: just move this line into a file called `auditenv_posix.go`, and create one called `auditenv_windows.go` that sets the location to a windows-friendly path. Not sure if it's required for this iteration of the feature though.

audit/schema.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE decrypt_event (
+  id SERIAL PRIMARY KEY,
+  timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  username TEXT,
+  file TEXT
+);
+
+CREATE RULE decrypt_event_delete_protection AS ON DELETE TO decrypt_event DO INSTEAD NOTHING;
```
jvehent: Do we actually want this? The GRANT rules already prevent the sops user from doing anything else but inserting, and I can think of a legitimate reason to purge the DB from old records.

audit/schema.sql
```diff
@@ -0,0 +1,35 @@
...
+);
+
+CREATE RULE encrypt_event_delete_protection AS ON DELETE TO encrypt_event DO INSTEAD NOTHING;
+
+CREATE TABLE rotate_event (
+  id SERIAL PRIMARY KEY,
+  timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  username TEXT,
+  file TEXT
+);
```
jvehent: All three tables have the same columns, so maybe they should just be one table with an `action` column added that reflects the decrypt, encrypt or rotate event?

cmd/sops/edit.go
```diff
@@ -87,17 +89,23 @@ func editExample(opts editExampleOpts) ([]byte, error) {
...
-		ShamirThreshold:   opts.GroupThreshold,
+	path, err := filepath.Abs(opts.InputPath)
+	if err != nil {
+		return nil, err
+	}
+	tree := sops.Tree{
+		Branch: branch,
+		Metadata: sops.Metadata{
+			KeyGroups:         opts.KeyGroups,
+			UnencryptedSuffix: opts.UnencryptedSuffix,
```
jvehent: nit: gofmt

cmd/sops/encrypt.go
```diff
@@ -28,17 +29,23 @@ func encrypt(opts encryptOpts) (encryptedFile []byte, err error) {
...
-		ShamirThreshold:   opts.GroupThreshold,
+	path, err := filepath.Abs(opts.InputPath)
+	if err != nil {
+		return nil, err
+	}
+	tree := sops.Tree{
+		Branch: branch,
+		Metadata: sops.Metadata{
+			KeyGroups:         opts.KeyGroups,
+			UnencryptedSuffix: opts.UnencryptedSuffix,
```
jvehent: nit: gofmt

cmd/sops/rotate.go
```diff
@@ -69,5 +70,8 @@ func rotate(opts rotateOpts) ([]byte, error) {
 	if err != nil {
 		return nil, common.NewExitError(fmt.Sprintf("Could not marshal tree: %s", err), codes.ErrorDumpingTree)
 	}
+	audit.SubmitEvent(audit.RotateEvent{
+		File: tree.FilePath,
+	})
```
jvehent: recording an audit event should happen before the operation is performed, not after. Otherwise a malicious user can simply block the DB connection and still use sops successfully.

sops.go
```diff
@@ -312,6 +315,9 @@ func (tree Tree) Encrypt(key []byte, cipher Cipher) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("Error walking tree: %s", err)
 	}
+	audit.SubmitEvent(audit.EncryptEvent{
+		File: tree.FilePath,
+	})
```
jvehent: move to top of function

sops.go
```diff
@@ -363,6 +369,9 @@ func (tree Tree) Decrypt(key []byte, cipher Cipher) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("Error walking tree: %s", err)
 	}
+	audit.SubmitEvent(audit.DecryptEvent{
+		File: tree.FilePath,
+	})
```
jvehent: move to top of function

audit/schema.sql
autrilla: Doing that could cripple us when we want to add new event types (I expressed my opinion on this here: https://github.com/mozilla/sops/pull/256#discussion_r142147597)

jvehent: One could argue that changing the DB schema to add new type is more cumbersome that putting everything into a single table. I agree with the argument that we may want extra data, and maybe adding a `jsonb` column with `details` would help?

Ultimately, this should just be a blockchain! :trollface:

cmd/sops/edit.go
ajvb: @jvehent This seems to be formatted already. What's making you think otherwise? Am I missing something?

audit/schema.sql
ajvb: Are these the only kind of audit events that will exist within sops? Events where it's "`action` on `file` by `user`?

ajvb: @jvehent @autrilla I'm impartial to this, but I would say that unless we expect the DB to have high throughput of writes migrating for either-or shouldn't be too difficult.

autrilla: I agree that migrating from either if it turns out the other way is better won't be difficult. @jvehent is right that changing the DB schema every time a new event type is added (hopefully this won't happen too often) is cumbersome, at least how we currently manage schema changes.

ajvb:
@autrilla Is this good to merge?

autrilla:
@ajvb yes, lgtm

PR #298, opened 2018-02-20
Rewrite set functionality to make it work with nested structures

### LINKED ISSUE #297: Sops --set destroys yaml structure (3.0.2)
I started with:
```
db:
    passwords:
        root_user: ENC...
...
```
After running `sops --set '["db"]["passwords"]["user1"] "user1password"' secrets.yaml` I ended up with:
```
root_user: ENC...
user1: ENC...
```

```
$ sops --version
sops 3.0.2 (latest)
```

autrilla:
The previous approach of truncating the tree was erroneous and it isn't possible to make it work by truncating the tree beforehand anyway because we treat the tree as an immutable data structure, so the whole tree has to be copied and then the appropriate values have to be modified.

Fixes #297

### DIFF
--- a/cmd/sops/set.go
+++ b/cmd/sops/set.go
@@ -40,14 +40,7 @@ func set(opts setOpts) ([]byte, error) {
 	}
 
 	// Set the value
-	key := opts.TreePath[len(opts.TreePath)-1]
-	path := opts.TreePath[:len(opts.TreePath)-1]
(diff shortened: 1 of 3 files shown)


jvehent:
Redirecting to @ajvb because he's just basking in free time :p

autrilla:
@ajvb all paths are tested now

PR #302, opened 2018-02-28
CircleCI config for building/pushing up container

ajvb:
Needs CircleCI project with Docker Hub env vars (`DOCKER_EMAIL`, `DOCKER_USER`, `DOCKER_PASS`).

May be worth adding flow for pushing up new releases as well?


circle.yml
```diff
@@ -0,0 +1,27 @@
...
+        "$CIRCLE_BUILD_URL"
+        > version.json
+
+    - docker build -t sops:build .
+
+deployment:
+  master:
+    branch: "master"
+    commands:
+      - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
```
autrilla: Could this leak the credentials in the CircleCI console? cc @g-k

circle.yml
```diff
@@ -0,0 +1,27 @@
...
+      - "docker tag sops:build docker.io/mozilla/sops:latest"
```
autrilla: I think we should also tag and push with the git sha1 or the tag, because if we use latest rolling back deployments will be complicated.

circle.yml
```diff
@@ -0,0 +1,27 @@
+machine:
```
autrilla: This looks like CircleCI 1.0, right? CircleCI 1.0 is being sunset in August this year. I think starting with CircleCI 2.0 might be worth it considering 1.0 will stop working relatively soon.

circle.yml
g-k: Yeah, when login fails it can print the user and password in the error message.

The shell script and docker v2 config in https://github.com/mozilla-services/tecken shouldn't have that problem.

ajvb:
I was clearly not paying attention. The current implementation in `master` is just fine.

PR #308, opened 2018-03-27
Set default logging level to Warn

deed02392:


autrilla:
Seems harmless enough, can't wait until someone asks for a way to enable logging for the decrypt package :p

deed02392:
@autrilla Hahaha, please do ping me when that happens!

PR #309, opened 2018-03-27
publish statically-linked binaries

willejs:
I am using helm-secrets in an alpine container, which calls sops under the hood. Obviously Alpine has uclibc installed, so the dynamically linked binaries against glibc that are published will not work.

Instead I am proposing we statically link the binaries across all platforms for portability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ before_deploy:
   - mkdir dist
   - make deb-pkg rpm-pkg
   - mv *.deb *.rpm dist/
-  - GOOS=darwin go build -o dist/sops-${TRAVIS_TAG}.darwin go.mozilla.org/sops/cmd/sops
-  - GOOS=windows go build -o dist/sops-${TRAVIS_TAG}.exe go.mozilla.org/sops/cmd/sops
-  - GOOS=linux go build -o dist/sops-${TRAVIS_TAG}.linux go.mozilla.org/sops/cmd/sops
+  - GOOS=darwin CGO_ENABLED=0 go build -o dist/sops-${TRAVIS_TAG}.darwin go.mozilla.org/sops/cmd/sops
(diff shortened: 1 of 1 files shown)


autrilla:
Why not publish both? If we only published statically linked binaries, what's the size increase by linking statically?

.travis.yml
```diff
@@ -22,9 +22,9 @@ before_deploy:
   - mkdir dist
   - make deb-pkg rpm-pkg
   - mv *.deb *.rpm dist/
-  - GOOS=darwin go build -o dist/sops-${TRAVIS_TAG}.darwin go.mozilla.org/sops/cmd/sops
-  - GOOS=windows go build -o dist/sops-${TRAVIS_TAG}.exe go.mozilla.org/sops/cmd/sops
-  - GOOS=linux go build -o dist/sops-${TRAVIS_TAG}.linux go.mozilla.org/sops/cmd/sops
+  - GOOS=darwin CGO_ENABLED=0 go build -o dist/sops-${TRAVIS_TAG}.darwin go.mozilla.org/sops/cmd/sops
+  - GOOS=windows CGO_ENABLED=0 go build -o dist/sops-${TRAVIS_TAG}.exe go.mozilla.org/sops/cmd/sops
```
jvehent: FYI
> The cgo tool is enabled by default for native builds on systems where it is expected to work. It is disabled by default when cross-compiling. You can control this by setting the CGO_ENABLED environment variable when running the go tool: set it to 1 to enable the use of cgo, and to 0 to disable it. The go tool will set the build constraint "cgo" if cgo is enabled. 
so this already shouldn't a problem on darwin and windows since we cross-compile from linux. That said, I'm all for being explicit, so +1.

ajvb:
:+1:

willejs:
Good question @autrilla! I should have said!
Dynamic 16.8MB 
Static 17.6MB
@jvehent yep, you can't dynamically link to libs you haven't got! Not a lot of people realise this behaviour though, so I thought i'd be explicit.

willejs:
Cheers @jvehent !!!

PR #312, opened 2018-03-28
Prevent files from being encrypted twice

autrilla:
Error message rendered:

```
❯ sops -e foo.enc.yaml
The file you have provided contains a top-level entry called 'sops'. This
is generally due to the file already being encrypted. SOPS uses a top-level
entry called 'sops' to store the metadata required to decrypt the file. For
this reason, SOPS can not encrypt files that already contain such an entry.

If this is an unencrypted file, rename the 'sops' entry.

If this is an encrypted file and you want to edit it, use the editor mode,
for example: `sops my_file.yaml`
```


jvehent:
r+ but maybe we need a flag to force reencryption, as odd as it may be?

autrilla:
I'd say if people do ask for it and have a legitimate use case, then we can add the flag, but as of now, I can't think of any reason for which you'd want to do this, so implementing the flag isn't worth the effort in my opinion

PR #313, opened 2018-03-28
dont create the file if its empty

devinburnette:
Current experience when running `sops file.yaml` and exiting without saving anything creates an empty file that throws a `sops metadata not found` error if re-opened with sops.

As a user, I would expect if I didn't save a new file in my editor, then there shouldn't be a file created.

This PR attempts to correct that by just not creating the file if the resulting output from an edit session is empty.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -486,7 +486,7 @@ func main() {
 			} else {
 				// File doesn't exist, edit the example file instead
 				var groups []sops.KeyGroup
-				groups, err := keyGroups(c, fileName)
+				groups, err = keyGroups(c, fileName)
 				if err != nil {
 					return toExitError(err)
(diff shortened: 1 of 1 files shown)


autrilla:
Thanks for the patch! While this fixes the issue, it's doing it the wrong way in my opinion. SOPS already has checks to prevent modifying or creating files if the file isn't modified, but that check doesn't seem to work for new files. This is due to the `err` name being rescoped on this line: https://github.com/devinburnette/sops/blob/d413945dcf4299cf222fc35221009f77fe167320/cmd/sops/main.go#L489

Changing it to `groups, err = keyGroups(c, fileName)` should fix it.

devinburnette:
Ahh.. Good catch, I've made that change.

Issue #77, opened 2016-08-03
`directory_regex` support (or something similar)?


chroto:
I have users of our SOPS implementation that would prefer to use subdirectories to determine what encryption keys to use.  Our project structure is as follows:

```
├── engineering
├── ops
├── resources
└── ssl
```

Right now, we have a .sops.yaml file in each directory and require users to cd to the directory before creating a new file.  I get a lot of forgetful users who will create secrets from the top level.

So instead of

```
cd engineering
sops foo
```

we get

```
sops engineering/foo
```

If I could extend the configuration featureset to include a `directory_regex` I can run a regex on the first filename that is checked and see if it matches.  If no match, pass through to the `filename_regex` as it works today.

```
creation_rules:
  - directory_regex: ^engineering/
    kms: '...'
    pgp: '...'
  - directory_regex: ^ops/
    kms: '...'
    pgp: '...'
```

happy to submit a patch if you are interested in accepting

jvehent:
I think it might be a useful feature. We may want to keep the existing `filename_regex` for backward compatibility, and have `directory_regex` only match on the dirname of a path. Another option would be to introduce `path_regex` that matches on both the dirname and basename of a file.

Happy to take a patch!

mumoshu:
@chroto @jvehent It seems like `file_regex` matches against the path rather than the basename as of today!

Reproduction step:

```yaml
creation_rules:
- filename_regex: foo/bar/baz/*
  kms: "arn:aws:kms:myregion:myawsaccount:key/mykeyid"
```

```
$ sops -e foo/bar/baz/test.yaml
```

So, can we close this issue as resolved now?

PR #314, opened 2018-03-28
deprecate filename_regex in favor of path_regex

devinburnette:
Firstly, I wanted to express how grateful we are for this tool. It's going to drastically change the way we do secrets management for the better across our engineering org. I was wondering if you think the following change is acceptable?

Currently, the `.sops.yaml` file takes an attribute called `filename_regex`. We've found that title to be a bit misleading as it also supports full directory paths as written.

This PR started out as simply an effort to document how we're currently using it with a test to make sure directory path support isn't later removed. Eventually, I ended up implementing [this comment](https://github.com/mozilla/sops/issues/77#issuecomment-237364812).

So what you see here is adding support for an attribute called `path_regex` that has the same behavior as `filename_regex` does today. `filename_regex` still works for backwards compatibility, but gets a message that it may later be removed when used.

I thought about using pointers instead of strings for the type assignment of these values to make the logic a bit easier to grok, but wanted to keep support for the case where the regex attribute isn't supplied at all in case anyone currently uses it that way.

Closes https://github.com/mozilla/sops/issues/77

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,11 +11,19 @@ import (
 	"regexp"
 
 	"github.com/mozilla-services/yaml"
+	"github.com/sirupsen/logrus"
 	"go.mozilla.org/sops"
 	"go.mozilla.org/sops/gcpkms"
 	"go.mozilla.org/sops/kms"
+	"go.mozilla.org/sops/logging"
 	"go.mozilla.org/sops/pgp"
 )
+
+var log *logrus.Logger
+
+func init() {
+	log = logging.NewLogger("CONFIG")
+}
 
 type fileSystem interface {
 	Stat(name string) (os.FileInfo, error)
@@ -72,6 +80,7 @@ type kmsKey struct {
 
 type creationRule struct {
 	FilenameRegex   string `yaml:"filename_regex"`
+	PathRegex       string `yaml:"path_regex"`
 	KMS             string
 	PGP             string
 	GCPKMS          string     `yaml:"gcp_kms"`
@@ -103,9 +112,25 @@ func loadForFileFromBytes(confBytes []byte, filePath string, kmsEncryptionContex
 	var rule *creationRule
 
 	for _, r := range conf.CreationRules {
-		if match, _ := regexp.MatchString(r.FilenameRegex, filePath); match {
+		if r.PathRegex == "" && r.FilenameRegex == "" {
 			rule = &r
 			break
+		}
+		if r.PathRegex != "" && r.FilenameRegex != "" {
+			return nil, fmt.Errorf("error loading config: both filename_regex and path_regex were found, use only path_regex")
+		}
+		if r.FilenameRegex != "" {
+			if match, _ := regexp.MatchString(r.FilenameRegex, filePath); match {
+				log.Warn("The key: filename_regex will be removed in a future release. Instead use key: path_regex in your .sops.yaml file")
+				rule = &r
+				break
+			}
+		}
+		if r.PathRegex != "" {
+			if match, _ := regexp.MatchString(r.PathRegex, filePath); match {
+				rule = &r
+				break
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


config/config_test.go
```diff
@@ -44,22 +44,38 @@ func TestFindConfigFileCurrentDir(t *testing.T) {
...
+    gcp_kms: baz
+`)
+
+var sampleConfigWithPath = []byte(`
+creation_rules:
+  - path_regex: foo/bar*
+    kms: "1"
+    pgp: "2"
+    gcp_kms: "3"
+  - filename_regex: "somefilename.yml"
```
devinburnette: left for testing backward compatibility

config/config.go
```diff
@@ -103,10 +104,26 @@ func loadForFileFromBytes(confBytes []byte, filePath string, kmsEncryptionContex
...
+		if r.PathRegex == "" && r.FilenameRegex == "" {
 			rule = &r
 			break
 		}
+		if r.PathRegex != "" && r.FilenameRegex != "" {
+			return nil, fmt.Errorf("error loading config: both filename_regex and path_regex were found, use only path_regex")
+		}
+		if r.FilenameRegex != "" {
+			if match, _ := regexp.MatchString(r.FilenameRegex, filePath); match {
+				fmt.Println("WARNING: filename_regex will be removed in a future release, use path_regex instead")
```
ajvb: May want to use `logrus` instead of `fmt`?

autrilla: as @ajvb said, prefer using `logrus` here for logging

config/config.go
```diff
@@ -103,10 +104,26 @@ func loadForFileFromBytes(confBytes []byte, filePath string, kmsEncryptionContex
...
-		if match, _ := regexp.MatchString(r.FilenameRegex, filePath); match {
+		if r.PathRegex == "" && r.FilenameRegex == "" {
 			rule = &r
 			break
 		}
+		if r.PathRegex != "" && r.FilenameRegex != "" {
+			return nil, fmt.Errorf("error loading config: both filename_regex and path_regex were found, use only path_regex")
+		}
+		if r.FilenameRegex != "" {
+			if match, _ := regexp.MatchString(r.FilenameRegex, filePath); match {
```
autrilla: I'm on the fence as to whether we should fix `filename_regex` and make it match on just the file name instead of the compete path or leaving it as it is. In the Python version `filename_regex` matched on just the file name, so technically this is a bug introduced in the Go version, which matches on the whole path. Thoughts?

cc @jvehent @ajvb

config/config.go
autrilla: I think this error message could be improved, it should tell users what they should do. If I put myself in the shoes of a user that hasn't touched their `.sops.yaml` file in months, I feel very uncertain about what action I need to take and why.

config/config.go
devinburnette: We're personally using it to match on the full path and it's amazing! That would make us sad if that went away.

ajvb: I'm obviously quite new to sops, so interested to hear what @jvehent thinks, but the fact that both @devinburnette and those in the discussion in #77 are taking advantage of the bug pushes me towards voting for this PR. Keep it how it is as it is a feature users want, but move towards a more accurate naming.

jvehent: mandatory xkcd

jvehent: With that said, I don't think we should change the current behavior of `filename_regex`. Matching on the full path is a useful feature.

autrilla:
Can you set up the logger like it's done here https://github.com/mozilla/sops/blob/master/pgp/keysource.go#L29-L33 so that the default log level can then be overridden through command line  flags?

config/config.go
```diff
@@ -11,12 +11,20 @@ import (
...
+	"go.mozilla.org/sops/logging"
 	"go.mozilla.org/sops/pgp"
 )
 
+var log *logrus.Logger
+
+func init() {
+	log = logging.NewLogger("CONFIG")
+}
+
```
devinburnette: @autrilla something like this?

autrilla: Yes, that's exactly what I was looking for

PR #317, opened 2018-03-30
Don't yaml.MarshalValue strings on extract

autrilla:
When using --extract with multi-line strings, the behavior doesn't match the docs:

```
~/Projects/go/src/go.mozilla.org/sops extract-string*
❯ sops -d foo.yaml 
hello: |
    multi-line
    string

~/Projects/go/src/go.mozilla.org/sops extract-string*
❯ sops -d --extract '["hello"]' foo.yaml                    
|
    multi-line
    string

```

When the output value is already a string, there is no need to pass it through `yaml.MarshalValue`, and doing so will change the result if for example the string contains new lines. This makes it annoying to build shell pipelines using --extract, as you have to pipe it through something that removes yaml formatting.

With this patch:

```
~/Projects/go/src/go.mozilla.org/sops extract-string*
❯ sops -d --extract '["hello"]' foo.yaml
multi-line
string

```


autrilla:
I've added a simple functional test for this

Issue #324, opened 2018-04-08
Add the ability to specify  --unencrypted-suffix in the .sops.yaml file


dovreshef:
Hi

It would help my use cases to specify the  --unencrypted-suffix in the creation_rules section in the .sops.yaml.

Is that something that you'd consider? I'm willing to implement it.

autrilla:
Yes, this is a feature we would merge. I assume you'd also want support for --encrypted-suffix too?

dovreshef:
@autrilla Yes :). 
Will finish working and --encrypted-suffix and start on this.

PR #325, opened 2018-04-08
Adds `unencrypted_suffix` and `encrypted_suffix` to `.sops.yaml`

dovreshef:
Fixes issue #324

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -704,8 +704,8 @@ func loadConfig(c *cli.Context, file string, kmsEncryptionContext map[string]*st
 		configPath = c.String("config")
 	} else {
 		// Ignore config not found errors returned from FindConfigFile since the config file is not mandatory
-		configPath, _ = config.FindConfigFile(".")
-		if configPath == "" {
+		configPath, err = config.FindConfigFile(".")
+		if err != nil {
 			// If we can't find a config file, but we were not explicitly requested to, assume it does not exist
 			return nil, nil
 		}
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -696,7 +696,7 @@ func keyGroups(c *cli.Context, file string) ([]sops.KeyGroup, error) {
 }
 
 // loadConfig will look for an existing config file, either provided through the command line, or using config.FindConfigFile.
-// Since a config file is not required, a return value of nil, nil is possible
+// Since a config file is not required, this function does not error when one is not found, and instead returns a nil config pointer
 func loadConfig(c *cli.Context, file string, kmsEncryptionContext map[string]*string) (*config.Config, error) {
 	var err error
 	var configPath string
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -374,6 +374,19 @@ func main() {
 
 		unencryptedSuffix := c.String("unencrypted-suffix")
 		encryptedSuffix := c.String("encrypted-suffix")
+		conf, err := loadConfig(c, fileName, nil)
+		if err != nil {
+			return toExitError(err)
+		}
+		if conf != nil {
+			// command line options have precedence
+			if unencryptedSuffix == "" {
+				unencryptedSuffix = conf.UnencryptedSuffix
+			}
+			if encryptedSuffix == "" {
+				encryptedSuffix = conf.UnencryptedSuffix
+			}
+		}
 		if unencryptedSuffix != "" && encryptedSuffix != "" {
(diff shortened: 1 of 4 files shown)


config/config_test.go
```diff
@@ -194,3 +221,20 @@ func TestKeyGroupsForFileWithGroups(t *testing.T) {
...
+
+func TestLoadConfigFileWithEncryptedSuffix(t *testing.T) {
+	conf, err := loadForFileFromBytes(sampleConfigWithUnencryptedSuffix, "barfoo", nil)
+	assert.Equal(t, nil, err)
+	assert.Equal(t, "_enc", conf.EncryptedSuffix)
+}
+
+func TestLoadConfigFileWithInvalidParameters(t *testing.T) {
+	_, err := loadForFileFromBytes(sampleConfigWithInvalidParameters, "foobar", nil)
+	assert.EqualError(t, err, "error loading config: cannot use both encrypted_suffix and unencrypted_suffix for the same path_regex")
```
autrilla: Don't check the error message, just checking that an error happened is enough.

config/config_test.go
```diff
@@ -92,6 +92,33 @@ creation_rules:
       - resource_id: baz
 `)
 
+var sampleConfigWithUnencryptedSuffix = []byte(`
```
autrilla: This variable name is confusing, as there's also an encrypted_suffix

config/config.go
```diff
@@ -138,6 +142,10 @@ func loadForFileFromBytes(confBytes []byte, filePath string, kmsEncryptionContex
 		return nil, fmt.Errorf("error loading config: no matching creation rules found")
 	}
 
+	if rule.UnencryptedSuffix != "" && rule.EncryptedSuffix != "" {
+		return nil, fmt.Errorf("error loading config: cannot use both encrypted_suffix and unencrypted_suffix for the same path_regex")
```
autrilla: Replace `path_regex` with `rule`

cmd/sops/main.go
```diff
@@ -687,27 +695,36 @@ func keyGroups(c *cli.Context, file string) ([]sops.KeyGroup, error) {
...
+
+func shamirThreshold(c *cli.Context, file string) (int, error) {
+	if c.Int("shamir-secret-sharing-threshold") != 0 {
+		return c.Int("shamir-secret-sharing-threshold"), nil
+	}
+	conf, err := loadConfig(c, file, nil)
+	if conf == nil {
+		// If shamir flag isn't set and we can't find a config file,
+		// assume we don't want Shamir
 		return 0, err
```
autrilla: Why does this return an error?

dovreshef: This encompass two cases which were separated previously.
1. No config was provided. Err will be nil and ShamirThreshold will be the default value of 0.
2. We did find a config file, but failed to load it. In that case the calling function will print the error and exit.

autrilla: OK, that makes sense. Could you document `loadConfig` so that this behavior is clear to other people?

dovreshef: Done. Also found bug :(.

cmd/sops/main.go
```diff
@@ -695,15 +695,18 @@ func keyGroups(c *cli.Context, file string) ([]sops.KeyGroup, error) {
 	return []sops.KeyGroup{group}, nil
 }
 
+// loadConfig will look for an existing config file, either provided through the command line, or using config.FindConfigFile.
+// Since a config file is not required, a return value of nil, nil is possible
```
autrilla: Since a config file is not required, this function does not error when one is not found, and instead returns a nil config pointer

dovreshef: Ok. Changing.

cmd/sops/main.go
```diff
@@ -695,15 +695,18 @@ func keyGroups(c *cli.Context, file string) ([]sops.KeyGroup, error) {
...
 	var configPath string
 	if c.String("config") != "" {
 		configPath = c.String("config")
 	} else {
-		configPath, err = config.FindConfigFile(".")
-		if err != nil {
-			// If we can't find a config file, but we were not request it, assume it does not exist
+		// Ignore config not found errors returned from FindConfigFile since the config file is not mandatory
+		configPath, _ = config.FindConfigFile(".")
+		if configPath == "" {
```
autrilla: Why change from checking the error to checking if configPath is empty? Afaict FindConfigFile only returns an error if no config file is found

dovreshef: You're correct. Fixed.
Can you rebase the merge into master? If you don't want the extra commits..

autrilla: Yeah, I can just squash on merge

autrilla:
Thanks for the patch!

PR #326, opened 2018-04-10
Create version.json for moz Dockerflow

ajvb:
Create a `version.json` file inline with https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,18 @@ jobs:
       - image: circleci/golang:1.8
     steps:
       - checkout
+      - run:
+          name: Create version.json
+          command: |
+            # create a version.json per https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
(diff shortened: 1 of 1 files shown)


ajvb:
@autrilla Ha yeah... totally skipped over that part. Let me check in with ops.

milescrabill:
This is something pretty specific to the cloudops deployment model (dockerflow et al) at mozilla. It's up to sops maintainers (and there's some overlap there to be sure) to decide whether this makes sense to integrate into packaged sops docker containers / the sops ci process.

As far as the cloudops specific mirroring of a sops docker container, you don't _need_ the `version.json` file in your container to mirror your docker image - the `pull-verify-push` function requires that, but we do other things in other projects where docker images are created either outside of ci or in a manner that doesn't fit - for example, quite a few projects publish to an internal-only ECR docker registry.

ajvb:
@milescrabill Thank you for the detailed write up. I'm inclined to go with the option you laid out of not including this here and handling the mirroring internally using the ECR registry.

PR #333, opened 2018-04-17
[PYTHON] Un-pin Python dependencies

rouge8:
Now that the Python version isn't being actively developed, it'll be nice for consumers of the library to not have conflicts with the pinned sops dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,9 +20,9 @@ setup(
     zip_safe=True,
     long_description=README,
     install_requires=[
-        'ruamel.yaml==0.11.7',
+        'ruamel.yaml>=0.11.7',
         'boto3>=1.1.3',
-        'cryptography==1.4',
+        'cryptography>=1.4',
         'setuptools>=18.8.1',
         'ordereddict>=1.1',
         'simplejson>=3.8'],
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,8 @@ setup(
     zip_safe=True,
     long_description=README,
     install_requires=[
-        'ruamel.yaml>=0.11.7',
+        'ruamel.yaml>=0.11.7,<0.12.0 ; python_version=="2.6"',
+        'ruamel.yaml>=0.11.7 ; python_version>"2.6"',
         'boto3>=1.1.3',
         'cryptography>=1.4',
         'setuptools>=18.8.1',


rouge8:
Hmmm, Python 2.6 fails with the latest ruamel.yaml. Thoughts on dropping Python 2.6 support vs. conditional ruamel.yaml dependencies vs. setting an upper bound on ruamel.yaml?

autrilla:
The main reason I don't want to unpin dependency versions is that it causes breakage, as I mentioned in #217. This is a great example, with sops breaking on new installs at some point without us changing anything, when a new ruamel.yaml version was released.

rouge8:
That makes sense when there’s regular releases, but it’s a lot easier for users to pin when installing for themselves than to unpin.

rouge8:
> Either we break support for old systems, or we maintain insecure/old dependencies for everyone.

One option to keep Py26 support would be environment markers on the ruamel.yaml dependency. I can add that if you’re interested. 

> How/Why are you still relying on python-sops?

We have a few projects that used it as a Python library and have run into compatibility issues a few times so it’s been easier to just add the python version to e.g. requirements.txt

jvehent:
> One option to keep Py26 support would be environment markers on the ruamel.yaml dependency. I can add that if you’re interested.

I think that would be good. I'm sure someone somewhere is running python sops on rhel 5...

> We have a few projects that used it as a Python library and have run into compatibility issues a few times so it’s been easier to just add the python version to e.g. requirements.txt

As much as I dislike making this advice, you may be better off shelling out to go sops. There are tons of improvements and bugfixes in sops 2.0+ that will break the sops 1.0 python parser.

rouge8:
> As much as I dislike making this advice, you may be better off shelling out to go sops. There are tons of improvements and bugfixes in sops 2.0+ that will break the sops 1.0 python parser.

Yeah that's definitely on the roadmap.

Pushed the Python 2.6 fix.

jvehent:
Looks good, and the tests now pass. Merging.

By the way, do you rely on the pip package? I have stopped uploading new version there, but it could be done.

jvehent:
I just pushed 1.17 up to pypi. Hopefully that didn't break anything, it's been a while since I pushed a release there...

PR #335, opened 2018-04-19
Refactor Store interface to reflect operations SOPS performs

### LINKED ISSUE #334: Binary input starting with a double quote not handled properly
I am trying to encrypt a plain text file but get `Error dumping file: No binary data found in tree` when trying to decrypt the file. The file I am trying to encrypt looks like the following:

`"text1" "text2"`

I do force `--input-type binary` and `--output-type json`. I have tried a few variations and most work for example, the following file inputs work: 
`"text1`
`"text1"`
`""`
`"" "`

But the following combinations result in the error because the resulting json does not have a `data` key:
`"" ""`
```
""
""
```
`"text1" "text2"`

This is how I am testing:
`$ sops -e --input-type binary --output-type json --gcp-kms <kms_location> --verbose -i testfile`

Thanks

autrilla:
Previously the Store interface tried to mimic the Marshaler and Unmarshaler
interfaces. This was a mistake, as it meant Stores had no idea whether the files they
were loading were encrypted or not.

Sort-of-fixes #334, but another bugfix is required to solve that issue "the right way".

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -19,9 +19,25 @@ type BinaryStore struct {
 	store Store
 }
 
-// Marshal takes a sops tree branch and returns a json formatted string
-func (store BinaryStore) Marshal(tree sops.TreeBranch) ([]byte, error) {
-	for _, item := range tree {
+func (store BinaryStore) LoadEncryptedFile(in []byte) (sops.Tree, error) {
+	return store.store.LoadEncryptedFile(in)
+}
+
+func (store BinaryStore) LoadPlainFile(in []byte) (sops.TreeBranch, error) {
+	return sops.TreeBranch{
+		sops.TreeItem{
+			Key:   "data",
+			Value: string(in),
+		},
+	}, nil
+}
+
+func (store BinaryStore) EmitEncryptedFile(in sops.Tree) ([]byte, error) {
+	return store.store.EmitEncryptedFile(in)
+}
+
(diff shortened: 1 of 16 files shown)


stores/json/store.go
```diff
@@ -19,43 +19,34 @@ type BinaryStore struct {
...
-	}
-	return nil, fmt.Errorf("No binary data found in tree")
+func (store BinaryStore) LoadEncryptedFile(in []byte) (sops.Tree, error) {
+	return store.store.LoadEncryptedFile(in)
 }
 
-// MarshalWithMetadata takes a sops tree branch and sops metadata and marshals them to json.
-func (store BinaryStore) MarshalWithMetadata(tree sops.TreeBranch, metadata sops.Metadata) ([]byte, error) {
-	return store.store.MarshalWithMetadata(tree, metadata)
+func (store BinaryStore) LoadPlainFile(in []byte) (sops.TreeBranch, error) {
```
autrilla: Basically the whole PR is name changes except this. In `master`, this would be done through `Unmarshal`, which would first try to load the file through the JSON store to check if it's encrypted (encrypted binary files are JSON). That's completely broken, because anything that seems like valid JSON will be loaded through the JSON store even though the binary store was desired.

With this PR, because the store knows whether the file being loaded is encrypted or not, there's no need to infer it through a hack like in `master`.

PR #357, opened 2018-06-01
Handle escaping in original json

### LINKED ISSUE #356: Using \\ in json keys is not handled
### Sops version
sops 3.0.5 (latest)

### Problem
Using escapes in json keys are not handled and produce unexpected behaviour.  
Trying to save a file with following content causes `Could not marshal tree: Error marshaling to json: invalid character 'd' in string escape code`
```json
{
	"key\\dummy": "value",
}
```
---
Additionally, using a file like this:
```json
{
	"key\\\\dummy": "value",
}
```
produces any encrypted file like:
```
{
	"test\\dummy": "...",
	"sops": {...}
}
```
Trying to open this file then causes `Error decrypting tree: Error walking tree: Could not decrypt value: Could not decrypt with AES_GCM: cipher: message authentication failed`

ajvb:
Fixes #356

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -170,7 +170,7 @@ func (store Store) encodeTree(tree sops.TreeBranch) ([]byte, error) {
 		}
 		k, err := json.Marshal(item.Key.(string))
 		if err != nil {
-			return nil, fmt.Errorf("Error encoding key %s: %s", v, err)
+			return nil, fmt.Errorf("Error encoding key %s: %s", k, err)
 		}
 		out += string(k) + `: ` + string(v)
 		if i != len(tree)-1 {
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"strings"
 
 	"go.mozilla.org/sops"
 	"go.mozilla.org/sops/stores"
@@ -73,14 +72,6 @@ func (store Store) sliceFromJSONDecoder(dec *json.Decoder) ([]interface{}, error
 
 var errEndOfObject = fmt.Errorf("End of object")
 
-func correctKeyForEncoding(key string) string {
-	// handles escaping in the original json
-	if strings.Contains(key, "\\") {
-		key = strings.Replace(key, "\\", "\\\\", -1)
-	}
-	return key
-}
-
 func (store Store) treeItemFromJSONDecoder(dec *json.Decoder) (sops.TreeItem, error) {
 	var item sops.TreeItem
 	key, err := dec.Token()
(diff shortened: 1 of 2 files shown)


stores/json/store.go
```diff
@@ -72,14 +73,22 @@ func (store Store) sliceFromJSONDecoder(dec *json.Decoder) ([]interface{}, error
 
 var errEndOfObject = fmt.Errorf("End of object")
 
+func correctKeyForEncoding(key string) string {
+	// handles escaping in the original json
+	if strings.Contains(key, "\\") {
+		key = strings.Replace(key, "\\", "\\\\", -1)
```
jvehent: two questions:
1. what does the json spec say about backslash encoding?
2. shouldn't the end-users take care of this before calling sops instead of doing magic here?

autrilla: There are plenty of other cases other than "\\" that need escaping as per [RFC 7159](https://tools.ietf.org/html/rfc7159#section-7).

I think the correct way to solve this would be to fix this https://github.com/ajvb/sops/blob/58524f7d7990d7cc8c2509d79446d912d5e590cf/stores/json/store.go#L180 so that it JSON-encodes strings before outputting.

ajvb: Thanks for asking these questions. In reading the spec, I tested out how `sops -e` functions with other escaping scenarios (like `"a_key_with\"quotes\"": "value"`) and it broke in the same way. I think what I just pushed is closer to what we want.

1. In looking at [RFC 7159#Strings](https://tools.ietf.org/html/rfc7159#section-7) and [ECMA-404](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf), they just call out that there are a few characters (like `"` and `\`) that need to be escaped
2. I don't think so, imo we should decode and encode compliant json. The examples in #356 are strange to me, but they are proper json keys.

stores/json/store_test.go
```diff
@@ -198,6 +198,27 @@ func TestDecodeJSONArrayOfObjects(t *testing.T) {
 	assert.Equal(t, expected, branch)
 }
 
+func TestDecodeJSONWithEscaping(t *testing.T) {
```
ajvb: this test no longer tests the correct error case, fixing.

stores/json/store.go
autrilla: I wonder if we should encode values like this too. I think we are emitting invalid JSON when we output decrypted JSON files if we do it the same way we did for the keys, as the encrypted values could contain characters that need to be escaped too, right?

ajvb: @autrilla ah, missed your first comment. Yeah totally agree.

In regards to values, we are already encoding using `json.Marshal` - https://github.com/mozilla/sops/blob/master/stores/json/store.go#L138 - that's actually where I realized that needed to be applied to keys as well.

autrilla: Oh, hah, okay, we already do the right thing for values!

PR #358, opened 2018-06-17
Add support for Azure Key Vault

### LINKED ISSUE #310: Azure KeyVault support
Azure KeyVault is pretty similar to KMS and can be used to wrap/unwrap keys. Are than plans to support it? I might have time to contribute the code, but I wanted to ask before starting to work on it.

carlpett:
This PR adds support for [Azure Key Vault](https://azure.microsoft.com/en-us/services/key-vault/) as a MasterKey implementation.

Adding some comments below where I have some minor questions, and very happy for a review since this is my first change in this code base.

Fixes #310.


azkv/keysource.go
```diff
@@ -0,0 +1,187 @@
...
+
+// NewMasterKeyFromResourceID takes an Azure Key Vault key URL and returns a new MasterKey
+// URL format is {vaultUrl}/keys/{key-name}/{key-version}
+func NewMasterKeyFromURL(url string) *MasterKey {
+	k := &MasterKey{}
+	re := regexp.MustCompile("^(https://[^/]+)/keys/([^/]+)/([^/]+)$")
+	parts := re.FindStringSubmatch(url)
+	if parts == nil || len(parts) < 2 {
+		log.Error("No match!")
+		// !?
```
carlpett: None of the others seem to handle errors in their corresponding functionality. This will cause a panic if it cannot extract the parts, is this acceptable or should I refactor it somehow?

azkv/keysource.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// Encrypt takes a sops data key, encrypts it with Key Vault and stores the result in the EncryptedKey field
+func (key *MasterKey) Encrypt(dataKey []byte) error {
+	c, err := newKeyVaultClient()
+	if err != nil {
+		return err
+	}
+	data := base64.RawURLEncoding.EncodeToString(dataKey)
+	p := keyvault.KeyOperationsParameters{Value: &data, Algorithm: keyvault.RSAOAEP256}
```
carlpett: The choices of algorithm is RSA-OAEP, RSA-OAEP-256 and RSA-1_5. My understanding is that RSA-1_5 is broken, but I don't understand the deeper implications of the different hashing algorithms in use between OAEP and OAEP-256.

azkv/keysource_test.go
```diff
@@ -0,0 +1,102 @@
...
+	assert.Equal(t, map[string]interface{}{
+		"vaultUrl":   key.VaultURL,
+		"key":        key.Name,
+		"version":    key.Version,
+		"enc":        "this is encrypted",
+		"created_at": "2016-10-31T10:00:00Z",
+	}, key.ToMap())
+}
+
+var azureKeyAcceptanceTestUrl = flag.String("azure-key", "", "URL to Azure Key Vault (note that this can incur real costs!)")
```
carlpett: Setting this flag will cause a real interaction with the Azure service, including the call-costs.

cmd/sops/main.go
```diff
@@ -78,6 +79,10 @@ func main() {
    (you need to setup google application default credentials. See
     https://developers.google.com/identity/protocols/application-default-credentials)
 
+   To encrypt or decrypt a document with Azure Key Vault, specify the
+   Azure Key Vault key URL in the --azure-kv flag or in the SOPS_AZURE_KEYVAULT_URL
+   environment variable.
```
carlpett: Noticed now that I'm missing the environment variables used to authenticate. Will add that.

cmd/sops/main.go
```diff
@@ -176,13 +185,18 @@ func main() {
...
 						kmsArns := c.StringSlice("kms")
+						azkvs := c.StringSlice("azkv")
 						var group sops.KeyGroup
 						for _, fp := range pgpFps {
 							group = append(group, pgp.NewMasterKeyFromFingerprint(fp))
 						}
 						for _, arn := range kmsArns {
 							group = append(group, kms.NewMasterKeyFromArn(arn, kms.ParseKMSContext(c.String("encryption-context"))))
 						}
+						// NOTE: Why isn't GCP here?
```
carlpett: I was confused by GCP not being present here - is there some reasoning for when it is needed and when not?

carlpett:
Oh, and I apparently didn't add the vendor dirs.

azkv/keysource.go
autrilla: I think all of those are considered secure, but I'd defer to @jvehent on that.

My main concern is that whatever we pick, the way this is implemented right now, it will be impossible to upgrade the algorithm in the future without breaking old files. Perhaps it would be worth it to default to RSA-OAEP-256, but store the used algorithm in the MasterKey, so that we use the same algorithm when decrypting as encrypting.

This isn't really a blocker to merge this, though. We could just make this change whenever we need to change the algorithm.

carlpett:
@autrilla Thanks for the review! I rolled back the vendoring, and you should already have push access on the branch. 
I'm fixing the parse error handling - is it more appropriate to skip invalid keys, or should it lead to exiting with error?

autrilla:
I think erroring is better than skipping and showing a warning. It'd be very easy to miss the warning and end up with a file missing keys.

carlpett:
Agreed. Pushed with this functionality now. Fixed the missing GCP thing as well.

carlpett:
@autrilla Could a new release be made including this code? :)

autrilla:
@jvehent ^, or ideally let me know what needs to be done to cut a new release

carlpett:
@autrilla See #360 before, ideally...

PR #362, opened 2018-06-21
Add readme for Azure

carlpett:


README.rst
```diff
@@ -207,23 +207,59 @@ And decrypt it using::
...
 	 $ sops --decrypt test.enc.yaml
 
+Encrypting using Azure Key Vault
+~~~~~~~~~~~~~~~~~~~~~~~~
+The Azure Key Vault integration uses service principals to access secrets in
+the vault. The following environment variables are used to authenticate:
+
+.. code:: bash
+
+	AZURE_TENANT_ID
```
autrilla: Where do I get these from?

I understand you can generate "identities" for applications, but most people will want to operate as their user. How do they do that?

carlpett: Yeah, I'd like that too... But from my understanding of the Azure SDK, you always need a client for your application. For example, the `az` CLI creates a new application identity/service principal for itself. 
The AZURE_USERNAME and AZURE_PASSWORD env vars are read, but you still need to provide a client id, and I haven't been able to figure out which client id works with those.

So as far as I know, you cannot use your normal user/password :disappointed:
Not all that deep into their SDKs yet, though, so I may be missing something. 

I'll add some docs on how to generate an SP, though.

carlpett: Ok, so there is something called a device flow, which can also be used (and which is what the az cli does to generate its SP). Perhaps that might be a good thing for interactive use

autrilla: A link to the Azure docs on what goes into each env var and how to generate each thing would be ideal. What's important is that users who aren't familiar with Azure (such as myself) have the basics to know how to use SOPS with it. For example, someone who's been using Azure for years surely knows how to get local tools to talk to Azure services, but I have no idea. I expected it to just work after doing `az login`. For example, this is why we explain you need your AWS credentials in `~/.aws` in the README. 

In summary, we need users who might be new to using third party CLI tools with their cloud provider to be able to use SOPS just by reading our docs (and the docs we link to from our docs).

Does this make sense?

I agree that we don't need the `device flow` on this PR. I just want users to be able to use Azure by reading this docs somehow.

carlpett: Agree in full! Did you see the new commit I pushed? It adds the cli invokations for creating a new principal, and how the output maps to the environment variables. I'll see if I can find some extra official docs to link to. They tend to not be quite focused on where to click in the portal, though...

autrilla: Oh, hah! You pushed this after I started writing this and the page didn't update. It's perfect now. Thanks!

README.rst
```diff
@@ -207,23 +207,77 @@ And decrypt it using::
...
+To create a Key Vault and assign your service principal permissions on it
+from the commandline:
+
+.. code:: bash
+
+	$ az keyvault --name sops --resource-group my-rg --location westeurope
+	$ az keyvault key create --name sops-key --vault-name sops --protection software --ops encrypt decrypt
+	$ az keyvault set-policy --name sops --resource-group my-rg --object-id $AZURE_CLIENT_ID \
+		--key-permissions encrypt decrypt
+	$ az keyvault key show --name sops-key --vault-name sops --query key.kid
```
autrilla: After following these instructions, I get permission denied here. Are you sure this is all you need?

README.rst
```diff
@@ -207,23 +207,77 @@ And decrypt it using::
...
+.. code:: bash
+
+	AZURE_TENANT_ID
+	AZURE_CLIENT_ID
+	AZURE_CLIENT_SECRET
+
+You can create a service principal using the cli like this:
+
+.. code:: bash
+	$ az ad sp create-for-rbac -n my-keyvault-sp
```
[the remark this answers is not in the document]
carlpett: Fixed!

README.rst
```diff
@@ -207,23 +207,77 @@ And decrypt it using::
...
+a key. This has the following form::
+
+	https://${VAULT_URL}/keys/${KEY_NAME}/${KEY_VERSION}
+
+To create a Key Vault and assign your service principal permissions on it
+from the commandline:
+
+.. code:: bash
+
+	$ az keyvault --name sops --resource-group my-rg --location westeurope
```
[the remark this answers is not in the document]
carlpett: Whoops, yes, copy/paste mishap. Fixed it.
Also added instructions for created a resource group

README.rst
```diff
@@ -207,23 +207,77 @@ And decrypt it using::
...
+	$ az keyvault key create --name sops-key --vault-name sops --protection software --ops encrypt decrypt
+	$ az keyvault set-policy --name sops --resource-group my-rg --object-id $AZURE_CLIENT_ID \
```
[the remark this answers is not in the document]
carlpett: No, the policy is actually for the entire vault, not per-key.

README.rst
carlpett: One of the policy flags was wrong (object-id vs spn). Not sure why it worked for me, I must have done something extra that I missed when writing this

carlpett:
Updated with corrected setup docs

README.rst
```diff
@@ -207,23 +207,84 @@ And decrypt it using::
...
+
+To create a Key Vault and assign your service principal permissions on it
+from the commandline:
+
+.. code:: bash
+
+	# Create a resource group if you do not have one:
+	$ az group create --name sops-rg --location westeurope
+	# Key Vault names are globally unique, so generate one:
+	$ keyvault_name=sops-$(uuidgen)
```
autrilla:
```
msrest.exceptions.ValidationError: Parameter 'vault_name' must conform to the following pattern: '^[a-zA-Z0-9-]{3,24}$'.
autrilla@iron ~/P/cloudops-deployment> echo $keyvault_name 
sops-7A04F824-9F70-43AE-ABF7-4F80188AA595
```

I think UUIDs are too long

autrilla:
The instructions work for me now, other than the UUID being too long. They Azure master key also works 🎉 !

README.rst
carlpett: Sigh, teaches me to do quick fixes... Pushing a small fix for it.

PR #372, opened 2018-08-15
[PYTHON] Fix compatibility with ruamel.yaml >= 0.15.52

rouge8:
ruamel.yaml 0.15.52 changed `CommentedMap` to no longer subclass `OrderedDict` and 0.15.55 changed `CommentedSeq` to no longer subclass `list`, which broke sops's logic for parsing YAML dicts vs lists.

This patch fixes that by checking `collections.abc.MutableMapping` and `collections.abc.MutableSequence` instead.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 setuptools>=18.8.1
 cryptography>=1.4
 boto3>=1.1.3
-ruamel.yaml>=0.11.7
+ruamel.yaml>=0.11.7,<0.12.0 ; python_version=="2.6"
+ruamel.yaml>=0.11.7 ; python_version>"2.6"
 ordereddict>=1.1
 simplejson>=3.8
 futures>=3.0.5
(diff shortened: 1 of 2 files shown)


autrilla:
We pin this to 0.11.7:

https://github.com/mozilla/sops/blob/34e58593e63a8babfbd29b5be83923772db58d5a/requirements.txt#L4

Should we increase the pinned version as well? Why do you want/need to use the newer version of `ruamel.yaml`?

rouge8:
> Should we increase the pinned version as well?

Ah, yes, didn't realize that was being used anywhere. Fixed.

> Why do you want/need to use the newer version of ruamel.yaml?

Wheel availability is a nice one. Also being able to use newer versions of ruamel.yaml in projects either using sops as a library or including as a dependency.

requirements.txt
```diff
@@ -1,7 +1,8 @@
 setuptools>=18.8.1
```
rouge8: this could probably mostly (or entirely) be replaced with `-e .` to install sops from `setup.py`...

autrilla:
Apparently I can't merge to `python-sops`, so r? @jvehent

PR #379, opened 2018-09-19
fixed symmetric

0xflotus:


autrilla:
Thanks a lot for sending a patch for this!

Issue #347, opened 2018-05-15
Better error message for audit db connection failing


ajvb:
At the moment, the error message for failing to connect to the audit db is:

```Pinging audit database failed: <error message>```

I've seen two cases where a user had a `/etc/sops/audit.yml` file and had forgotten about it and the above error is all they get.

Should include further information on where the file is, what's going on, etc.

olevchyk:
@autrilla can we consider this as solved ?

autrilla:
Yes! Thanks for reminding me!

PR #383, opened 2018-10-04
auditdb improve logging

olevchyk:
I'm trying to pick up easy  issues to improve my golang coding.   
This pull request should address: https://github.com/mozilla/sops/issues/347
Waiting for the feedback.


audit/audit.go
```diff
@@ -34,8 +37,24 @@ func init() {
 	if flag.Lookup("test.v") != nil {
 		return
 	}
+	var auditErrors []error
+	re := regexp.MustCompile("postgres://(.*?):(.*?)@(.*)/(.*)")
```
autrilla: I'm not sure we should perform this validation. In any case, I think this regex is wrong, because `postgres://localhost/db` should work (it should expand to `postgres://current_user@localhost:5432/db`).

I think it might be easier to just let the postgres library handle this. If the user provides an invalid connection string, it'll return an error that we can just display.

olevchyk: agree with that, but not matching regex will just print connection string as is, the only reason to have this regex is somehow to tell user which connection is failing and mask the password, assuming we have couple of databases defined. If I shouldn't worry about this I will happy to omit this part.

autrilla: If we're worried about that, I think maybe getting the error message returned by libpq and replacing the connection string with `<connection string redacted>` or something along those lines would be better. But I don't think showing the connection string on stderr is a big deal, so I don't think we need this.

audit/audit.go
```diff
@@ -34,8 +36,23 @@ func init() {
...
+		auditDb, err := NewPostgresAuditor(pgConf.ConnStr)
+		if err != nil {
+			auditErrors = append(auditErrors, errors.Wrap(err, fmt.Sprintf("connectStr: %s, err", pgConf.ConnStr)))
+		}
+		auditors = append(auditors, auditDb)
+	}
+	if len(auditErrors) > 0 {
+		log.Errorf("connecting to audit database, defined in %s", configFile)
+		for i, err := range auditErrors {
+			if i == len(auditErrors)-1 {
```
autrilla: Sorry, missed this on my first pass. I think it'd be better to fail if _any_ of the audit backends fail, not just if the last one fails.

Other than this, lgtm. Could you give me permission to push to your branch, so I can vendor `pkg/errors`? There should a be checkbox for you to enable that within the PR.

autrilla: We prefer to vendor dependencies ourselves because it'd be very easy for someone to include malicious code in the vendoring commit, as we don't review the diffs for those.

olevchyk: done  - "Allow edits from maintainers" is checked.

olevchyk:
thank you.

Issue #368, opened 2018-07-19
Create an encryption/decryption selector


andor-pierdelacabeza:
So, to help creating automatic tools and following diffs, there are parts of a yaml/json file that shouldn't be encrypted, and that cannot be done easily with encrypted/unencrypted_suffix.

My use case is Kubernetes yaml/json secrets definitions. Lets see an example:

```yaml
apiVersion: v1
kind: Secret
metadata:
 namespace: staging
 name: web-secrets
type: Opaque
data:
 GOOGLE_CLIENT_ID: yadayadayada
 GOOGLE_CLIENT_SECRET: yadayadayada
```
Here, the only part that needs to be encrypted is the `data` section. Having `metadata`, `type` and the other sections unencrypted helps to follow changes, especially by people with no permission to encryption, and automatic tools.

Doing this with `unencrypted_suffix` is difficult, as I can only select one key.
Also, doing it with `encrypted_suffix` is difficult too, as, if I select the `data` suffix, I'll encrypt both `data` and `metadata`.

I can't alter the structure of this yaml, as it's inherent to Kubernetes, so it'd be great to have a feature that would allow me to select just one section of the yaml/json file, something like:

sops --encrypt --encryption-selector '.data' file.yml`

Extra:
 - Unencryption selectors
 - Multiple selectors

autrilla:
I think Kubernetes is a big enough use case that it might be simpler to just have a Kubernetes-specific store ("format") as we do for plain YAML and JSON files. What you propose is more flexible, but it feels harder to implement to me.

andor-pierdelacabeza:
It may fit another use cases, I don't know, lets see if anybody chips in :+1:  :)

stephenmoloney:
an alternative is to use bitnami sealed-secrets.

andor-pierdelacabeza:
I've seen now there is an [--encrypted-regex option](https://github.com/mozilla/sops#49encrypting-only-parts-of-a-file).

So, using a command like:
`sops --encrypt --gcp-kms [REDACTED] --encrypted-regex '^(data)$' secret.yml`

Would result in:

```
apiVersion: v1                                                                                                                                        
kind: Secret
metadata:
    namespace: staging
    name: web-secrets
type: Opaque
data:
[...]

PR #385, opened 2018-10-20
add --encrypted-selector option

dictcp:
per #368, sometimes we need to pick only a subtree to encrypt (eg. Kubernetes secret). then it should be able to encrypt Kubernetes secret just like https://github.com/shyiko/kubesec

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -429,6 +429,7 @@ func main() {
 
 		unencryptedSuffix := c.String("unencrypted-suffix")
 		encryptedSuffix := c.String("encrypted-suffix")
+		encryptedSelector := c.String("encrypted-selector")
 		conf, err := loadConfig(c, fileName, nil)
 		if err != nil {
 			return toExitError(err)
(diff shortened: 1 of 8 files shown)


sops.go
```diff
@@ -318,6 +318,13 @@ func (tree Tree) Encrypt(key []byte, cipher Cipher) (string, error) {
 				}
 			}
 		}
+		if tree.Metadata.EncryptedSelector != "" {
```
autrilla: This is good, but I think a better approach would be to convert the selector to a `[]string` (in `main.go`, so the SOPS API itself would take the `[]string`, and the command line will handle conversion) and then compare the slices. It saves us some computations we currently do for every decryption, and more importantly would provide a cleaner API to those using SOPS as a Go library.

What do you think?

cmd/sops/main.go
```diff
@@ -429,6 +429,7 @@ func main() {
 
 		unencryptedSuffix := c.String("unencrypted-suffix")
 		encryptedSuffix := c.String("encrypted-suffix")
+		encryptedSelector := c.String("encrypted-selector")
```
autrilla: You need to add the flag to the list of flags here: https://github.com/mozilla/sops/blob/457297faa309c5a0b4536ef5c808b5324e025fb6/cmd/sops/main.go#L409

jvehent:
This PR is stale and conflicts with a lot of other files. Should we close it?

autrilla:
It can be reopened if needed, so I guess so.

PR #387, opened 2018-10-23
Added an output flag for saving the output to a specified file

### LINKED ISSUE #386: Add `--output-file` option
When automating sops operations, it would be useful to be able to specify an output file, rather than redirecting stdout. Also, a shell isn't always available to do the redirecting, for example when running sops as a Cloud Build step to decrypt application credentials during a deploy.

hellozee:
Fixes #386


cmd/sops/main.go
```diff
@@ -419,7 +418,7 @@ func main() {
 		fileName := c.Args()[0]
 		if _, err := os.Stat(fileName); os.IsNotExist(err) {
 			if c.String("add-kms") != "" || c.String("add-pgp") != "" || c.String("add-gcp-kms") != "" || c.String("add-azure-kv") != "" ||
-				c.String("rm-kms") != "" || c.String("rm-pgp") != "" || c.String("rm-gcp-kms") != "" || c.String("rm-azure-kv") != "" {
+				c.String("rm-kms") != "" || c.String("rm-pgp") != "" || c.String("rm-gcp-kms") != "" || c.String("rm-azure-kv") != "" || c.String("output") != "" {
```
autrilla: I'm not sure why this change is needed, can you explain?

cmd/sops/main.go
```diff
@@ -619,6 +618,19 @@ func main() {
...
+				return common.NewExitError(fmt.Sprintf("Could not open output file for writing: %s", err), codes.CouldNotWriteOutputFile)
+			}
+			defer file.Close()
+			_, err = file.Write(output)
+			if err != nil {
+				return toExitError(err)
+			}
+			log.Info("Output File written successfully")
+		}
 		_, err = os.Stdout.Write(output)
```
autrilla: Hmm, so this will write to stdout as well as the file. Maybe it'd be best to introduce a variable here that's an `io.Writer`. Something like this (no guarantee this will compile):

```go
var outputFile io.Writer
if c.String("output") != "" {
  // ...
  outputFile = file
} else {
  outputFile = os.Stdout
}
outputFile.Write(output)

File: cmd/sops/main.go (hunk 1 above)
Comment (hellozee) [reply cmd/sops/main.go]: Well I myself can't explain the change now, probably saw the other flags so I added mine too, :wink: :stuck_out_tongue:

File: cmd/sops/main.go (hunk 2 above)
Comment (hellozee) [reply cmd/sops/main.go]: I think there should be a more elegant way of doing this

Comment (autrilla) [reply cmd/sops/main.go]: Hmm, I'd think what I propose would be pretty idiomatic Go code. What makes you think otherwise? I'm certainly open to other options here!

**Comment by hellozee** [comment]
@autrilla sure would love to add this flag to the documentation, can you point out to where should I add it in the readme? 
If no one is onto writing a test for it, I would like to give it a try, learning a new language isn't a bad thing afterall

File: cmd/sops/main.go (hunk 2 above)
Comment (autrilla) [reply cmd/sops/main.go]: Also, now that I look at the code, I think this would be even better, because after this, we'll have three possible files to output to:

1. Standard out
2. The input file, with `--in-place`
3. The output file specified with `--output`

I'm not sure allowing combinations of these is a good idea. E.g. `sops --output bar.yaml --in-place --decrypt foo.yaml` would write the output to two different places.

**Comment by autrilla** [comment]
@hellozee I'd add it under "Usage", after section 2.12 (or whatever the current latest is).

You can certainly go ahead and write a test for it!

File: cmd/sops/main.go (hunk 2 above)
Comment (jvehent) [reply cmd/sops/main.go]: > I'm not sure allowing combinations of these is a good idea.

It definitely isn't and will surprise the user. If both `--in-place` and `--output` are set, we should exit with an error. If either are set, only logs should be displayed on the terminal.

Comment (hellozee) [reply cmd/sops/main.go]: @jvehent what kind of error message should we print in case of the above mentioned situation?

File: cmd/sops/main.go [hunk 3]
```diff
@@ -415,6 +415,9 @@ func main() {
 		if c.NArg() < 1 {
 			return common.NewExitError("Error: no file specified", codes.NoFileSpecified)
 		}
+		if c.Bool("in-place") && c.String("output") != "" {
+			return common.NewExitError("Error: cannot operate on both --output and --in-file", codes.ErrorConflictingParameters)
```
autrilla: typo: `--in-file` instead of `--in-place`

cmd/sops/main.go
```diff
@@ -415,7 +415,7 @@ func main() {
 		if c.NArg() < 1 {
 			return common.NewExitError("Error: no file specified", codes.NoFileSpecified)
 		}
-		if c.Bool("in-place") && c.String("output") != "" {
+		if c.Bool("in-file") && c.String("output") != "" {
```
autrilla: No, I meant the correct string is `--in-place`, as that's the value of the flag for operating on the file in-place. We don't use `--in-file`.

hellozee: Oh sorry my bad

hellozee:
@autrilla Did I miss something?

autrilla:
@hellozee I thought you wanted to add the functional tests for this. Let me know if you don't and I'll do it myself.

hellozee:
@autrilla ohh I tried a couple of times to learn rust, but it didn't go too far, thought my C++ background would but nevertheless, writing tests in rust is not my thing, sadly, :wink:

Issue #70, opened 2016-05-05
INI format support


imsky:
would be great to have ini support - much simpler format

jvehent:
You mean support `ini` for the `.sops.yaml` configuration file documented at https://github.com/mozilla/sops#using-sops-yaml-conf-to-select-kms-pgp-for-new-files ?

I don't think that's a problem as long as the loaded configuration fits into the same dictionary. But, to be honest, that `yaml` formal is so trivial I don't really see a benefit to supporting `ini` as well.

imsky:
i meant ini as another option vs. json/yaml for encrypted files

jvehent:
Ah, that's a different story then :smile: 
I'm not familiar enough with the `ini` format to say that it's possible to do. Do you have a link to a formal specification? If `ini` can be loaded as a key/value tree like yaml and json, then we should be able to encrypt the leaves.

Of course, you can always treat `ini` files as binary, but you lose the readability of keeping the file structure in cleartext.

imsky:
i don't think there's a formal spec (not one i could easily find anyway). https://docs.python.org/3/library/configparser.html is probably all that's necessary here. i think key/value tree access works out of the box.

jcassee:
How about supporting simple .env files first? This is literally the syntax:
```
VAR1=val1
VAR2=val2
```

I'm looking to use sops to decrypt secrets for [Kustomize](https://github.com/kubernetes-sigs/kustomize), and encrypting only values for .env files sounds like a nice quality of life feature.

I have very little experience with Go, though. @jvehent some pointers on how to add this file format?

autrilla:
@jcassee you essentially need to write something that implements the [`Store` interface](https://github.com/mozilla/sops/blob/master/sops.go#L442-L480) and then plug that in to the command line tool. Here's the [YAML store](https://github.com/mozilla/sops/blob/master/stores/yaml/store.go) which you can use as an example.

The main challenge here is that SOPS needs to store its metadata (a non-flat structure) with the file, so you'd have to figure out how to store that for .env files. For YAML and JSON, we just use a `"sops"` top level map entry.

jcassee:
@autrilla Thanks for the pointers. #391 implements the new store.

ajvb:
Fixed with https://github.com/mozilla/sops/pull/400

PR #391, opened 2018-10-29
Add .env store implementation

jcassee:
This is an implementation for #70. This is my first contribution to a Go project, so I welcome all comments, including stylistic and nitpicking ones.

The code still includes a couple of FIXMEs, most of which are questions I'd like some feedback on.


stores/env/store.go
```diff
@@ -0,0 +1,119 @@
...
+		buffer.WriteString(line)
+	}
+	return buffer.Bytes(), nil
+}
+
+func (Store) EmitValue(v interface{}) ([]byte, error) {
+	// FIXME: Whot should this function do?
+	panic("implement me")
+}
+
```
jcassee: The code below is an attempt to encode the metadata. Is this the best way?

cmd/sops/common/common.go
```diff
@@ -12,6 +12,7 @@ import (
 	"go.mozilla.org/sops/keyservice"
 	"go.mozilla.org/sops/stores/json"
 	"go.mozilla.org/sops/stores/yaml"
+	"go.mozilla.org/sops/stores/env"
```
jvehent: I don't think we should call this "env". It's too generic and confusing compared to regular env variables (which aren't files).

stores/stores.go
```diff
@@ -27,60 +27,60 @@ type SopsFile struct {
 	// in the SOPS file by checking for nil. This way we can show the user a
 	// helpful error message indicating that the metadata wasn't found, instead
 	// of showing a cryptic parsing error
-	Metadata *Metadata `yaml:"sops" json:"sops"`
+	Metadata *Metadata `yaml:"sops" json:"sops" env:"sops"`
```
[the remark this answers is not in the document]
autrilla: This is indeed not needed

cmd/sops/common/common.go
autrilla: Agreed. Maybe 'dotenv' would be a better name.

stores/env/store.go
autrilla: It is a way :) It has the downside that it's not obvious what options the file has, which is easy to see with the JSON/YAML implementations currently. Maybe it'd be possible to instead "flatten" the metadata struct into fields, e.g.

```
SOPS_VERSION=3.1.1
SOPS_KMS_KEY_1=blah
```

But this can get very annoying to do.

jcassee: I'll see how far I can carry that idea.

stores/stores.go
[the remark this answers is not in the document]
jcassee: I'll revert.

cmd/sops/common/common.go
jcassee: I like "dotenv". Will change.

stores/env/store.go
jcassee: I used `_metadata`, but I think `sops` is better because 1) lowercase is often ignored by parsers of dotenv files, 2) it is also used by the YAML and JSON file types.

It is relatively easy to encode the string field of [Metadata](https://github.com/mozilla/sops/blob/7c72237f5a49f7b8815efababe55e00e60fecf4f/sops.go#L425) into `sops_version` etc. (See the current code.) However, writing this out for the structured data does not seem to be the way to go. Those raw strings at the end of the metadata are starting to make sense. :-) I think there is a stable translation possible like you suggest but I don't know how to write something like that in Go.

autrilla: Yeah, flattening the structure isn't exactly trivial. Since the main concern here is being able to easily see the metadata with a normal text editor, maybe it would be better to just encode the metadata to JSON (basically the same thing the JSON store does) and shove it into a field in the dotenv file. It's not great, because it will all be in one line, but imo it's better than having an essentially unreadable base64. Many differs can handle same-line diffs just fine, so I think it's at least an improvement.

stores/dotenv/store.go
```diff
@@ -0,0 +1,152 @@
...
+
+const SopsPrefix = "sops_"
+
+// Store handles storage of dotenv data
+type Store struct {
+}
+
+func (store *Store) LoadEncryptedFile(in []byte) (sops.Tree, error) {
+	branch, err := store.LoadPlainFile(in)
+	if err != nil { return sops.Tree{}, err }
```
autrilla: Have you ran `gofmt` (or even better, `goimports`) on this file? Not sure if it forces it, but generally Go programmers put the if body in a new line, like so:

```go
if err != nil {
    return sops.Tree{}, err
}
```

stores/dotenv/store.go
```diff
@@ -0,0 +1,152 @@
...
+
+	resultBranch := make(sops.TreeBranch, 0)
```
autrilla: I'd just write `var resultBranch sops.TreeBranch` here

stores/dotenv/store.go
```diff
@@ -0,0 +1,152 @@
...
+	metadata := stores.Metadata{}
+	md_found := false
```
autrilla: Go uses camelCase for local variables, so `mdFound` or even `metadataFound` would be better here. I personally prefer longer names, but many Go programmers prefer shorter. Up to you.

stores/dotenv/store.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil { return sops.Tree{}, err }
+
+	return sops.Tree{
+		Branch:   resultBranch,
+		Metadata: internalMetadata,
+	}, nil
+}
+
+func (store *Store) LoadPlainFile(in []byte) (sops.TreeBranch, error) {
+	branch := make(sops.TreeBranch, 0)
```
autrilla: Same as above, I'd use `var branch sops.TreeBranch` here.

stores/dotenv/store.go
```diff
@@ -0,0 +1,152 @@
...
+	reader := bytes.NewReader(in)
```
autrilla: [`bytes.Split`](https://golang.org/pkg/bytes/#Split) should work here as well, instead of needing to use buffered IO.

Simple example: https://play.golang.org/p/NpP1IcidSIt

stores/env/store.go
jcassee: Your comment got me thinking. It's possible to [convert the metadata struct to a map](https://stackoverflow.com/a/42849112). I can write a function that converts that to flat key/value items and back. I'll give that a try and if I get stuck I'll just do a single JSON item.

cmd/sops/main.go
```diff
@@ -397,7 +398,8 @@ func main() {
 		},
 		cli.StringFlag{
 			Name:  "set",
-			Usage: `set a specific key or branch in the input JSON or YAML document. value must be a json encoded string. (edit mode only). eg. --set '["somekey"][0] {"somevalue":true}'`,
+			// FIXME: Warn about complex value types and dotenv files?
```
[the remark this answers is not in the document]
jcassee: Okay, sounds good.

Related to that, the code and docs often mention "YAML and JSON" filetypes. What shall we do with that, add "dotenv" or remove the explicit mention of file formats?

stores/dotenv/store.go
jcassee: I'll run `goimports` on it, thanks for the pointer.

cmd/sops/main.go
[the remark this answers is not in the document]
autrilla: Yeah, as we add more file formats, I think we should change mentions from "YAML and JSON" to "multiple file formats", or whatever is relevant. Eventually having a list of ten different supported formats will get repetitive 😄

stores/env/store.go
autrilla: Yeah, but I think you'll run into the same issue. The map won't be a flat structure either.

stores/dotenv/store.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func mapToMetadata(m map[string]interface{}) (stores.Metadata, error) {
+	m = unflatten(m)
+	var md stores.Metadata
+	inrec, err := json.Marshal(m)
+	if err != nil {
+		return md, err
+	}
+	err = json.Unmarshal(inrec, &md)
```
jcassee: This line returns the error "json: cannot unmarshal object into Go struct field Metadata.pgp of type []stores.pgpkey". Seems the JSON field translation does not work. @autrilla any idea how to fix this?

jcassee: Never mind, I know what the problem is. :smile:

jcassee:
Commit a05f862 adds flattening of the metadata into dotenv variables. It was quite tricky to write and might be too convoluted. Worth it?

autrilla:
Awesome, thanks for taking the time to implement that. It is indeed a bit convoluted, but I think it's a good thing to have. It'll be useful for other formats that don't support nesting (e.g. INI). I'll see if I can simplify the implementation a bit or make it easier to read, and see if I can put it somewhere so it can be reused by other stores.

jcassee:
Cool, I'm very interested to see it simplified. :smile: Only things left are probably documentation updates and maybe more tests. What needs can I improve before this PR can be merged?

autrilla:
@jcassee I have refactored your code and rudely pushed the refactor onto your branch. It'd be good if you could review my change in turn and let me know if you find any issues. We can even revert my commit if you want 😄 . It's slightly more complex than your code because it addresses a bug your code had, namely the handling of map keys that were stringified numbers (think `"0"`). We need to use different separators for maps and lists for this reason.

jcassee:
Thanks for that. No problem pushing to my branch, I like working together on code. Smart idea to use different separators for lists and maps and separating tokenization from unflattening. There is some almost-but-not-quite code duplication in unflatten that I have a feeling could be fixed using some function-oriented magic, but haven't spent enough time on that.

I made some small changes, but feel free to revert or change even more. This should be code that you are comfortable maintaining.

autrilla:
Awesome, thanks for the style changes. yeah, there's a bunch of duplication but I think it's pretty hard to remove without macros. I tried splitting it out into functions but it ends up being about the same amount of lines.

I'm going to give this a final review pass and merge if everything is alright 🎉

jcassee:
@autrilla do you know when a version with this change will be released?

autrilla:
@jcassee 3.2.0, which was just released 😄

jcassee:
@autrilla Great, thanks a lot!

PR #392, opened 2018-11-08
Add multidoc encrypt/decrypt for YAML sources

jar-o:
Hi this adds multidoc (or "streams") support for YAML files with multiple documents, e.g.

```
key: value
---
key: value
key2: value
```

It works and passes existing tests, but I wanted to get feedback before going too much further in case my approach needs revising. I still need to add tests and update docs.

The significant changes codewise I think are:

* I've added `Branches` (`[]TreeBranch`) to the `Tree` struct. If the document format understands and has "multidocs", this is used, otherwise it will use the existing `Branch` field -- even for single document YAML files.
* The `Emitter` and `Loader` interfaces have changed to accommodate `Branches`
* This required an update to `vendor/github.com/mozilla-services/yaml` and those sources have been updated (I made a previous PR to that repo already)


autrilla:
Thanks for taking the time to add this feature!

The approach looks good in general. I think it would be simpler to get rid of the `Branch` field altogether, and let stores handle multiple branches if they can, or they can error out otherwise. It's easy for stores to get the first branch anyway, and we could always add a helper function to the stores package to get the first branch or an error if there's more than one branch. I think that'd make the data types and emitter and loader interfaces simpler.

What do you think?

As a side note, this will only store the SOPS metadata in the first YAML document, right? That seems fine to me.

jar-o:
Ok, yeah ditching `Branch` makes sense. I'll update. Oh and I meant to mention that in my original comments, but you're right -- the SOPS metadata is computed for the physical file (i.e. `hash` etc) rather than each individual stream. It seemed simplest and didn't seem necessary to change.

jar-o:
ok, I've removed `tree.Branch` in favor of `tree.Branches`, updated docs, and added tests. please let me know if there are any other changes I should make.

sops.go
```diff
@@ -464,7 +491,7 @@ type EncryptedFileEmitter interface {
 // to emit plain text files from the internal SOPS representation so that they can be
 // shown
 type PlainFileEmitter interface {
-	EmitPlainFile(TreeBranch) ([]byte, error)
+	EmitPlainFile(Tree) ([]byte, error)
```
autrilla: I think this should still just take `[]TreeBranch` (or `TreeBranches`), as it doesn't use the metadata.

sops.go
```diff
@@ -284,50 +287,64 @@ func (branch TreeBranch) walkBranch(in TreeBranch, path []string, onLeaves func(
...
 			}
+			return in, nil
+		})
+		return err
+	}
+
+	for _, branch := range tree.Branches {
+		err = walk(branch)
+		if err != nil {
+			break
```
autrilla: We could just return the error here instead of breaking and checking the error again later.

sops.go
```diff
@@ -342,59 +359,69 @@ func (tree Tree) Decrypt(key []byte, cipher Cipher) (string, error) {
...
-			}
-			hash.Write(bytes)
+			return v, nil
+		})
+		return err
+	}
+	for _, branch := range tree.Branches {
+		err = walk(branch)
+		if err != nil {
+			break
```
autrilla: Same as above

stores/yaml/store.go
```diff
@@ -118,45 +122,89 @@ func (store *Store) LoadEncryptedFile(in []byte) (sops.Tree, error) {
...
-	for i, item := range data {
+	for i, item := range data[0] {
 		if item.Key == "sops" {
-			data = append(data[:i], data[i+1:]...)
+			data[0] = append(data[0][:i], data[0][i+1:]...)
 		}
 	}
+
+	var branches sops.TreeBranches
+	for _, doc := range data {
```
autrilla: I feel like either this or the previous for loop are redundant, no? We only store the metadata in the first document, and we're first removing the metadata entries from the first tree and then from *all* trees.

stores/yaml/store.go
```diff
@@ -118,45 +122,89 @@ func (store *Store) LoadEncryptedFile(in []byte) (sops.Tree, error) {
...
+			}
+			yamlMap := store.treeBranchToYamlMap(branch)
+			yamlMap = append(yamlMap, yaml.MapItem{Key: "sops", Value: stores.MetadataFromInternal(in.Metadata)})
+			tout, err := (&yaml.YAMLMarshaler{Indent: 4}).Marshal(yamlMap)
+			if err != nil {
+				return nil, fmt.Errorf("Error marshaling to yaml: %s", err)
+			}
+			out = append(out, tout...)
+		}
+	} else {
```
autrilla: Why is this `if ... else` needed? The loop you wrote would just emit the first branch and not add the  `---\n`, right? I don't think we need to special-case the 'single-branch' scenario with how you wrote the code.

stores/yaml/store.go
```diff
@@ -118,45 +122,89 @@ func (store *Store) LoadEncryptedFile(in []byte) (sops.Tree, error) {
...
+				out = append(out, "---\n"...)
+			}
+			yamlMap := store.treeBranchToYamlMap(branch)
+			tmpout, err := (&yaml.YAMLMarshaler{Indent: 4}).Marshal(yamlMap)
+			if err != nil {
+				return nil, fmt.Errorf("Error marshaling to yaml: %s", err)
+			}
+			out = append(out[:], tmpout[:]...)
+		}
+	} else {
```
autrilla: Same as above, I'm not sure why we need to handle the 'single-branch' case separately.

sops.go
jar-o: oh I see. ok

stores/yaml/store.go
jar-o: ah looks like I just missed this in the refactor. thanks for the catch, I'll fix.

jar-o:
ok I believe I addressed all your comments. Also added back a section to docs for "YAML Streams". Please let me know if there is anything else missing or in need of straightening out. Thanks. :)

jar-o:
certainly! thanks for the guidance.

PR #399, opened 2018-11-27
[PYTHON] Fix --extract on Python 3

rouge8:
Previously, `--extract` would fail on Python 3:

```
  File "/Users/andy/.virtualenvs/sops-py3/bin/sops", line 11, in <module>
    load_entry_point('sops', 'console_scripts', 'sops')()
  File "/Users/andy/forks/sops/sops/__init__.py", line 278, in main
    write_file(tree, path=dest, filetype=otype)
  File "/Users/andy/forks/sops/sops/__init__.py", line 1321, in write_file
    sys.stdout.write(tree.encode('utf-8'))
TypeError: write() argument must be str, not bytes
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ functional-tests-once:
 	fi;
 	echo "Testing --extract"
 	python sops/__init__.py -d --extract '["firstName"]' example.json
-	[ $$(python sops/__init__.py -d --extract '["firstName"]' example.json) == $$(python sops/__init__.py -d example.json | grep firstName | cut -d '"' -f 4) ]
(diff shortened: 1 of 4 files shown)


rouge8:
I'm not really sure what to do about the flake8 errors:

```
sops/__init__.py:897:10: W605 invalid escape sequence '\['
sops/__init__.py:901:6: W605 invalid escape sequence '\]'
sops/__init__.py:1425:7: W605 invalid escape sequence '\d'
```

The fix on Python 3 would be to change the regexes from `b''` strings to `rb''` strings, but that's invalid syntax on Python 2. On Python 2, we could fix it by changing to `r''` strings, but that will be the wrong type on Python 3. Maybe just ignore that error?

autrilla:
Yeah, feel free to add the appropriate # noqa comments to ignore that error.

rouge8:
Done ✅

PR #400, opened 2018-11-28
Added support for ini file store

oldpatricka:
Hi there!

This PR adds support for ini files, in an effort to implement https://github.com/mozilla/sops/issues/70. It looks like https://github.com/mozilla/sops/pull/391 was an attempt to implement ini, but .env is distinct from .ini as far as I can see. For example, .env doesn't support sections.

I've attempted to encode the sops metadata in a way that will be readable in diffs, but this is a bit challenging for ini files since things like arrays aren't really supported.

I'm happy to take any suggestions if something looks off. :)

### DIFF
--- a/stores/ini/store.go
+++ b/stores/ini/store.go
@@ -1,0 +1,369 @@
+package ini //import "go.mozilla.org/sops/stores/ini"
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"go.mozilla.org/sops"
+	"go.mozilla.org/sops/stores"
+	"strconv"
+	"reflect"
+	"gopkg.in/ini.v1"
+	"strings"
+	"sort"
+)
+
+// Store handles storage of ini data.
+type Store struct {
+}
+
+func (store Store) encodeTree(branches sops.TreeBranches) ([]byte, error) {
+	iniFile := ini.Empty()
+	for _, branch := range branches {
+		for _, item := range branch {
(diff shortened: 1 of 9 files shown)


oldpatricka:
HI there, I've cleaned up the metadata encoding to match the dotenv implementation, so it should be ready to review. 👋

stores/ini/store.go
```diff
@@ -0,0 +1,321 @@
...
+	"go.mozilla.org/sops"
+	"go.mozilla.org/sops/stores"
+	"gopkg.in/ini.v1"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+// Store handles storage of ini data.
+type Store struct {
```
autrilla: Is this file `goimports`/`gofmt`ed? I feel like they'd move the braces to the same line.

oldpatricka:
Thanks for the review! I'll get to that this week!

stores/ini/store.go
oldpatricka: Good, point. I was using an editor that doesn't automatically do this. I've run gofmt/goimports on the new files now.

oldpatricka:
@autrilla I've added an example file to the root of the directory, as well as a special tree for new ini files, since the existing example uses a tree structure with nested structures not supported by ini.

I've also added some clarification of filetypes supported to the README, but if you had something else in mind, I'd be happy to write it up.

autrilla:
Thanks a lot for taking the time to submit this patch!

Issue #381, opened 2018-09-24
add support for kms key aliases


snstanton:
It would be convenient be able to use a key alias as in:
`aws kms encrypt --key-id alias/my-key ...`
rather than having to provide the full arn.

autrilla:
Duplicate of #366

snstanton:
This isn't actually a duplicate of 366.  I was talking about the AWS key alias mechanism (e.g. `aws kms list-keys`) not just adding a string label in the .sops file.

autrilla:
Oh, cool, I didn't know that even existed! Reopening

rosscdh:
This does not seem to be fixed in `3.3.0` as documented in changelog not in HEAD of master

```
func (key MasterKey) createSession() (*session.Session, error) {
	re := regexp.MustCompile(`^arn:aws[\w-]*:kms:(.+):[0-9]+:(key|alias)/.+$`)
```
requires the full arn (which defies the point of an alias?)

autrilla:
I have no idea how that AWS feature works, but #415 shows you don't need to provide the full ARN. You need the account ID and the alias name. We still have to tell the AWS SDK that it's an alias somehow.

I'm not sure I understand why this is an issue for you. My guess (but it's just that, a guess) is that you're specifying the key through command line flags, which to me feels like doing it wrong. It is convenient for one-off operations and examples, but I don't think it really is how you should actually use `sops`. Using a `.sops.yaml` file is a better approach in my opinion.

rosscdh:
Hey there thanks for the reply.

I'll check the sops yaml setup (am using it via the kustomize sops plugin FYI) but i suspect based on the code i pasted above the full arn will still be necessary

Also re the purpose of a kms alias: its point appears to be to not require the region nor acc id as per https://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html

```
--key-id (string)

A unique identifier for the customer master key (CMK).

To specify a CMK, use its key ID, Amazon Resource Name (ARN), alias name, or alias ARN. When using an alias name, prefix it with "alias/" . To specify a CMK in a different AWS account, you must use the key ARN or alias ARN.

For example:

Key ID: 1234abcd-12ab-34cd-56ef-1234567890ab
Key ARN: arn:aws:kms:us-east-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab
Alias name: alias/ExampleAlias
[...]

PR #415, opened 2019-01-10
Support referencing AWS KMS keys by aliases

darend:
## what

Allow KMS keys to be referenced by alias instead of the complete ARN

Example use:
```
creation_rules:
        - kms: "arn:aws:kms:us-west-2:1234567890:alias/my-sops-key,
                arn:aws:kms:us-east-1:1234567890:key/ba2dad9b-032f-42bd-aa69-6dd4012931f4"
```

## why

Addresses issue https://github.com/mozilla/sops/issues/381

## testing

Unit tests are passing, its not clear how to run the functional tests


darend:
cc @snstanton

darend:
> lgtm. I'm curious why you prefer using aliases vs the full ARN, though. They seem to be about as hard (read: impossible) to memorize as the ARN, since you'd have to memorize the account id, right?

We have multiple environments, each with its own set of keys. When we generate the key we use the environment name as the alias. This makes it very easy to configure sops since our tooling can derive the list of keys.

autrilla:
I see, good to know!

Wolfsrudel:
> > lgtm. I'm curious why you prefer using aliases vs the full ARN, though. They seem to be about as hard (read: impossible) to memorize as the ARN, since you'd have to memorize the account id, right?
> 
> We have multiple environments, each with its own set of keys. When we generate the key we use the environment name as the alias. This makes it very easy to configure sops since our tooling can derive the list of keys.

It's not just that: when the key rotates, the ID changes and you have to adjust the SOPS file. However, with an alias, this change is obsolete as the alias automatically points to the new (rotated) ID.

PR #420, opened 2019-01-23
Add comment support to dotenv store

autrilla:
It seems most (all?) dotenv implementations support using `#` for comments, so let's do that.


autrilla:
cc @jcassee who implemented this initially

jcassee:
Nice clean change, I like it!

Super nitpicky: there is an asymmetry between [the switch statement in LoadEncryptedFile](https://github.com/mozilla/sops/pull/420/files#diff-c6c1ae711dba52512d7d13fc2026ec14R28) and [the cast in EmitPlainFile](https://github.com/mozilla/sops/pull/420/files#diff-c6c1ae711dba52512d7d13fc2026ec14R111). These pieces of code both check a value type, right?

PR #416, opened 2019-01-11
AWS Profile

marky-mark:
closed #389 

Added aws_profile

    * Decrypt
    * Command Line
    * Master Key
    * Encrypt
    * .sops.yaml

Tested locally, the tests pass except for 1...i think i'm missing an imported key

PR #422, opened 2019-01-24
Revert "AWS Profile"

autrilla:
Reverts mozilla/sops#416

@marky-mark it looks like your PR broke SOPS assuming roles when not specifying a profile.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -326,23 +326,6 @@ When removing keys, it is recommended to rotate the data key using ``-r``,
 otherwise owners of the removed key may have add access to the data key in the
 past.
 
-KMS AWS Profiles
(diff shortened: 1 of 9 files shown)


jvehent:
How did this not get caught by the CI tests?

autrilla:
> How did this not get caught by the CI tests?

The CI tests only use the real GPG implementation, everything else is mocked because we don't have AWS/GCP/Azure creds to run tests. So the AWS implementation mock assumes assuming roles will work.

jvehent:
I think we should create test kms in aws and gcp and use them in CI.

marky-mark:
Left a comment, think it’s the credentials type

Issue #389, opened 2018-10-25
Support AWS profile setting in creation_rules


max-rocket-internet:
Hi,

We have a `.sops.yaml` file like this:

```
creation_rules:
  - path_regex: secrets/environments/production/secrets.yaml
    kms: 'arn:aws:kms:us-east-1:666666666:key/xxxx-x-xxxxx'
  - path_regex: secrets/environments/staging/secrets.yaml
    kms: 'arn:aws:kms:us-east-1:3333333333:key/xxxx-x-xxxxx'
```

We have many different keys spread across different accounts. Normally when we are switching between working with different files we have to change our `AWS_PROFILE` environment variable to match a profile that is associated with the account that has the KMS key. It's fine but it's a bit of a pain.

It would be cool if we could have this:

```
creation_rules:
  - path_regex: secrets/environments/production/secrets.yaml
    kms: 'arn:aws:kms:us-east-1:666666666:key/xxxx-x-xxxxx'
    aws_profile: 'production'
  - path_regex: secrets/environments/staging/secrets.yaml
    kms: 'arn:aws:kms:us-east-1:3333333333:key/xxxx-x-xxxxx'
    aws_profile: 'staging'
```

Then we would never have to worry about setting and changing the `AWS_PROFILE` environment variable 🙂

jvehent:
It's potentially an interesting feature. We don't use profiles like that because we require MFA, so changing profiles is done by https://github.com/oremj/aws-assume-role. That said, if someone wants to implement profile support in a multi-cloud way (it should work with aws, gcp, azure, etc), I don't have any issue with merging it.

max-rocket-internet:
We also use assumed roles but use SSO via SAML to Google. Also like Okta or any other SAML SSO solution. 2FA is enforced on our Google accounts in this arrangement:

```
$ aws sts get-caller-identity
{
    "UserId": "AROAYYYYYXXXXXXXXX:first.last@domain.com",
    "Account": "6666666666",
    "Arn": "arn:aws:sts::6666666666666:assumed-role/sso-role1/first.last@domain.com"
}
```

PR #423, opened 2019-01-25
AWS Profile fix

marky-mark:
closes #422 #389 #416

I added in this part. Would something like this be ok? I tested locally and works for with and without aws_profile. 

```      
       config := aws.Config{Region: aws.String(matches[1])}

	if key.AwsProfile != "" {
		config.Credentials = credentials.NewSharedCredentials("", key.AwsProfile)
	}```


autrilla:
Seems to work fine now!

PR #430, opened 2019-02-26
Fixed panic when nano available, but vim is not installed

sergey-shambir:
Command `which vim nano` returns 1 when only nano editor is installed
The solutions is to lookup several times

Also improved error reporting: the old code violated two rules from [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments): "Don't panic" and "Error Strings"

### DIFF
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -314,6 +314,6 @@ func lookupAnyEditor(editorNames ...string) (editorPath string, err error) {
 			return editorPath, nil
 		}
 	}
-	err = fmt.Errorf("no editor available: sops supports %s", strings.Join(editorNames, ", "))
+	return "", fmt.Errorf("no editor available: sops attempts to use the editor defined in the EDITOR environment variable, and if that's not set defaults to any of %s, but none of them could be found", strings.Join(editorNames, ", "))
 	return "", err
 }
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"go.mozilla.org/sops/stores/ini"
 	"io/ioutil"
 	"os"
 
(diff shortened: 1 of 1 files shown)


autrilla:
Since we're changing this, I think we should actually just use [`os/exec.LookPath`](https://golang.org/pkg/os/exec/#LookPath) and avoid calling out to `which`. You're correct that the current code is broken though, `which` does appear to return 1 if any of the provided arguments is not found.

sergey-shambir:
I've updated commit: now it uses `exec.LookPath`

cmd/sops/edit.go
```diff
@@ -307,3 +306,14 @@ func runEditor(path string) error {
...
 }
+
+func lookupAnyEditor(editorNames ...string) (editorPath string, err error) {
+	for _, editorName := range editorNames {
+		editorPath, err = exec.LookPath(editorName)
+		if err == nil {
+			return editorPath, nil
+		}
+	}
+	err = fmt.Errorf("no editor available: sops supports %s", strings.Join(editorNames, ", "))
```
autrilla: We can improve this error message a bit, something like:

```suggestion
	return "", fmt.Errorf("no editor available: sops attempts to use the editor defined in the EDITOR environment variable, and if that's not set defaults to any of %s, but none of them could be found", strings.Join(editorNames, ", "))
```

cmd/sops/edit.go
```diff
@@ -307,3 +306,14 @@ func runEditor(path string) error {
...
+
+func lookupAnyEditor(editorNames ...string) (editorPath string, err error) {
+	for _, editorName := range editorNames {
+		editorPath, err = exec.LookPath(editorName)
+		if err == nil {
+			return editorPath, nil
+		}
+	}
+	return "", fmt.Errorf("no editor available: sops attempts to use the editor defined in the EDITOR environment variable, and if that's not set defaults to any of %s, but none of them could be found", strings.Join(editorNames, ", "))
+	return "", err
```
autrilla:
```suggestion
```

This line is superfluous now

autrilla:
Thanks for the patch!

PR #440, opened 2019-03-12
Add support for custom GPG key servers

rabadin:
Add support for the SOPS_GPG_KEYSERVER environment variable to
override the default GPG key server (gpg.mozilla.org).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -533,8 +533,9 @@ Example: place the following in your ``~/.bashrc``
 Specify a different GPG key server
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -529,6 +529,21 @@ Example: place the following in your ``~/.bashrc``
 
 	SOPS_GPG_EXEC = 'your_gpg_client_wrapper'
 
+
+Specify a different GPG key server
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, ``sops`` uses the key server ``gpg.mozilla.org`` to retrieve GPG keys.
```
autrilla: I'd note that it does this only if the key is not found in the local keyring

rabadin: Thanks for the review @autrilla. Makes sense, let me add that.

PR #444, opened 2019-03-19
Initial set of AWS KMS tests

ajvb:
Adds two functional tests that use AWS KMS.

The one using encryption context will flap, therefore I'm pointing this PR at a staging branch we can use to combine this PR, https://github.com/mozilla/sops/pull/435, and an incoming migration path PR.

### DIFF
--- a/functional-tests/src/lib.rs
+++ b/functional-tests/src/lib.rs
@@ -76,16 +76,19 @@ mod tests {
 
     #[test]
     fn encrypt_json_file_kms() {
+	let kms_arn = match env::var(KMS_KEY) {
+	    Ok(val) => val,
+	    _ => "".to_string(),
+	};
+        if kms_arn == "" {
+            return;
+        }
+
         let file_path = prepare_temp_file("test_encrypt_kms.json",
                                           b"{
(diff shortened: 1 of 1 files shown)


functional-tests/src/lib.rs
```diff
@@ -410,6 +446,43 @@ b: ba"#
...
 
+    #[test]
+    fn roundtrip_kms_encryption_context() {
+        let file_path = prepare_temp_file("test_roundtrip_kms_encryption_context.json",
+                                          b"{
+    \"foo\": 2,
+    \"bar\": \"baz\"
+}");
+
+	let kms_arn = match env::var(KMS_KEY) {
```
metadave: can this block be replaced with `env::var(KMS_KEY).expect(..)` ?

ajvb: Using `.expect()` when it was panicking makes sense, would I still want to use it with the change I made related to just ignoring the test if the env var is unset?

autrilla: No, since `.expect()` will panic if it's `None`.

PR #445, opened 2019-03-19
Autofix AWS KMS encryption context bug

ajvb:
This will prompt the user to automatically fix the bug that was described here: https://github.com/mozilla/sops/pull/435

### DIFF
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -159,13 +159,19 @@ func DetectKMSEncryptionContextBug(tree *sops.Tree) (bool, error) {
 	return false, nil
 }
 
-func GetKMSKeyWithEncryptionCtx(tree *sops.Tree) (kgndx int, keyndx int, key *kms.MasterKey) {
+func GetKMSKeyWithEncryptionCtx(tree *sops.Tree) (keyGroupIndex int, keyIndex int, key *kms.MasterKey) {
 	for i, kg := range tree.Metadata.KeyGroups {
 		for n, k := range kg {
 			kmsKey, ok := k.(*kms.MasterKey)
 			if ok {
 				if kmsKey.EncryptionContext != nil && len(kmsKey.EncryptionContext) >= 2 {
-					return i, n, kmsKey
+					duplicateValues := map[string]int{}
+					for _, v := range kmsKey.EncryptionContext {
+						duplicateValues[*v] = duplicateValues[*v] + 1
+					}
+					if len(duplicateValues) > 1 {
+						return i, n, kmsKey
+					}
 				}
 			}
 		}
@@ -181,6 +187,8 @@ type GenericDecryptOpts struct {
 	KeyServices []keyservice.KeyServiceClient
 }
 
+// LoadEncryptedFileWithBugFixes is a wrapper around LoadEncryptedFile which includes
+// check for the issue described in https://github.com/mozilla/sops/pull/435
 func LoadEncryptedFileWithBugFixes(opts GenericDecryptOpts) (*sops.Tree, error) {
 	tree, err := LoadEncryptedFile(opts.InputStore, opts.InputPath)
 	if err != nil {
@@ -201,30 +209,37 @@ func LoadEncryptedFileWithBugFixes(opts GenericDecryptOpts) (*sops.Tree, error) 
 	return tree, nil
 }
 
+// FixAWSKMSEncryptionContextBug is used to fix the issue described in https://github.com/mozilla/sops/pull/435
 func FixAWSKMSEncryptionContextBug(opts GenericDecryptOpts, tree *sops.Tree) (*sops.Tree, error) {
-	// TODO - Finalize messaging here.
 	message := "Up until version 3.3.0 of sops there was a bug surrounding the " +
-		"use of encryption context with AWS KMS."
+		"use of encryption context with AWS KMS." +
+		"\nYou can read the full description of the issue here:" +
+		"\nhttps://github.com/mozilla/sops/pull/435" +
+		"\n\nIf a TTY is detected, sops will ask you if you'd like for this issue to be " +
+		"automatically fixed, which will require re-encrypting the data keys used by " +
+		"each key." +
+		"\n\nIf you are not using a TTY, sops will fix the issue for this run.\n\n"
 	fmt.Println(wordwrap.WrapString(message, 75))
 
-	if !terminal.IsTerminal(int(os.Stdout.Fd())) {
-		return nil, NewExitError("Need to run this from a tty to resolve this issue.", codes.TTYRequired)
-	}
+	persistFix := false
 
-	var response string
-	for response != "y" && response != "n" {
-		fmt.Println("Would you like sops to automatically fix this issue? (y/n): ")
-		_, err := fmt.Scanln(&response)
-		if err != nil {
-			return nil, err
+	if terminal.IsTerminal(int(os.Stdout.Fd())) {
+		var response string
+		for response != "y" && response != "n" {
+			fmt.Println("Would you like sops to automatically fix this issue? (y/n): ")
+			_, err := fmt.Scanln(&response)
+			if err != nil {
+				return nil, err
+			}
 		}
-	}
-	if response == "n" {
-		return nil, fmt.Errorf("Exiting. User responded no.")
+		if response == "n" {
+			return nil, fmt.Errorf("Exiting. User responded no.")
+		} else {
+			persistFix = true
+		}
 	}
 
 	dataKey := []byte{}
-	success := false
 	// If there is another key, then we should be able to just decrypt
 	// without having to try different variations of the encryption context.
 	dataKey, err := DecryptTree(DecryptTreeOpts{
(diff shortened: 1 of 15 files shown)


version/version.go
```diff
@@ -10,15 +10,15 @@ import (
 	"gopkg.in/urfave/cli.v1"
 )
 
-const version = "3.2.0"
+const Version = "3.3.0"
```
ajvb: @autrilla Doing this here as it's needed to make the fix, but not sure if this PR is the right place to actually do this. Just wanted to call it out.

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return tree, nil
+}
+
+func FixAWSKMSEncryptionContextBug(opts GenericDecryptOpts, tree *sops.Tree) (*sops.Tree, error) {
+	// TODO - Finalize messaging here.
```
ajvb: Need to finalize this messaging before this is merged.

version/version.go
autrilla: I think it's fine to do it in this PR

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+		_, _, key := GetKMSKeyWithEncryptionCtx(tree)
+		if key != nil {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func GetKMSKeyWithEncryptionCtx(tree *sops.Tree) (kgndx int, keyndx int, key *kms.MasterKey) {
```
autrilla: I'd prefer more explicit names since it can be a bit confusing to people who aren't familiar with SOPS

```suggestion
func GetKMSKeyWithEncryptionCtx(tree *sops.Tree) (keyGroupIndex int, keyIndex int, key *kms.MasterKey) {
```

autrilla: This wouldn't handle files with more than one KMS key and encryption context, right? So the user would need to decrypt once for every key they have that would trigger the bug? If we were to support handling more than one key at a time, I'd wrap the current return values in a struct and return a slice of that struct.

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+	for i, kg := range tree.Metadata.KeyGroups {
+		for n, k := range kg {
+			kmsKey, ok := k.(*kms.MasterKey)
+			if ok {
+				if kmsKey.EncryptionContext != nil && len(kmsKey.EncryptionContext) >= 2 {
```
autrilla: It's unlikely that it'll happen, but this could have false positives. The bug is only triggered if there's at least two unique values for the encryption context

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+type GenericDecryptOpts struct {
+	Cipher      sops.Cipher
+	InputStore  sops.Store
+	InputPath   string
+	IgnoreMAC   bool
+	KeyServices []keyservice.KeyServiceClient
+}
+
+func LoadEncryptedFileWithBugFixes(opts GenericDecryptOpts) (*sops.Tree, error) {
+	tree, err := LoadEncryptedFile(opts.InputStore, opts.InputPath)
```
autrilla: This decrypts before fixing the bug, which will probably fail

cmd/sops/common/common.go
autrilla: I'd add a link to the original issue since it's very detailed. I'd also mention what specifically SOPS is going to do to fix the issue automatically, i.e. reencrypt the data key with the affected KMS keys.

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+}
+
+func FixAWSKMSEncryptionContextBug(opts GenericDecryptOpts, tree *sops.Tree) (*sops.Tree, error) {
+	// TODO - Finalize messaging here.
+	message := "Up until version 3.3.0 of sops there was a bug surrounding the " +
+		"use of encryption context with AWS KMS."
+	fmt.Println(wordwrap.WrapString(message, 75))
+
+	if !terminal.IsTerminal(int(os.Stdout.Fd())) {
+		return nil, NewExitError("Need to run this from a tty to resolve this issue.", codes.TTYRequired)
```
autrilla: If there's no TTY, maybe it'd be better to keep going without fixing the file, just to preserve past behavior? There might be people out there with half-broken CI systems, and doing this will fully break them

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+		IgnoreMac:   opts.IgnoreMAC,
+		Tree:        tree,
+		KeyServices: opts.KeyServices,
+	})
+	if err != nil {
+		kgndx, kndx, originalKey := GetKMSKeyWithEncryptionCtx(tree)
+
+		keyToEdit := *originalKey
+
+		encCtxVals := []string{}
```
autrilla: Nit: could make this a `map[string]struct{}` since we don't need to handle duplicate values

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+	success := false
+	// If there is another key, then we should be able to just decrypt
+	// without having to try different variations of the encryption context.
+	dataKey, err := DecryptTree(DecryptTreeOpts{
+		Cipher:      opts.Cipher,
+		IgnoreMac:   opts.IgnoreMAC,
+		Tree:        tree,
+		KeyServices: opts.KeyServices,
+	})
+	if err != nil {
```
autrilla: I'd extract this whole branch into a function to make it clearer what it's doing. Something along the lines of `RecoverDataKeyFromBuggyKMS`. A link to the Github issue in the code is probably helpful as well.

cmd/sops/common/common.go
```diff
@@ -136,3 +140,178 @@ func DefaultStoreForPath(path string) Store {
...
+				tree.Metadata.Version = version.Version
+				break
+			}
+		}
+	} else {
+		success = true
+	}
+
+	if !success {
+		return nil, NewExitError("Failed to decrypt, meaning there is likely another problem from the encryption context bug.", codes.ErrorDecryptingTree)
```
autrilla: It'd be nice to show the user what the error is here

cmd/sops/common/common.go
ajvb: I thought so as well, but it actually does work with files with more than one. Since decrypting the datakey once works, we then re-encrypt it for all keys, including any other KMS keys with encryption context.

cmd/sops/common/common.go
ajvb: Maybe I'm misunderstanding something, but doesn't that happen in [`common.DecryptTree`](https://github.com/mozilla/sops/blob/master/cmd/sops/common/common.go#L43) which gets called later?

cmd/sops/common/common.go
autrilla: Oh, right, we only need to recover the data key from one. That makes sense!

cmd/sops/common/common.go
autrilla: You're totally right! Sorry, misunderstood how this worked.

cmd/sops/common/common.go
ajvb: What is the recommended way to do this within `sops` cli? Just `fmt.Println` it?

autrilla: The simple solution would be to just wrap it, like:

```go
NewExitError(fmt.Sprintf("failed to decrypt, meaning there is likely another problem from the encryption context bug: %s", err), codes.ErrorDecryptingTree)
```

The more complicated solution would be implementing a [`UserError`](https://github.com/mozilla/sops/blob/master/usererrors.go#L15) for it. Or implementing some sort of `UserError` wrapping behavior to avoid having to make its own type for it.

ajvb: Oh gotcha, yeah I completely misunderstood what you meant. Fixing.

cmd/sops/common/common.go
```diff
@@ -136,3 +140,203 @@ func DefaultStoreForPath(path string) Store {
...
+		return nil, NewExitError(fmt.Sprintf("Could not marshal tree: %s", err), codes.ErrorDumpingTree)
+	}
+
+	file, err := os.Create(opts.InputPath)
+	if err != nil {
+		return nil, NewExitError(fmt.Sprintf("Could not open file for writing: %s", err), codes.CouldNotWriteOutputFile)
+	}
+	_, err = file.Write(encryptedFile)
+	if err != nil {
+		file.Close()
```
autrilla: uber-nit: could `defer file.Close()` after `os.Create(opts.InputPath)` instead of calling it on each branch

ajvb: I *think* we want to keep it like this, only because directly after this (within this function) we call `LoadEncryptedFile()`.

ajvb: But, I can remove the duplicate calls and call it directly after the `file.Write()`

ajvb: jk, using `defer` works as desired here.

PR #448, opened 2019-03-21
KMS Encryption Context bug fix, autofix, and additional testing

ajvb:
Includes:
* Bug fix
* Regression unit tests
* AWS KMS Functional tests (including a regression test)
* Autofix support within `rotate`, `edit`, `set`, and `decrypt` commands

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_deploy:
   - GOOS=linux CGO_ENABLED=0 go build -o dist/sops-${TRAVIS_TAG}.linux go.mozilla.org/sops/cmd/sops
   - |
       if [ ! -z "$TRAVIS_TAG" ]; then
-        version="$(grep '^const version' cmd/sops/version.go |cut -d '"' -f 2)"
+        version="$(grep '^const Version' version/version.go |cut -d '"' -f 2)"
         if [ "$version" != "$TRAVIS_TAG" ]; then
             echo "Git tag $TRAVIS_TAG does not match version $version, update the source!"
             exit 1
--- a/README.rst
+++ b/README.rst
@@ -1343,7 +1343,7 @@ The core team is composed of:
 * Julien Vehent @jvehent
 * AJ Banhken @ajvb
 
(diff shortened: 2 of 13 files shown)


jpsrn:
Awesome stuff, really great work! I've got one thing in my mind, though. I built a sops binary from this branch and tried it against one of my encryption context test files I had previously encrypted with `3.2.0` release, and noticed that it successfully prompted to fix the problem in the file. Then, I ran the same command again, but I was surprised to find that sops prompted again to fix the same problem it had already corrected. I inspected the encrypted the file contents and realized that it will prompt again, because the file version field is not being updated from `3.2.0` when the autofix code runs. 

Should sops bump the version field to `3.3.0` when it fixes the bug to prevent re-prompting, or should there be an alternative way to detect that the bug has already been fixed?

autrilla:
> Should sops bump the version field to `3.3.0` when it fixes the bug to prevent re-prompting

Yes! I missed that during review.

ajvb:
> Should sops bump the version field to `3.3.0` when it fixes the bug to prevent re-prompting, or should there be an alternative way to detect that the bug has already been fixed?

That's really strange, because when I test locally off of this branch the version is getting correctly updated to `3.3.0`.

ajvb:
> r+ but one thing I realize is we have to way to tell if the AWS tests are actually running

That's a good point. It would be easy to have the tests print something to stdout, but that would require going and looking at the test output to confirm.

I wonder if we had an additional check in `master` (and potentially in `develop`) for CI that did a simple assertion that the env var is set?

jpsrn:
> That's really strange, because when I test locally off of this branch the version is getting correctly updated to `3.3.0`.

@ajvb I checked it again and it looks like you're correct. This time I created a completely fresh test file with official sops `3.2.0` binary and tested the autofix again and it did set the version correctly. What may have happened is that I had previously created those sops test files with a build containing only my KMS fix, and of course the sops version in that build was set to `3.2.0`, causing this new sops `3.3.0` to attempt to autofix it. But since the file had been correctly encrypted due the to fix being present, sops autofix succesfully decrypted file, skipping the code path that would have actually applied the fix and bumped the file version field.

PR #450, opened 2019-03-27
Adding stable install instructions to README

EricTendian:
I noticed that there wasn't any clear install instructions for the stable builds, so I added some for each OS.

The one thing that this changes is that whenever a new release comes out, the README will need to be updated with the new version numbers, but I think that's a small price to pay for having more helpful docs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,6 @@ Stable release
 ~~~~~~~~~~~~~~
 Binaries and packages of the latest stable release are available at `https://github.com/mozilla/sops/releases <https://github.com/mozilla/sops/releases>`_.
 
-
 **Ubuntu/Debian**
 
 .. code:: bash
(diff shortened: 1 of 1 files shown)


jvehent:
Is this a pain point to anyone? I find it easier to redirect to our release page where the latest packages are published.
We also don't control third party repos like brew or arch, so I'm not sure we should encourage people to use them. They should use their best judgment.

EricTendian:
@jvehent We use sops at my company and quite a few engineers have asked me how to install it, so I think this would be good for those not as familiar with the command line. This is a dependency for https://github.com/futuresimple/helm-secrets, and while that library automatically installs sops, it does not keep it updated.

EricTendian:
Since it doesn't seem like there's any interest in having clear install docs, closing this PR.

ceroloy:
@EricTendian I am reopening this. I found it a bit strange that an open source tool lacks the installation doc for the stable branch.

ceroloy:
Similar issue #777

towens:
Maybe helpful if one downloaded the binary, darwin for example.
Assuming the `pwd` is on your `$PATH` 
```
mv ~/Downloads/sops-v1.2.3.darwin sops && chmod +x sops
```

PR #459, opened 2019-04-25
[WIP] migrate to go mod from govend for dependency management

tariq1890:
Changing times call for new dependency management tools.


autrilla:
I don't think we want to get rid of the vendored code in-repo, because there'll still be people that don't use go modules.

tariq1890:
We aren't getting rid of the vendored code. I think the diff is misleading. You can check my branch to confirm.

autrilla:
Before we merge, I'll have to remove all changes you've made to vendor/, since it's impossible to review the diff otherwise, and it'd be easy for attackers to sneak in malicious code. One this is all working, could you provide instructions on what you did to generate the changes to vendor/ so I can do that and commit the results myself?

tariq1890:
Totally understand the concern. In that case, I'd suggest you commit all of the changes. It wouldn't make sense to merge this PR. Emailing you.

Issue #465, opened 2019-05-15
`path_regex` is matching on relative path of working directory only


kentso:
I have such set up in `.sops.yaml` which is located in the project root:
```yaml
creation_rules:
    - path_regex: .*staging.*
      key_groups:
      - kms:
        - arn: 'arn:aws:kms:*:**:key/aa'
          role: arn:aws:iam::**:role/staging-role
          context:
            Environment: staging

    - path_regex: .*production.*
      key_groups:
      - kms:
        - arn: 'arn:aws:kms:*:**:key/aa'
          role: arn:aws:iam::**:role/production-role
          context:
            Environment: production

    - kms: 'arn:aws:kms:*:**:key/bb'
```

When I try to encrypt my secret

- current working director: `<project_root>/asia/production/a/b/c/`
- command: `sops secrets.enc.yaml`

The encrypted file did not use the expected rule for production, instead it used the fallback rule which is using kms key `arn:aws:kms:*:**:key/bb`.

To make this work, I need to specify the matching regex part in the path like this:

- current working director: `<project_root>/asia/production/a/b/c/`
- command: `sops ../../../../production/a/b/c/secrets.enc.yaml`

or

- current working director: `<project_root>/`
- command: `sops asia/production/a/b/c/secrets.enc.yaml`

i.e. `production` must be in the path passing to sops.

There is no extra explanation in README saying what path `path_regex` is matching to.
https://github.com/mozilla/sops#using-sops-yaml-conf-to-select-kms-pgp-for-new-files

I am expecting `path_regex` with be matching the path starting from the `.sops.yaml` it belongs to. Then my first case will work as expected.

PS. I have not tested but I think it is possible to create a `.sops.yaml` under every environment (e.g. `/asia/prodcution/.sops.yaml`, `/eu/staging/.sops.yaml`) to achieve what I want to do. However it will be tedious as the environment combination can grow large.

autrilla:
I think this is a duplicate of #242

autrilla:
I think a simple solution that would do what you want is expanding the path here:

https://github.com/mozilla/sops/blob/8f804bce1c96267fac61278e98f462f9413cda37/cmd/sops/main.go#L433

Such that for instance if you call `sops secrets.enc.yaml`, it would evaluate `secrets.enc.yaml` to the full path.

kentso:
Thanks for the answer.
I have created a pull request https://github.com/mozilla/sops/pull/466
Can you have a check?

[...]

PR #466, opened 2019-05-16
expand file path to full path before processing

kentso:
so that `path_regex` can match on the full path. It will fix issue https://github.com/mozilla/sops/issues/465

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -471,7 +471,7 @@ Let's take an example:
 * file named **something.dev.yaml** should use one set of KMS A
 * file named **something.prod.yaml** should use another set of KMS B
 * other files use a third set of KMS C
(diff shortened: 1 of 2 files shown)


kentso:
Are `tests::roundtrip_kms_encryption_context` and `tests::encrypt_json_file_kms` supposed to be ignored or not? When I execute function-test in my local they are ignored.

autrilla:
> 
> 
> Are `tests::roundtrip_kms_encryption_context` and `tests::encrypt_json_file_kms` supposed to be ignored or not? When I execute function-test in my local they are ignored.

They require AWS credentials, so we skip them everywhere except for on CI. It looks like CI for this is broken though, but the test failure is irrelevant to your patch.

PR #472, opened 2019-06-06
Vendoring update

ajvb:
It's been around 9 months since our last vendor update. This is also needed for some new features being worked on for sops workspace. (i.e. [`sops  publish`](https://github.com/mozilla/sops/compare/develop...sops-publish?expand=1))

Additionally, this PR regenerates the kms mocks.

Letting this sit in `develop` for a bit would also probably be a good idea.


ajvb:
@autrilla :D

PR #473, opened 2019-06-06
Implement `sops publish` command

ajvb:
### Summary

Publishes a file to a pre-configured destination (this lives in the sops config file). Additionally, support re-encryption rules that work just like the creation rules. Initial support for S3/GCS.

This is a part of the sops-workspace v2.0 project.

Includes the addition of a new dependency: `github.com/googleapis/gax-go/v2`

### Feature Description

sops (and sops workspace) is not designed for secrets provisioning. Other tools are better suited for this, such as deploying from S3 buckets, or relying on specialized services like Kubernetes secrets, AWS secrets manager, or Hashicorp Vault. The goal of `sops publish` is to allow sops to publish secrets out to these specialized services as easily and securely as possible.

Upon running `sops publish $file`, sops will publish the file to a per-configured destination. This destination must be configured within the `.sops.yaml` configuration file. The plan is to have this feature support pushing to any destination the user issuing the command has access to, such as S3, Hashicorp Vault, K8s secrets, etc. In this PR, only S3 and GCS are implemented.

An example configuration would look like:
```yaml
creation_rules:
  - pgp: 1022470DE3F0BC54BC6AB62DE05550BC07FB1A0A
destination_rules:
   - s3_bucket: "sops-secrets"
     s3_prefix: "secrets/"
     path_regex: s3/*
     recreation_rule:
       pgp: F69E4901EDBAD2D1753F8C67A64535C4163FB307
  - gcs_bucket: "sops-secrets"
    path_regex: gcs/*
    recreation_rule:
      pgp: F69E4901EDBAD2D1753F8C67A64535C4163FB307
```

example usage:
```
$ sops publish s3/app.yaml 
uploading /home/ajvb/sops/s3/foo.yaml to s3://sops-secrets/secrets/foo.yaml ? (y/n): y
```

The file that lands within `s3://sops-secrets/secrets/foo.yaml` will have been decrypted and re-encrypted with `pgp: F69E4901EDBAD2D1753F8C67A64535C4163FB307`. These recreation rules support everything that sops creation rules support (i.e. AWS KMS, GCP KMS, Azure KMS).


ajvb:
@autrilla somethings up with running the new `publish` functional test on travis ci and will work on fixing tomorrow.

ajvb:
@autrilla functional tests passing! :dancer: should be good to go whenever you have time to take a look. Am also going to send this to a few other moz folks for feedback.

jvehent:
> This destination must be configured within the .sops.yaml configuration file.

Are you planning to add a command line flag to specify an upload destination without requiring a config file?

cmd/sops/main.go
```diff
@@ -104,6 +105,48 @@ func main() {
...
+			Name:      "publish",
+			Usage:     "Publish sops file to a configured destination",
+			ArgsUsage: `file`,
+			Flags: append([]cli.Flag{
+				cli.BoolFlag{
+					Name:  "yes, y",
+					Usage: `pre-approve all changes and run non-interactively`,
+				},
+				cli.BoolFlag{
+					Name:  "verbose",
```
jvehent: isn't this already covered by top-level flags?

cmd/sops/subcommand/publish/publish.go
```diff
@@ -0,0 +1,114 @@
...
+		return fmt.Errorf("can't operate on a directory")
+	}
+	_, fileName := filepath.Split(path)
+
+	conf, err := config.LoadForFile(opts.ConfigPath, opts.InputPath, make(map[string]*string))
+	if err != nil {
+		return err
+	}
+	if conf.Destination == nil {
+		return errors.New("No destination configured for this file.")
```
jvehent: really a styling nit, but I find that error messages render better when they are kept lowercase and without trailing periods.

cmd/sops/subcommand/publish/publish.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// Re-encrypt if settings exist to do so
+	if len(conf.ReEncryptionKeyGroups) != 0 {
+		key, err := tree.Metadata.GetDataKeyWithKeyServices(opts.KeyServices)
+		if err != nil {
+			return common.NewExitError(err, codes.CouldNotRetrieveKey)
+		}
+		tree.Metadata.KeyGroups = conf.ReEncryptionKeyGroups
+		tree.Metadata.ShamirThreshold = min(tree.Metadata.ShamirThreshold, len(tree.Metadata.KeyGroups))
```
jvehent: I don't think this should silently pick the lowest value between number of keygroups and shamir threshold. If a shamir threshold is configured with a value > 1, and there are not enough keygroups to satisfy the requirement, then we should exit with an error.

cmd/sops/subcommand/publish/publish.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Check that this is a sops-encrypted file
+	tree, err := common.LoadEncryptedFile(opts.InputStore, opts.InputPath)
+	if err != nil {
+		return err
+	}
+
+	// Re-encrypt if settings exist to do so
+	if len(conf.ReEncryptionKeyGroups) != 0 {
+		key, err := tree.Metadata.GetDataKeyWithKeyServices(opts.KeyServices)
```
jvehent: should this be interactive to make sure the user accepts to remove the existing keygroups and reencrypt?

cmd/sops/subcommand/publish/publish.go
```diff
@@ -0,0 +1,114 @@
...
+	if opts.Interactive {
+		var response string
+		for response != "y" && response != "n" {
+			fmt.Printf("\nuploading %s to %s ? (y/n): ", path, conf.Destination.Path(fileName))
+			_, err := fmt.Scanln(&response)
+			if err != nil {
+				return err
+			}
+		}
+		if response == "n" {
```
jvehent: there is an infinite number of inputs people use to signify "not yes", so perhaps using a switch case with a default that cancels publishing is safer.

cmd/sops/subcommand/publish/publish.go
```diff
@@ -0,0 +1,114 @@
...
+
+	err = conf.Destination.Upload(fileContents, fileName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func min(a, b int) int {
```
jvehent: https://golang.org/pkg/math/#Min

config/config.go
```diff
@@ -87,8 +89,16 @@ type azureKVKey struct {
 	Version  string `yaml:"version"`
 }
 
+type destinationRule struct {
+	PathRegex        string       `yaml:"path_regex"`
+	S3Bucket         string       `yaml:"s3_bucket"`
+	S3Prefix         string       `yaml:"s3_prefix"`
+	GCSBucket        string       `yaml:"gcs_bucket"`
+	GCSPrefix        string       `yaml:"gcs_prefix"`
```
jvehent: would it be more elegant to use a single variable to store the target, then switch on the protocol method? It would allow us to add new publication protocols without changing this struct.

ajvb: So something like?:
```
destination_rules:
  - location: "s3://my-bucket/secrets"
```

That could work, but could become weird if certain publish destinations don't map well to that. For example, could we specify all of the needed parameters for a Vault destination in a single variable? I don't think it's a problem to have to make additions to the config format to support a new destination?

cmd/sops/main.go
ajvb: Yes-ish, with a slight change we can support `sops --verbose publish file`, but we lose support for `sops publish --verbose`. Making a change to support both.

ajvb:
>Are you planning to add a command line flag to specify an upload destination without requiring a config file?

Yeah,

cmd/sops/subcommand/publish/publish.go
ajvb: It might be worth adding the confirmation step prior to this and include it within the confirmation message (similar to `sops updatekeys`). ie:
```
$ sops publish s3/app.yaml
uploading /home/ajvb/sops-testing/publish/s3/app.yaml to s3://sops-secrets/testing/app.yaml

and the following changes will be made to the file's groups:
+++ F69E4901EDBAD2D1753F8C67A64535C4163FB307
--- 1022470DE3F0BC54BC6AB62DE05550BC07FB1A0A

Is this okay? (y/n): 
```

cmd/sops/subcommand/publish/publish.go
autrilla: I think it might be best to let users define what Shamir threshold they want in their reencryption rules. Essentially, I'd make the reencryption rules have the same format as normal creation rules, which already let you specify a Shamir threshold.

README.rst
```diff
@@ -768,6 +768,33 @@ By default ``sops`` just dumps all the output to the standard output. We can use
...
+
+.. code:: yaml
+
+   creation_rules:
+      - pgp: 1022470DE3F0BC54BC6AB62DE05550BC07FB1A0A
+   destination_rules:
+      - s3_bucket: "sops-secrets"
+        path_regex: s3/*
+        reencryption_rule:
+           pgp: F69E4901EDBAD2D1753F8C67A64535C4163FB307
```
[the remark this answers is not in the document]
ajvb: I wonder if it would be better to name it `creation_rule` while still having it nested like it is? This would also require handling all of the features that creation rules currently support. Potentially, if this is done along side something like https://github.com/mozilla/sops/pull/473#discussion_r291751560, it could assist users in better understanding what changes will be made when they use `sops publish`.

But I'm not sure, need to think about this one a little bit...

cmd/sops/subcommand/publish/publish.go
autrilla: For operations like this one that could potentially break things, I like terraform's approach of only accepting `yes` and nothing else as a prompt to continue, and everything else cancels the operation. Although that'd make us inconsistent with updatekeys...

config/config.go
jvehent: if it's a string, it could also be `vault_user=foo vault_pass=bar vault_url=http://example.net`

README.rst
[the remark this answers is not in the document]
jvehent: if you're going to reuse the `creation_rule`, perhaps calling it `recreation_rule` would be clear enough?

publish/gcs.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (gcsd *GCSDestination) Upload(fileContents []byte, fileName string) error {
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+	wc := client.Bucket(gcsd.gcsBucket).Object(gcsd.gcsPrefix + fileName).NewWriter(ctx)
+	_, err = wc.Write(fileContents)
```
adrianosela:
```
defer wc.Close()
```

cmd/sops/subcommand/publish/publish.go
ajvb: I'm happy to come back through and align `updatekeys` and the `keyservice` prompt if we determine that only `yes` works. I am a fan of that, or making `no` the default if you just press enter.

ajvb:
@jvehent @autrilla with https://github.com/mozilla/sops/pull/473/commits/3b6220e2b82cea4ac4e3a16c40a55cba85109194 - switched `reencryption_rule` to `recreation_rule` and added full support for all `creation_rule` features (like actually specifying shamir). Also changed how the recreation works, so now it's more akin to doing `sops -d -i $file; sops -e -i $file`

cmd/sops/subcommand/publish/publish.go
autrilla: I'm not 100% that breaking current behavior for this change is worth it. I doubt there's anyone doing `yes | sops ...` but who knows...

config/config.go
```diff
@@ -192,11 +165,87 @@ func loadForFileFromBytes(confBytes []byte, filePath string, kmsEncryptionContex
...
+	conf := configFile{}
+	err := conf.load(confBytes)
+	if err != nil {
+		return nil, fmt.Errorf("error loading config: %s", err)
+	}
+
+	var rule *creationRule
+	var dRule *destinationRule
+
+	if getFromDestinationRule {
```
autrilla: I would make this two separate functions instead of one with this branching. You can extract common code into helper functions if needed

config/config.go
ajvb: True, but then we will have to support more and more complex parsing logical and will be forced to pack more and more options into a single line rather than various kv pairs.

imo before we ship this feature, it would be worth getting more destinations supported (k8s secrets, vault, etc) and having some users try it out.

ajvb:
@autrilla Mind taking a quick look at the last commit? Added support for printing diffs of the key group changes by re-using the updatekeys code. If that looks good then this should be ready to merge.

autrilla:
That last commit looks good to me!

Issue #483, opened 2019-06-19
encrypting empty file panics


jasonkuhrt:
Not a great onboarding experience; in the following I try helm-secrets (which uses sops under the hood) but get a panic.

```
touch foo.yaml
❯ helm secrets enc foo.yaml
Encrypting foo.yaml
panic: runtime error: index out of range

goroutine 1 [running]:
main.encrypt(0x1a88ce0, 0xc00009a630, 0x2615200, 0x2089cf8, 0x2615200, 0x2089cf8, 0x7ffeefbfe0ef, 0x14, 0xc000064990, 0x1, ...)
	/private/tmp/sops-20190418-79323-1cw3pqc/sops-3.3.0/src/go.mozilla.org/sops/cmd/sops/encrypt.go:66 +0x98a
main.main.func5(0xc000226000, 0x0, 0x0)
	/private/tmp/sops-20190418-79323-1cw3pqc/sops-3.3.0/src/go.mozilla.org/sops/cmd/sops/main.go:483 +0x2ed8
go.mozilla.org/sops/vendor/gopkg.in/urfave/cli%2ev1.HandleAction(0x17f8060, 0x1947088, 0xc000226000, 0x0, 0x0)
	/private/tmp/sops-20190418-79323-1cw3pqc/sops-3.3.0/src/go.mozilla.org/sops/vendor/gopkg.in/urfave/cli.v1/app.go:490 +0xc8
go.mozilla.org/sops/vendor/gopkg.in/urfave/cli%2ev1.(*App).Run(0xc000214000, 0xc0000b8000, 0x8, 0x8, 0x0, 0x0)
	/private/tmp/sops-20190418-79323-1cw3pqc/sops-3.3.0/src/go.mozilla.org/sops/vendor/gopkg.in/urfave/cli.v1/app.go:264 +0x590
main.main()
	/private/tmp/sops-20190418-79323-1cw3pqc/sops-3.3.0/src/go.mozilla.org/sops/cmd/sops/main.go:649 +0x2578
Error: plugin "secrets" exited with error
```

Maybe a more useful message would be `Nothing to encrypt` or just make encryption work but it be, well, empty string.

ajvb:
Thanks for reporting @jasonkuhrt! Seems to be yaml specific.

fzyzcjy:
Hi the issue seems to still exist...

Moskovych:
@ajvb , Is this issue planned to be fixed or somehow omitted?
[...]

PR #485, opened 2019-07-01
encrypt.go: Fix panic on reading empty file

mr-karan:
Fixes a panic crash when a file is loaded without any contents.
Adds a new error `FileContentsEmpty` in `codes.go`.

Fixes https://github.com/mozilla/sops/issues/483

### DIFF
--- a/cmd/sops/codes/codes.go
+++ b/cmd/sops/codes/codes.go
@@ -28,4 +28,5 @@ const (
 	NoEditorFound                          int = 201
 	FailedToCompareVersions                int = 202
 	FileAlreadyEncrypted                   int = 203
+	FileContentsEmpty                      int = 204
 )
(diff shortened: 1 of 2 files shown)


autrilla:
Thanks for the patch! I don't think this will catch the case where a user runs `sops file.yaml`, and then empties the file

ajvb:
Thanks @mr-karan! imo we should also make sure to add a regression test with the patch.

mr-karan:
@autrilla Um I am not sure if I got you. When the user runs `sops -e file.yaml` the file contents are loaded in memory, so it won't really matter what the user does to the file *after* running this command.
Did you mean something else?

@ajvb On it :+1:

autrilla:
> @autrilla Um I am not sure if I got you. When the user runs `sops -e file.yaml` the file contents are loaded in memory, so it won't really matter what the user does to the file _after_ running this command.
> Did you mean something else?

It doesn't panic, but creating a new file with `sops new_file.yaml`, emptying the contents, and saving it, creates an empty file with no sops section. It should probably just error, since it doesn't make sense. And it's the same if you edit an existing file with `sops existing_file.yaml` and empty the contents. I think we should emit the same error in those cases.

jvehent:
@mr-karan Are you still making changes to this PR or should we close it?

PR #490, opened 2019-07-08
Fix linting and formatting issues

adrianosela:
Trying to get featured in https://github.com/avelino/awesome-go

The repo's maintainer has asked we take a look at our [go report card topics](https://goreportcard.com/report/github.com/mozilla-services/sops)

I've remediated all entries which do not require a significant change of code, it is mostly just the addition of comments to exported functions. 

### Note3: A report card topic which I have not addressed are cyclomatic complexities, which I too have left out on purpose for a subsequent PR.

PR #491, opened 2019-07-08
Address Go Lint messages

adrianosela:
Follow up PR for https://github.com/mozilla/sops/pull/490

and once again:

### Note: I am not extremely well acquainted with the code base -- I ask that you carefully read all comments, and if something does not look perfect write a comment on the line and I will update the PR shortly afterwards

This is also a step forward in getting sops in avelino/awesome-go
https://github.com/avelino/awesome-go/pull/2603

### DIFF
--- a/audit/audit.go
+++ b/audit/audit.go
@@ -9,7 +9,9 @@ import (
 
 	"github.com/pkg/errors"
 
+	// empty import as per https://godoc.org/github.com/lib/pq
 	_ "github.com/lib/pq"
+
 	"github.com/sirupsen/logrus"
 	"go.mozilla.org/sops/logging"
 	"gopkg.in/yaml.v2"
@@ -67,36 +69,46 @@ type config struct {
 
 var auditors []Auditor
 
+// SubmitEvent handles an event for all auditors
 func SubmitEvent(event interface{}) {
 	for _, auditor := range auditors {
 		auditor.Handle(event)
 	}
 }
 
+// Register registers a new Auditor in the global auditor list
 func Register(auditor Auditor) {
 	auditors = append(auditors, auditor)
 }
 
+// Auditor is notified when noteworthy events happen, for example when a file is encrypted or decrypted. 
 type Auditor interface {
 	Handle(event interface{})
 }
 
+// DecryptEvent contains fields relevant to a decryption event
 type DecryptEvent struct {
 	File string
 }
 
+// EncryptEvent contains fields relevant to an encryption event
 type EncryptEvent struct {
 	File string
 }
 
+// RotateEvent contains fields relevant to a key rotation event
 type RotateEvent struct {
 	File string
 }
 
+// PostgresAuditor is a Postgres SQL DB implementation of the
+// Auditor interface
 type PostgresAuditor struct {
 	DB *sql.DB
 }
 
+// NewPostgresAuditor is the constructor for a new PostgresAuditor object
+// initialized with the given db connection string
 func NewPostgresAuditor(connStr string) (*PostgresAuditor, error) {
 	db, err := sql.Open("postgres", connStr)
 	pg := &PostgresAuditor{DB: db}
@@ -113,6 +125,8 @@ func NewPostgresAuditor(connStr string) (*PostgresAuditor, error) {
 	return pg, nil
 }
 
+// Handle is the PostgresAuditor implementation of the function required by the
+// Auditor interface
 func (p *PostgresAuditor) Handle(event interface{}) {
 	u, err := user.Current()
 	if err != nil {
--- a/audit/audit.go
+++ b/audit/audit.go
@@ -81,8 +81,12 @@ func Register(auditor Auditor) {
 	auditors = append(auditors, auditor)
 }
 
-// Auditor is notified when noteworthy events happen, for example when a file is encrypted or decrypted.
+// Auditor is notified when noteworthy events happen,
+// for example when a file is encrypted or decrypted.
 type Auditor interface {
+	// Handle() takes an audit event and attempts to persists it;
+	// how it is persisted and how errors are handled is up to the
+	// implementation of this interface.
 	Handle(event interface{})
 }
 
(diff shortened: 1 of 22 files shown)


cmd/sops/common/common.go
```diff
@@ -161,6 +176,7 @@ func DetectKMSEncryptionContextBug(tree *sops.Tree) (bool, error) {
 	return false, nil
 }
 
+// GetKMSKeyWithEncryptionCtx returns a KMS key corresponding to a tree
```
autrilla: Maybe something like

```suggestion
// GetKMSKeyWithEncryptionCtx returns the first KMS key affected by the encryption context bug as well as its location in the key groups.
```

publish/publish.go
```diff
@@ -1,5 +1,7 @@
 package publish
 
+// Destination represnts actions which all destination types
```
autrilla:
```suggestion
// Destination represents actions which all destination types
```

sops.go
```diff
@@ -100,7 +100,8 @@ type TreeItem struct {
 // TreeBranch is a branch inside sops's tree. It is a slice of TreeItems and is therefore ordered
 type TreeBranch []TreeItem
 
-// Trees usually have more than one branch
+// TreeBranches is a collection of TreeBranch
```
autrilla: This looks like it might all be one sentence, let's do:

```
TreeBranches is a collection of TreeBranch.
Trees usually have more than one branch.
```

To clarify

sops.go
```diff
@@ -150,31 +149,28 @@ func set(branch interface{}, path []interface{}, value interface{}) interface{}
...
-			return branch
+			branch[position] = set(branch[position], path[1:], value)
 		}
+		return branch
 	default:
 		return valueFromPathAndLeaf(path, value)
 	}
 }
 
+// Set sets a value on a given tree
```
autrilla:
```suggestion
// Set sets a value on a given tree for the specified path
```

audit/audit.go
```diff
@@ -67,36 +69,46 @@ type config struct {
...
 	File string
 }
 
+// PostgresAuditor is a Postgres SQL DB implementation of the
+// Auditor interface
 type PostgresAuditor struct {
 	DB *sql.DB
 }
 
+// NewPostgresAuditor is the constructor for a new PostgresAuditor object
```
ajvb: Should call it a `struct` rather than an `object`.

audit/audit.go
```diff
@@ -113,6 +125,8 @@ func NewPostgresAuditor(connStr string) (*PostgresAuditor, error) {
 	return pg, nil
 }
 
+// Handle is the PostgresAuditor implementation of the function required by the
```
ajvb: Let's describe this function (at a high-level) rather than just describing it as "required by the interface"

publish/gcs.go
```diff
@@ -7,19 +7,23 @@ import (
...
 func NewGCSDestination(gcsBucket string, gcsPrefix string) *GCSDestination {
 	return &GCSDestination{gcsBucket, gcsPrefix}
 }
 
+// Path returns a the GCS path for a file within this GCS Destination
 func (gcsd *GCSDestination) Path(fileName string) string {
 	return fmt.Sprintf("gcs://%s/%s%s", gcsd.gcsBucket, gcsd.gcsPrefix, fileName)
 }
 
+// Upload uploads contents to a new file in GCS
```
ajvb: Let's remove the "new" here, as it may be overwriting an existing file.

publish/s3.go
```diff
@@ -9,19 +9,23 @@ import (
...
 func NewS3Destination(s3Bucket string, s3Prefix string) *S3Destination {
 	return &S3Destination{s3Bucket, s3Prefix}
 }
 
+// Path returns the S3 path of a file in an S3 Destination (bucket)
 func (s3d *S3Destination) Path(fileName string) string {
 	return fmt.Sprintf("s3://%s/%s%s", s3d.s3Bucket, s3d.s3Prefix, fileName)
 }
 
+// Upload uploads contents to a new file in an S3 Destination (bucket)
```
ajvb: same here as with GCS's `Upload()` func

stores/yaml/store.go
```diff
@@ -101,6 +101,8 @@ func (store Store) treeBranchToYamlMap(in sops.TreeBranch) yaml.MapSlice {
 	return branch
 }
 
+// LoadEncryptedFile loads the contexts of an encrypted yaml file onto a
```
ajvb: Did you mean to write "contents" rather than "context"? Also present in the comment below.

audit/audit.go
autrilla: I think it should be described at a high level on the interface and not on this particular implementation. We might want to document some specifics to postgres here, though.

cmd/sops/common/common.go
```diff
@@ -24,10 +24,13 @@ import (
 	"gopkg.in/urfave/cli.v1"
 )
 
+// ExampleFileEmitter represents actions needed for a struct to be comliant
```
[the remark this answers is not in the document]
adrianosela: ✅

sops.go
adrianosela: ✅

audit/audit.go
adrianosela: ✅

audit/audit.go
adrianosela: +1 @autrilla

adrianosela: All PR comments have been addressed except for this one. I will let you two decide what to write :)

stores/yaml/store.go
```diff
@@ -136,6 +138,8 @@ func (store *Store) LoadEncryptedFile(in []byte) (sops.Tree, error) {
 	}, nil
 }
 
+// LoadPlainFile loads the contexts of a plaintext yaml file onto a
```
ajvb: s/contexts/contents/?

adrianosela: #copypastafail

audit/audit.go
ajvb: @autrilla agreed

@adrianosela 
on the interface, append something like: "Handle() takes an audit event and attempts to persists it. How it is persisted and how errors are handled is up to the implementation of this interface."

on `PostgresAuditor.Handle()`, something like: "Persists the audit event by writing a row to the 'audit_event' table. Errors with writing to the database will output a log message and the process will exit with status set to 1."

PR #497, opened 2019-07-18
Fix errors during key rotation being ignored

### LINKED ISSUE #484: --rotate --in-place results in empty file if user cannot decrypt
Related to #365 , the following scenario results in an empty file (unexpected) instead of a failure (expected) or no action:

* create .sops.yaml:
```
creation_rules:
  - path_regex: test.yaml
    key_groups:
      - pgp:
        # userA
        - FD43E10B0D145AA3E865E3088554CA96EF7FF338
        # userB
        - 75385B17ADC57E09BA14A523A7208B9C8B55CCB8
```
* create test.yaml
```
$ sops test.yaml
```
* modify .sops.yaml, removing userB (yourself)
```
creation_rules:
  - path_regex: test.yaml
    key_groups:
      - pgp:
        # userA
        - FD43E10B0D145AA3E865E3088554CA96EF7FF338
```
* update keys
```
$ sops updatekeys -y test.yaml
...
2019/06/26 09:57:28 File test.yaml synced with new keys
```
* rotate data key to remove access for removed user
```
$ sops -r -i test.yaml ; echo $?
0
# I would expect 'Failed to get the data key required to decrypt the SOPS file.' and no change to test.yaml
$ cat test.yaml
$
```

The purpose of rotating the data key after updating the user keys is to prevent removed users from decrypting the files.

Am I approaching this the wrong way?

autrilla:
#263 rises from the dead :)

Fixes #484

@ajvb do you think a functional for this is in order?

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -569,7 +569,8 @@ func main() {
 			for _, k := range gcpkms.MasterKeysFromResourceIDString(c.String("add-gcp-kms")) {
 				addMasterKeys = append(addMasterKeys, k)
 			}
-			azureKeys, err := azkv.MasterKeysFromURLs(c.String("add-azure-kv"))
+			var azureKeys []*azkv.MasterKey
+			azureKeys, err = azkv.MasterKeysFromURLs(c.String("add-azure-kv"))
 			if err != nil {
 				return err
 			}


cmd/sops/main.go
```diff
@@ -569,7 +569,8 @@ func main() {
 			for _, k := range gcpkms.MasterKeysFromResourceIDString(c.String("add-gcp-kms")) {
 				addMasterKeys = append(addMasterKeys, k)
 			}
-			azureKeys, err := azkv.MasterKeysFromURLs(c.String("add-azure-kv"))
```
autrilla: This shadowed the outer `err`, which is handled later.

autrilla:
This whole ordeal is pretty bad, since it's already happened twice. This would be fine a stopgap, but we should find a way to prevent this from happening in the first place.

ajvb:
>@ajvb do you think a functional for this is in order?

Yeah, I think so.

>This whole ordeal is pretty bad, since it's already happened twice. This would be fine a stopgap, but we should find a way to prevent this from happening in the first place.

Agreed. I'm not sure I understand exactly what is going on with all of these issues in mind. Is the core of the problem how "gopkg.in/urfave/cli.v1" `app.Action` handles returned errors? Should we use variations on the `err` variable in `app.Action` combined with wrapping everything in `toExitError()`, or is that not enough?

PR #498, opened 2019-07-29
feat(sops): moved dependency management from govend to go modules

gitirabassi:
This PR is aimed at move sops from using `govend` as dependency management to `go mod`


autrilla:
It looks like this is removing vendoring entirely -- any reason why we should do that? I'm kind of against it until the Go ecosystem provides some sort of immutable package repository. Right now we're pinning stuff based on git commit hashes, and those could disappear.

gitirabassi:
For now it doesn’t changes anything compared to before. We’ll still have the `vendor` folder with everything inside, just like with govend:  `go mod vendor` does this exactly

There are already a coupled of modules proxies from the google team and the Microsoft team as well preventing commits to “just disappear”.

gitirabassi:
@autrilla I realized I didn't added the pieces to make the `vendor` folder to be used during pipeline run

gitirabassi:
I'm about to open a PR to add Hashicorp Vault transit backend and this change would make things easier

autrilla:
Can you provide instructions on how you made the changes to the `vendor` directory? Did you just run `make vendor`? We prefer long-time SOPS contributors or Mozilla employees (or maybe even just the latter, cc @ajvb) make changes to vendoring directories as the code is hard to review and it's easy to introduce malicious code that way.

Basically, I want to revert your changes such that the diff is reviewable, and then I can run any command that will make the large code changes required.

gitirabassi:
@autrilla yes, just `make vendor`. You can copy `go.mod` and `go.sum` to fresh `develop` branch. then run `GO111MODULE=on go mod vendor` and you would get the same vendoring folder I got here

autrilla:
Merging since it looks like Go modules is where the community is going overall.

PR #499, opened 2019-07-30
Remove vendor directory from repo

ajvb:
With #498 being merged and making the switch to Go modules, there does not seem to be a need to store this directory within the repo anymore. As well, it will make reviewing dependency updates and viewing the PR's within Github's UI much more manageable.

This PR also includes:
* adding `vendor/` to `.gitignore`
* making use of https://proxy.golang.org/ within the make commands
* switching the Dockerfile to use `make install`
* changing the Makefile so that `make test` will run `make vendor` first.


autrilla:
This is largely okay, but we have the risk of someone removing a commit we’re pinned to or deleting the repository entirely and breaking the build. Not sure what we’d do in that case.

ajvb:
@autrilla As of Go 1.13, this is actually solved via https://proxy.golang.org/

ajvb:
@autrilla Updated the PR description to include some additional changes I made to the Makefile and Dockerfile

Issue #508, opened 2019-08-13
[Question] Several encryption suffixes


naadev:
Greetings...

I'm really glad to have discovered SOPS. It's really good stuff.

I'm working on a proposal to use SOPS as part of our standard security policies. In our case, I've found there could be, just to make it more transparent to final users that there are two suffix I'd like to see encrypted, in example all entries ending with:

- Password (i.e. ServicePassword)
- Key (i.e. ApiKey)

I'm wondering if there is any way to provide two suffixes when encrypting, I've tried with:

-  sops -e --output file.sops --encrypted-suffix Password --encrypted-suffix Key file.json
-  sops -e --output file.sops --encrypted-suffix Password,Key file.json
 
Is this a supported scenario somehow?

autrilla:
That's not supported right now, but I guess we can turn this issue into a feature request for multiple encrypted/unencrypted suffixes :)

jvehent:
Note for later, in case we decide to do this: don't use a regex or
interpreted string, since it could mess with the way people use it today,
but instead allow the flag to be set multiple times and turn it into an
[]string.

bobmhong:
I just discovered the same limitation today and would like to see the ability to pass a list of encrypted suffixes.

naadev:
I will try to spare some time today to provide a pull request to work with list... :)

Sent from my iPhone

On 20 Aug 2019, at 01:41, Bob Hong <notifications@github.com<mailto:notifications@github.com>> wrote:


I just discovered the same limitation today and would like to see the ability to pass a list of encrypted suffixes.

autrilla:
I'm not sure multiple suffixes are worth implementing with #509 in the pipeline, as you can achieve the same behavior with both

naadev:
That's what I was looking for... thank you @autrilla

PR #509, opened 2019-08-14
added encrypted-regex option

jpriebe:
Our team would like to use sops to protect kubernetes secrets in YAML files.

Here's what we want to do:

- encrypt all secret literals, regardless of whether they are under the "data" or "stringData" keys
- *not* encrypt any other values in the file, so we can still easily visually scan the encrypted files (so for instance, we could see the unencrypted secret names in a git diff)
- *not* require developers to put any kind of suffix onto key names
- *not* require any additional pre- or post-processors; we want to use `sops decrypt` and pipe straight to `kubectl apply`.

Here's what we've tried:

- without using encrypted suffix, all values will be encrypted, which makes it impossible to visually scan an encrypted YAML file for fields like metadata.name
- you can add an encrypted suffix like "-encrypted" to the "data" and "stringData" keys, but then the decoded YAML will not be legal, and you'll need a post-processor to apply the files to a k8s cluster (and you'd have to rely on your developers always adding the "-encrypted", or they risk putting unencrypted secrets into the git repo)
- you could use an encrypted suffix like "ata" to automatically encrypting up the "data" and "stringData" keys, but that has the side effect of encrypting the "metadata" keys, which means you can't visually inspect the metadata.name while the file is encrypted
- you could use an encrypted suffix like "stringData" and try to enforce a policy that your developers always put their secrets into "stringData" not "data", but all it takes is one mistake to commit a secret to the git repo

None of these were satisfactory.  So I added a new option: `encrypted-regex`, which lets you encrypt only those values whose keys match the given regular expression.  With this option, you can do this:

```
sops --encrypt --encrypted-regex '^(data|stringData)$'
```

This will encrypt our data and stringData values, but nothing else.

I know there was an objection to PR 385 (https://github.com/mozilla/sops/pull/385) last year on the grounds that the behavior should be moved into the library.  I am hoping that you won't view this PR the same way.  I'm not sure I have the resources for a deep dive into the sops codebase.  But this small change I'm making brings (IMHO) tremendous value to the product wrt kubernetes secrets files.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1193,10 +1193,21 @@ The unencrypted suffix can be set to a different value using the
 Conversely, you can opt in to only encrypt some values in a YAML or JSON file,
 by adding a chosen suffix to those keys and passing it to the ``--encrypted-suffix`` option.
 
+A third method is to use the ``--encrypted-regex`` which will only encrypt values under
+keys that match the supplied regular expression.  For example, this command:
+
+.. code:: bash
+
+	$ sops --encrypt --encrypted-regex '&(data|stringData)' k8s-secrets.yaml
+
+will encrypt the values under the ``data`` and ``stringData`` keys in a YAML file
+containing kubernetes secrets.  It will not encrypt other values that help you to
+navigate the file, like ``metadata`` which contains the secrets' names.
+
 You can also specify these options in the ``.sops.yaml`` config file.
 
-Note: these two options ``--unencrypted-suffix`` and ``--encrypted-suffix`` are mutually exclusive and
-cannot both be used in the same file.
+Note: these three options ``--unencrypted-suffix``, ``--encrypted-suffix``, and ``--encrypted-regex`` are 
+mutually exclusive and cannot both be used in the same file.
 
 Encryption Protocol
 -------------------
--- a/README.rst
+++ b/README.rst
@@ -1207,7 +1207,7 @@ navigate the file, like ``metadata`` which contains the secrets' names.
 You can also specify these options in the ``.sops.yaml`` config file.
 
 Note: these three options ``--unencrypted-suffix``, ``--encrypted-suffix``, and ``--encrypted-regex`` are 
-mutually exclusive and cannot both be used in the same file.
+mutually exclusive and cannot all be used in the same file.
 
 Encryption Protocol
 -------------------
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -440,6 +440,10 @@ func main() {
 			Usage: "override the encrypted key suffix. When empty, all keys will be encrypted, unless otherwise marked with unencrypted-suffix.",
 		},
 		cli.StringFlag{
+			Name:  "encrypted-regex",
+			Usage: "set the encrypted key suffix. When specified, only keys matching the regex will be encrypted.",
+		},
+		cli.StringFlag{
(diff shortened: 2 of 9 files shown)


cmd/sops/main.go
```diff
@@ -503,13 +508,29 @@ func main() {
...
+			cryptRuleCount++
+		}
+
+		if cryptRuleCount > 1 {
+			return common.NewExitError("Error: cannot use more than one of encrypted_suffix, unencrypted_suffix, or encrypted_regex in the same file", codes.ErrorConflictingParameters)
+		}
+
+		// only supply the default UnencryptedSuffix when EncryptedSuffix and EncryptedRegex are not provided
+		if cryptRuleCount == 0 {
+			return common.NewExitError("Error: cannot use both encrypted_suffix and unencrypted_suffix in the same file", codes.ErrorConflictingParameters)
```
autrilla: This should set unencryptedSuffix to sops.DefaultUnencryptedSuffix, no?

autrilla:
Thanks for the detailed comment explaining your use case and what you've tried. My only objection to this is that it complicates the CLI and code a bit. 

Julien also had [this concern](https://github.com/mozilla/sops/issues/508#issuecomment-520834451) but I don't think it applies here as it's entirely a new flag.

cmd/sops/main.go
jpriebe: Yep.  This was a bad copy-and-paste job.  I will correct it.

jpriebe: That was a good catch.  I've pushed the fix.

jpriebe:
Thanks for your consideration.  I can see the case for multiple instances of encrypted-suffix (or even something more explicit like encrypted-key).

I implemented encrypted-regex because it was the shortest path to getting the functionality I need.  Supporting multiple flag instances and []string would require much more significant refactoring of the existing code.

jpriebe:
Caught an issue with config_test.go.  I didn't realize that "go test" didn't run all the tests -- had to run ./test.sh to get all tests running, which exposed my mistake.

README.rst
```diff
@@ -1193,10 +1193,21 @@ The unencrypted suffix can be set to a different value using the
...
+will encrypt the values under the ``data`` and ``stringData`` keys in a YAML file
+containing kubernetes secrets.  It will not encrypt other values that help you to
+navigate the file, like ``metadata`` which contains the secrets' names.
+
 You can also specify these options in the ``.sops.yaml`` config file.
 
-Note: these two options ``--unencrypted-suffix`` and ``--encrypted-suffix`` are mutually exclusive and
-cannot both be used in the same file.
+Note: these three options ``--unencrypted-suffix``, ``--encrypted-suffix``, and ``--encrypted-regex`` are 
+mutually exclusive and cannot both be used in the same file.
```
autrilla:
```suggestion
mutually exclusive and cannot all be used in the same file.
```

autrilla:
@ajvb: based on the number of people who've been active on related issues, I'd say this feature is useful enough to be added to sops. Let me know if you have any concerns, otherwise I'll merge it.

autrilla:
Looks like merging this broke some tests

autrilla:
> Looks like merging this broke some tests

It actually didn't, that seems unrelated because when reverting this, they still fail.

PR #530, opened 2019-09-13
fix --encrypted-regex documentation

oskarkook:
Seems like a typo. The old example leaves the whole file unencrypted.


autrilla:
Can you make this merge to `develop` instead of `master`? Thanks!

oskarkook:
Changed the branch to `develop`.

oskarkook:
Alright, should be good to go now.

Issue #527, opened 2019-09-12
Fix exec mod for Windows


ajvb:
Follow up to https://github.com/mozilla/sops/pull/526

Either add Windows support for `exec-env` and `exec-file` or make `exec-env` and `exec-file` linux/darwin specific.

cc: @ancat

ajvb:
Fixed in #532

PR #532, opened 2019-09-15
exec-file and exec-env subcommands, take two

ancat:
This time I split up windows and not-windows functionality, since to my understanding the APIs for darwin/linux/freebsd should all be the same here. I was able to compile to all four of these targets, and this split lets the two subcommands work on Windows for the most part (just no setuid or fifos). This should hopefully address #527.

cc @ajvb


cmd/sops/subcommand/exec/exec.go
```diff
@@ -0,0 +1,100 @@
+package exec
+
+import (
+	"log"
```
ajvb: Instead of using the stdlib `log` package, can we switch to using `logrus` and importing the logrus config in sops?

i.e.
https://github.com/mozilla/sops/blob/14a22d7a7060a0fba06c5dd716898d47728f0fce/cmd/sops/subcommand/publish/publish.go#L15-L26

cmd/sops/subcommand/exec/exec.go
```diff
@@ -0,0 +1,100 @@
...
+	if opts.User != "" {
+		SwitchUser(opts.User)
+	}
+
+	if runtime.GOOS == "windows" && opts.Fifo {
+		log.Print("no fifos on windows, use --no-fifo next time")
+		opts.Fifo = false
+	}
+
+	dir, err := ioutil.TempDir("/tmp/", ".sops")
```
davidovich: using `/tmp/` here will break Windows users. Can you use the `""` empty string to allow os dependant temp location ?

cmd/sops/subcommand/exec/exec.go
```diff
@@ -0,0 +1,100 @@
...
+		go WritePipe(filename, opts.Plaintext)
+	} else {
+		handle := GetFile(dir)
+		handle.Write(opts.Plaintext)
+		handle.Close()
+		filename = handle.Name()
+	}
+
+	placeholdered := strings.Replace(opts.Command, "{}", filename, -1)
+	cmd := exec.Command("/bin/sh", "-c", placeholdered)
```
davidovich: Why not call the command directly ? calling through `/bin/sh` will break in native Windows.

cmd/sops/subcommand/exec/exec.go
ancat: good catch, will do.

cmd/sops/subcommand/exec/exec.go
ancat: Properly splitting up the placeholdered command into an array is not that simple, it would require tokenizing the command (naively splitting by spaces risks changing the semantics) so it's easier to let `/bin/sh` do it for me. however you do have a good point that this will not work on Windows. Instead, I can do the same thing but use `cmd.exe /C`, like in [this StackOverflow post](https://stackoverflow.com/a/13013520). What do you think of that?

davidovich: Yes, I think that would be a good alternative:
```
	invoker := []string{"/bin/sh", "-c"}
	if runtime.GOOS == "windows" {
		invoker = []string{"cmd.exe", "/C"}
	}
	invoker = append(invoker, placeholdered...)
	cmd := exec.Command(invoker[0], invoker[1:]...)
```

ancat:
Alright, I made the requested changes! Take a look and let me know what you think. Thanks!

ajvb:
@autrilla Nope! Was just waiting for your +1

PR #541, opened 2019-09-26
[REVIEW] alias fix, needs test

rosscdh:
sops should support aws kms alias in the same way the aws cli does as per https://github.com/mozilla/sops/issues/381

I've added a simple check, please review and ill try add tests for it (first go contribution, go easy please)

### DIFF
--- a/kms/keysource.go
+++ b/kms/keysource.go
@@ -191,10 +191,12 @@ func (key MasterKey) createSession() (*session.Session, error) {
 	if matches != nil {
 		config := aws.Config{Region: aws.String(matches[1])}
 	} else {
-		if alias_matches == nil {
-				return nil, fmt.Errorf("No valid ARN found in %q", key.Arn)
-			} else {
-				config := aws.Config{Region: aws.String(alias_matches[0])}
+
+		if alias_matches != nil {
+			// its an alias the region is irrelevant
+			config := aws.Config{Region: aws.String("us-east-1")}
+		} else {
+			return nil, fmt.Errorf("No valid ARN found in %q", key.Arn)
 		}
 	}
 
--- a/kms/keysource.go
+++ b/kms/keysource.go
@@ -192,9 +192,9 @@ func (key MasterKey) createSession() (*session.Session, error) {
 		config := aws.Config{Region: aws.String(matches[1])}
 	} else {
 		if alias_matches == nil {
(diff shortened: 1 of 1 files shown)


kms/keysource.go
```diff
@@ -183,12 +183,24 @@ func (key MasterKey) createStsSession(config aws.Config, sess *session.Session)
 
 func (key MasterKey) createSession() (*session.Session, error) {
 	re := regexp.MustCompile(`^arn:aws[\w-]*:kms:(.+):[0-9]+:(key|alias)/.+$`)
+	// hmm, have to change arn? this is ugly but we must have the region
+	// implies us-east-1:alias/my-alias-name
+	re_alias := regexp.MustCompile(`^(.+):alias/.+$`)
```
autrilla: Follow Go naming conventions, local variables should be camelCase.

kms/keysource.go
```diff
@@ -183,12 +183,24 @@ func (key MasterKey) createStsSession(config aws.Config, sess *session.Session)
...
+	} else {
 
-	config := aws.Config{Region: aws.String(matches[1])}
+		if alias_matches != nil {
+			// its an alias the region is irrelevant
+			config := aws.Config{Region: aws.String(matches[0])}
+		} else {
+			return nil, fmt.Errorf("No valid ARN found in %q", key.Arn)
+		}
+	}
```
autrilla: Unnest this into an else if and an else.

rosscdh: Thanks! I'm not sure about my butchering the arn pattern to force the region into the alias?

autrilla: Have you tested what happens if you provide a bogus region to the alias? e.g. `blah:alias/foo`.

I don't currently have an AWS account set up so this is not particularly easy for me to test :(

rosscdh:
I imagine a ErrSharedConfigInvalidCredSource would be thrown? But in truth
no, in fact I've not yet written any tests for this component as I'm
floating the idea to get some feedback on the need for alias support?

autrilla:
I don't mean automated tests, just whether you've tested that this actually works with aliases, and what happens when you provide either the wrong region, or a region that doesn't exist. I'm not asking about what behavior should be implemented, I'm just wondering what the behavior of the AWS SDK is.

PR #545, opened 2019-10-08
add filestatus command

### LINKED ISSUE #460: Implement some kind of query/status option
It'd be really useful to be able to query the status of a file. I'm currently using `grep -q -x  "sops:" file.yml` to test if a file is encrypted for:

- pre-commit hook, to test if a file needs to be encrypted 
- git diff, to test if a file needs to be decrypted. Just using `sops -d` can error if the diff is between an encrypted file and an unencrypted one with the git diff:

```diff
-        # last lines of yaml
+sops metadata not found
```

Relying on some generic test (like `grep`) is not that difficult but
- it would be harder to make portable for different formats
- it'd feel a lot nicer if sops could tell us about files itself since it already does that for error cases.

## Proposal

I'm thinking something like:

```bash
$ sops --status file.yml
decrypted
$ sops -e -i file.yml
$ sops --status file.yml
encrypted
```

## Alternate ideas

- This seems conceptually similar to the `--verify` option proposed in #437 but I don't really understand what problem #437 is trying to solve
- Sops could implement a `--decrypt-or-cat` function that would also be quiet useful and would address the _git diff_ portion of my use case

endorama:
Closes #460

Add `filestatus` command, reporting if the file is in encrypted or unencrypted state.

I reused `ensureNoMetadata` logic, thus the command would return `0` when the file is not encrypted and `203` when the file is encrypted (respecting the `codes.FileAlreadyEncrypted` error code)

It does not output the error message as returned by `ensureNoMetadata`, preferring a simpler output: "File is encrypted" or "File is unencrypted".

As I'm not so fond on `sops` internals, I'm sure there are multiple things to review, I'll gladly update the code to reflect any suggestion.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/getsops/sops/v3/cmd/sops/codes"
 	"github.com/getsops/sops/v3/cmd/sops/common"
 	"github.com/getsops/sops/v3/cmd/sops/subcommand/exec"
+	filestatuscmd "github.com/getsops/sops/v3/cmd/sops/subcommand/filestatus"
 	"github.com/getsops/sops/v3/cmd/sops/subcommand/groups"
 	keyservicecmd "github.com/getsops/sops/v3/cmd/sops/subcommand/keyservice"
 	publishcmd "github.com/getsops/sops/v3/cmd/sops/subcommand/publish"
@@ -425,6 +426,38 @@ func main() {
 					log.Errorf("Error running keyservice: %s", err)
 					return err
 				}
+				return nil
+			},
+		},
+		{
+			Name:      "filestatus",
+			Usage:     "check the status of the file, returning encryption status",
+			ArgsUsage: `file`,
+			Flags:     []cli.Flag{},
+			Action: func(c *cli.Context) error {
+				if c.NArg() < 1 {
+					return common.NewExitError("Error: no file specified", codes.NoFileSpecified)
+				}
+
+				fileName := c.Args()[0]
+				inputStore := inputStore(c, fileName)
+				opts := filestatuscmd.Opts{
+					InputStore: inputStore,
+					InputPath:  fileName,
+				}
+
+				status, err := filestatuscmd.FileStatus(opts)
+				if err != nil {
+					return err
+				}
+
+				json, err := encodingjson.Marshal(status)
+				if err != nil {
+					return common.NewExitError(err, codes.ErrorGeneric)
+				}
+
+				fmt.Println(string(json))
+
 				return nil
 			},
 		},
--- a/cmd/sops/subcommand/filestatus/filestatus.go
+++ b/cmd/sops/subcommand/filestatus/filestatus.go
@@ -1,0 +1,60 @@
+package filestatus
+
+import (
+	"fmt"
+
+	"github.com/getsops/sops/v3"
+	"github.com/getsops/sops/v3/cmd/sops/common"
+)
+
+// Opts represent the input options for FileStatus
+type Opts struct {
+	InputStore sops.Store
+	InputPath  string
+}
+
+// Status represents the status of a file
+type Status struct {
+	// Encrypted represents whether the file provided is encrypted by SOPS
+	Encrypted bool `json:"encrypted"`
+}
+
+// FileStatus checks encryption status of a file
+func FileStatus(opts Opts) (Status, error) {
+	encrypted, err := cfs(opts.InputStore, opts.InputPath)
+	if err != nil {
+		return Status{}, fmt.Errorf("cannot check file status: %w", err)
+	}
+	return Status{Encrypted: encrypted}, nil
+}
+
+// cfs checks and reports on file encryption status.
+//
+// It tries to decrypt the input file with the provided store.
+// It returns true if the file contains sops metadata, false
+// if it doesn't or Version or MessageAuthenticationCode are
+// not found.
+// It reports any error encountered different from
+// sops.MetadataNotFound, as that is used to detect a sops
+// encrypted file.
+func cfs(s sops.Store, inputpath string) (bool, error) {
+	tree, err := common.LoadEncryptedFile(s, inputpath)
+	if err != nil && err == sops.MetadataNotFound {
+		return false, nil
+	}
+	if err != nil {
+		return false, fmt.Errorf("cannot load encrypted file: %w", err)
+	}
+
+	// NOTE: even if it's a file that sops recognize as containing
+	// valid metadata, we want to ensure some metadata are present
+	// to report the file as encrypted.
+	if tree.Metadata.Version == "" {
+		return false, nil
+	}
+	if tree.Metadata.MessageAuthenticationCode == "" {
+		return false, nil
+	}
+
+	return true, nil
+}
--- a/cmd/sops/subcommand/filestatus/filestatus_internal_test.go
+++ b/cmd/sops/subcommand/filestatus/filestatus_internal_test.go
@@ -1,0 +1,52 @@
+package filestatus
+
+import (
+	"path"
+	"testing"
+
+	"github.com/getsops/sops/v3/cmd/sops/common"
+	"github.com/getsops/sops/v3/config"
+	"github.com/stretchr/testify/require"
+)
+
+const repoRoot = "../../../../"
+
+func fromRepoRoot(p string) string {
+	return path.Join(repoRoot, p)
+}
+
+func TestFileStatus(t *testing.T) {
+	tests := []struct {
+		name              string
+		file              string
+		expectedEncrypted bool
+	}{
+		{
+			name:              "encrypted file should be reported as such",
+			file:              "example.yaml",
+			expectedEncrypted: true,
+		},
+		{
+			name: "plain text file should be reported as cleartext",
+			file: "functional-tests/res/plainfile.yaml",
+		},
+		{
+			name: "file without mac should be reported as cleartext",
+			file: "functional-tests/res/plainfile.yaml",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f := fromRepoRoot(tt.file)
+			s := common.DefaultStoreForPath(config.NewStoresConfig(), f)
+			encrypted, err := cfs(s, f)
+			require.Nil(t, err, "should not error")
+			if tt.expectedEncrypted {
+				require.True(t, encrypted, "file should have been reported as encrypted")
+			} else {
+				require.False(t, encrypted, "file should have been reported as cleartext")
+			}
+		})
+	}
+}
--- a/functional-tests/res/plainfile.yaml
+++ b/functional-tests/res/plainfile.yaml
@@ -1,0 +1,1 @@
+hello: world


cmd/sops/main.go
```diff
@@ -443,6 +444,55 @@ func main() {
...
+		{
+			Name:      "filestatus",
+			Usage:     "check the status of the file, returning encryption status",
+			ArgsUsage: `file`,
+			Flags:     []cli.Flag{},
+			Action: func(c *cli.Context) error {
+				// _, err := config.FindConfigFile(".")
+				// if err != nil {
+				// 	return common.NewExitError(err, codes.ErrorGeneric)
+				// }
```
endorama: Should this command check for the config file?

cmd/sops/main.go
```diff
@@ -443,6 +444,55 @@ func main() {
...
+				}
+				branches, err := opts.InputStore.LoadPlainFile(fileBytes)
+				if err != nil {
+					return common.NewExitError(fmt.Sprintf("Error unmarshalling file: %s", err), codes.CouldNotReadInputFile)
+				}
+				if err := ensureNoMetadata(opts, branches[0]); err != nil {
+					return common.NewExitError("File is encrypted", codes.FileAlreadyEncrypted)
+				}
+
+				fmt.Println("File is unencrypted")
```
endorama: Is `fmt` the proper way to output such an informative message?

cmd/sops/main.go
```diff
@@ -443,6 +444,55 @@ func main() {
...
+				if err != nil {
+					return common.NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
+				}
+				branches, err := opts.InputStore.LoadPlainFile(fileBytes)
+				if err != nil {
+					return common.NewExitError(fmt.Sprintf("Error unmarshalling file: %s", err), codes.CouldNotReadInputFile)
+				}
+				if err := ensureNoMetadata(opts, branches[0]); err != nil {
+					return common.NewExitError("File is encrypted", codes.FileAlreadyEncrypted)
+				}
```
endorama: This code is a duplicate of https://github.com/mozilla/sops/blob/4b5b7ae308839fb4bbd6cc35851f93c01322643b/cmd/sops/encrypt.go#L58-L69

Should this duplication be avoided? If so, what would be the best way?

cmd/sops/main.go
autrilla: Not for now, at least, since we aren't using anything from the config file

cmd/sops/main.go
autrilla: Looks fine to me. I'm not sure whether we should use a machine-parseable and extensible format (probably JSON), though. It would let us put more status information there in the future, e.g. whether the MAC is okay.

cmd/sops/main.go
autrilla: Yeah, I think we could extract that function to the `sops/cmd/common` package so that we can call it from both places.

cmd/sops/main.go
```diff
@@ -443,6 +444,55 @@ func main() {
 				return nil
 			},
 		},
+		{
+			Name:      "filestatus",
+			Usage:     "check the status of the file, returning encryption status",
+			ArgsUsage: `file`,
+			Flags:     []cli.Flag{},
+			Action: func(c *cli.Context) error {
```
autrilla: Could you extract the body of this function to a package in the `subcommands` package? Like with https://github.com/mozilla/sops/blob/master/cmd/sops/subcommand/updatekeys/updatekeys.go and others. I think it makes the code easier to navigate vs having a lot of unrelated functionality in the same file.

cmd/sops/main.go
```diff
@@ -443,6 +444,55 @@ func main() {
...
+				fileBytes, err := ioutil.ReadFile(opts.InputPath)
+				if err != nil {
+					return common.NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
+				}
+				branches, err := opts.InputStore.LoadPlainFile(fileBytes)
+				if err != nil {
+					return common.NewExitError(fmt.Sprintf("Error unmarshalling file: %s", err), codes.CouldNotReadInputFile)
+				}
+				if err := ensureNoMetadata(opts, branches[0]); err != nil {
+					return common.NewExitError("File is encrypted", codes.FileAlreadyEncrypted)
```
autrilla: Hmmm, I'm wondering whether we should return an error when the file is not encrypted, or the other way around. What was your reasoning for choosing this over the alternative?

cmd/sops/main.go
autrilla: Never mind, I misread -- no need to deduplicate this.

cmd/sops/main.go
endorama: Would something like `{"encrypted": true|false}` be ok?

cmd/sops/main.go
endorama: I'll remove this then.

cmd/sops/main.go
endorama: Mainly due to the fact that there was an error code for "file is already encrypted" and not the other way round. So I would expect users to be already used to such an error.

IMO none is an error, and using JSON output like you requested in a comment above everything makes more sense and this error can probably be removed.

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	branches, err := opts.InputStore.LoadPlainFile(fileBytes)
+	if err != nil {
+		return Status{}, common.NewExitError(fmt.Sprintf("Error unmarshalling file: %s", err), codes.CouldNotReadInputFile)
+	}
+	hasMetadata := checkMetadata((branches[0]))
+	return Status{Encrypted: hasMetadata}, nil
+}
+
+func checkMetadata(branch sops.TreeBranch) bool {
```
endorama: As `ensureNoMetadata` is declared private in `main` package (`cmd/sops/encrypt.go:48`) I duplicated the code, but `checkMetadata` could be in `common` package and being used in `ensureNoMetadata` too to avoid duplication.

cmd/sops/main.go
endorama: done in https://github.com/mozilla/sops/pull/545/commits/6e32720f45eb9a149f810540ee64e689d8662308

cmd/sops/main.go
autrilla: Yeah, I agree. Let's make none of them be an error for now.

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+
+	"go.mozilla.org/sops"
+	"go.mozilla.org/sops/cmd/sops/codes"
+	"go.mozilla.org/sops/cmd/sops/common"
+)
+
+// Opts rapresent input option for FileStatus
```
autrilla:
```suggestion
// Opts represent the input options for FileStatus
```

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+type Opts struct {
+	InputStore sops.Store
+	// OutputStore       sops.Store
+	InputPath string
+	// KeyServices       []keyservice.KeyServiceClient
+	// KeyGroups         []sops.KeyGroup
+	// GroupThreshold    int
```
autrilla: Remove the commented lines since they're not used

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// Status rapresent file status
```
autrilla:
```suggestion
// Status represents the status of a file
```

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+type Status struct {
+	Encrypted bool
```
autrilla:
```
// Encrypted represents whether the file provided is encrypted by SOPS
```

cmd/sops/subcommand/filestatus/filestatus.go
autrilla: Eh, this is simple enough that I think it's okay.

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+		return Status{}, common.NewExitError(fmt.Sprintf("Error unmarshalling file: %s", err), codes.CouldNotReadInputFile)
+	}
+	hasMetadata := checkMetadata((branches[0]))
+	return Status{Encrypted: hasMetadata}, nil
+}
+
+func checkMetadata(branch sops.TreeBranch) bool {
+	for _, b := range branch {
+		if b.Key == "sops" {
+			return true
```
jvehent: isn't that a bit weak as a test? I'd suggest at least verifying that a `mac` is set and one key exists under the various sections.

cmd/sops/subcommand/filestatus/filestatus.go
```diff
@@ -0,0 +1,47 @@
...
+	// KeyGroups         []sops.KeyGroup
+	// GroupThreshold    int
+}
+
+// Status rapresent file status
+type Status struct {
+	Encrypted bool
+}
+
+func FileStatus(opts Opts) (Status, error) {
```
ajvb: Would be nice to have a function documentation comment for this function since it is public.

endorama: Added in commit https://github.com/mozilla/sops/pull/545/commits/80686aba224a6fed3aec410d0f75fc50065d2384

cmd/sops/subcommand/filestatus/filestatus.go
endorama: Removed in commit https://github.com/mozilla/sops/pull/545/commits/80686aba224a6fed3aec410d0f75fc50065d2384

cmd/sops/subcommand/filestatus/filestatus.go
endorama: Added in commit https://github.com/mozilla/sops/pull/545/commits/80686aba224a6fed3aec410d0f75fc50065d2384

cmd/sops/main.go
autrilla: Looks good to me.

cmd/sops/subcommand/filestatus/filestatus.go
autrilla: It's indeed a bit weak. It's what we do elsewhere as an easy test to check whether the file is encrypted, but maybe as you say we should have a stricter test here.

endorama: May I propose to go with this version for consistency, and work on a more complex test (that we can discuss more) on a separate one? I'm willing to work on that.
We can also extend a bit the scope and extract the check so both places can use the same code.

kowalczykp:
Any update on that? It would be a really nice feature

mazdack:
any update? would be really nice to have this command

autrilla:
No updates. The review comments haven't been addressed in this PR, plus there's merge conflicts. We'd be happy to review a PR that builds on this one an handles the review comments.

endorama:
@autrilla May you let me know your opinion on https://github.com/mozilla/sops/pull/545#discussion_r339676101?

I was waiting on a reply there, other concerns should have been addressed as requested.

cmd/sops/subcommand/filestatus/filestatus.go
endorama: Resolved in 80686aba224a6fed3aec410d0f75fc50065d2384

autrilla:
> @autrilla May you let me know your opinion on [#545 (comment)](https://github.com/mozilla/sops/pull/545#discussion_r339676101)?

It wouldn't take a lot of work to have a slightly better check by doing what jvehent suggests, e.g. ensuring there's a `mac` field.

endorama:
@autrilla I resolved conflicts and rebased onto latest `develop` so this should be now safe to merge again.

I started working on how to improve the check and I found out that using an implementation of `EncryptedFileLoader.LoadEncryptedFile(in []byte) (Tree, error)` on a non-encrypted file results in a `MetadataNotFound` error. Would this be ok as a deeper test? It feels "right" from my point of view. Is it enough or once metadata are found should it perform some other checks?

autrilla:
> @autrilla I resolved conflicts and rebased onto latest `develop` so this should be now safe to merge again.
> 
> I started working on how to improve the check and I found out that using an implementation of `EncryptedFileLoader.LoadEncryptedFile(in []byte) (Tree, error)` on a non-encrypted file results in a `MetadataNotFound` error. Would this be ok as a deeper test? It feels "right" from my point of view. Is it enough or once metadata are found should it perform some other check?

Trying to load the file sounds good to me.

onedr0p:
@autrilla any status here, or are you waiting on the author to rebase this PR again?

onedr0p:
Bump <3

autrilla:
Not all the issues are addressed. In particular, I asked for a better test that the file is encrypted. It's also missing documentation and tests, and as you mention, there's conflicts.

endorama:
Hello,

I have redone the implementation.   
The check is performed in the function `cfs` and:
- tries to load the encrypted file
- if it fails, report as not encrypted
- if encryption works, checks for `version` and `mac` keys to be present

`filestatus` subcommand now output JSON as requested (`{"encrypted": true|false}`).

I rebased onto master, so I'm not sure why it's still complaining about the conflict.

@autrilla is this implementation ok? I also added some test for it. The `cfs` function is private in the `filestatus` package, let me know where is the most appropriate place to put it into so it can be reused in other areas as well.

Thanks and sorry for the long wait!

Pehesi97:
How is the progress in this PR? Do you need any help?

aDingil:
alt text

ggordan:
Hi @ajvb @autrilla @jvehent, I see that roughly a year ago the milestone for this issue was changed to [v3.8.0](https://github.com/mozilla/sops/milestone/4).

Would that suggest that this PR is "complete" meaning that when v3.8.0 is released, this will be included, or are there still things that need to be addressed for the maintainers to be happy to merge this? If the answer is no, can we get a indication of what is needed, I'd be more than happy to help out.

Subsequently, is there a rough estimate on when v3.8.0 is expected to land?

Thanks,
Gordan

felixfontein:
@endorama would you mind rebasing another time and signing off your commits? Would be great if we could finally get this merged :)

davinkevin:
Can't wait to use it!

endorama:
Hello everybody and sorry for the long wait! I'd love to see this finally being merge and I have some time to dedicate to it.

Rebasing the old work was quite complex, so given the overall addition in this PR was small I re-implemented it on top of the latest `master` branch. I kept the implementation  the same with the same tests.

(The second force push is due to the DCO, which I didn't sign the first time)

Waiting on the review ❤

felixfontein (approved):

felixfontein:
I did some extensive tests, and it looks good! The method of detecting encrypted files is pretty good; obviously you can fake it with fake values, like the follwing YAML file:
```
sops:
    lastmodified: "2020-01-01T00:00:00Z"
    mac: x
    pgp:
        - created_at: "2020-01-01T00:00:00Z"
          enc: |-
            x
          fp: x
    version: x
```
(and similar variants for other storages), but I think that's acceptable, and it also seems to agree to what all reviewers so far asked for.

If nobody objects (especially from @getsops/maintainers side), I propose to merge this soon!

sabre1041 (approved):
LGTM

felixfontein:
@endorama thanks a lot for contributing this!

Also thanks a lot to everyone who reviewed this and commented on this over all these years! I'm glad this finally made it :)

Issue #548, opened 2019-10-15
`decrypt.Data` does not support `ini` store by default


dnozay:
Is there a reason why the cli supports `.ini` files out of the box, but `decrypt.Data(...)` doesn't?
should it be using `DefaultStoreForPath` instead?

autrilla:
> Is there a reason why the cli supports `.ini` files out of the box, but `decrypt.Data(...)` doesn't?

Oversight!

> should it be using `DefaultStoreForPath` instead?

Maybe, yeah. We should probably just `"." + format` and pass that over to `DefaultStoreForPath`. Or maybe even improve the whole thing and keep a map of extensions to `Store`s.

dnozay:
added #549 to try and address this.

PR #549, opened 2019-10-15
fix for #548 - handle .ini files in `decrypt.Data`, add other helper

dnozay:


dnozay:
retest

dnozay:
looks like sometimes tests are failing with

```
$ vault secrets enable -version=1 kv
Error enabling: Post http://127.0.0.1:8200/v1/sys/mounts/kv: dial tcp 127.0.0.1:8200: connect: connection refused
```

dnozay:
@autrilla ^ WDYT?

cmd/sops/common/common.go
```diff
@@ -139,6 +139,24 @@ func IsIniFile(path string) bool {
...
+	case "json":
+		return &json.Store{}
+	case "dotenv":
+		return &dotenv.Store{}
+	case "ini":
+		return &ini.Store{}
+	case "binary":
+		return &json.BinaryStore{}
+	}
+	return &json.BinaryStore{}
```
ajvb: Instead of defaulting to `return &json.BinaryStore{}` here, I feel like it should fall back to `DefaultStoreForPath`, as that would keep with the current behavior.

ajvb:
@dnozay Could you change this PR to be merged against the `develop` branch?

cmd/sops/common/common.go
dnozay: fixed in `DefaultStoreForPathOrFormat`

dnozay:
@ajvb i changed the merge base to `develop` branch

decrypt/decrypt.go
```diff
@@ -24,26 +22,15 @@ func File(path, format string) (cleartext []byte, err error) {
...
-	case "json":
-		store = &sopsjson.Store{}
-	case "yaml":
-		store = &sopsyaml.Store{}
-	case "dotenv":
-		store = &sopsdotenv.Store{}
-	default:
-		store = &sopsjson.BinaryStore{}
-	}
+func DataWithStore(data []byte, store sops.Store) (cleartext []byte, err error) {
```
autrilla: Do you have any use case for which this should be public?

dnozay: Choosing the store is the contentious part. There are several strategies:
- based on a format
- based on an extension
- based on whatever custom logic
Example: you can use `go-bindata` to include assets.

if `decrypt` has the only stable api, then basically that says "trust me i know what i am doing"

dnozay:
I don't have a strong opinion with `Data` taking a format.

- Sometimes you don't know the format, so you can't use `decrypt.Data`; and you need to use non-stable APIs (anything outside of `decrypt`) to guess.
- Sometimes you don't have a file to work with (e.g. asset from `go-bindata`) but you may have a path, so you can't use `decrypt.File`.

cmd/sops/common/common.go
```diff
@@ -36,6 +36,37 @@ type Store interface {
...
+	return &json.Store{}
+}
+
+func newYamlStore() Store {
+	return &yaml.Store{}
+}
+
+// StoreConstructors is a mapping of formats to constructors.
+var StoreConstructors = map[string]storeConstructor{
+	"binary": newBinaryStore,
```
autrilla: I would prefer to store these as something similar to an enum:

```go
type Format int

const (
  Binary Format = iota
  Json
  Ini
  // ...
)
```

And

```go
var storeConstructors = map[Format]storeConstructor{
   Yaml: newYamlStore,
// ...
```

cmd/sops/common/common.go
```diff
@@ -36,6 +36,37 @@ type Store interface {
...
+func newJsonStore() Store {
+	return &json.Store{}
+}
+
+func newYamlStore() Store {
+	return &yaml.Store{}
+}
+
+// StoreConstructors is a mapping of formats to constructors.
+var StoreConstructors = map[string]storeConstructor{
```
autrilla: This does not need to be public, right?

cmd/sops/common/common.go
```diff
@@ -139,19 +170,40 @@ func IsIniFile(path string) bool {
 	return strings.HasSuffix(path, ".ini")
 }
 
+// DefaultStoreForFormat returns the correct format-specific implementation
+// of the Store interface given the format string
+func DefaultStoreForFormat(format string) Store {
```
autrilla: With my suggested changes, this should not be needed anymore.

cmd/sops/common/common.go
```diff
@@ -139,19 +170,40 @@ func IsIniFile(path string) bool {
...
+	storeConst, found := StoreConstructors[format]
+	if !found {
+		storeConst = StoreConstructors["binary"] // default
+	}
+	return storeConst()
+}
+
 // DefaultStoreForPath returns the correct format-specific implementation
 // of the Store interface given the path to a file
 func DefaultStoreForPath(path string) Store {
```
autrilla: We should instead return a `Format` for a given path.

cmd/sops/common/common.go
```diff
@@ -139,19 +170,40 @@ func IsIniFile(path string) bool {
...
-		return &ini.Store{}
+		storeConst = StoreConstructors["ini"]
 	}
-	return &json.BinaryStore{}
+	return storeConst()
+}
+
+// DefaultStoreForPathOrFormat returns the correct format-specific implementation
+// of the Store interface given the format if specified, or the path to a file
+func DefaultStoreForPathOrFormat(path, format string) Store {
```
autrilla: This shouldn't be needed anymore -- if a `Format` is provided, a store for that format should exist.

We could potentially add a lint to make sure this is the case for us, since the Go compiler won't let us prove it.

cmd/sops/common/common.go
dnozay: the `sops` CLI uses strings, so i'm not sure what you want you want me to use there.

cmd/sops/common/common.go
dnozay: Either `StoreConstructors` needs to be public or `DefaultStoreForFormat` needs to be public, so the CLI can consume it.

cmd/sops/common/common.go
dnozay: 1. this would not be backwards compatible.
2. how do you want me to name it? suggestions?

autrilla: This is not part of the public API, so it's okay to not be backwards compatible. `FormatForPath` would be good, I think.

cmd/sops/common/common.go
autrilla: The CLI should also convert from strings to `Format` ASAP

cmd/sops/common/common.go
autrilla: Yeah, you're right. I think I'd expose `Format` and the associated way to get a Store from a given format in the `go.mozilla.org/sops` package directly (so at the root of the repository) since it isn't command-line specific. It's not super important since it's not part of the public API, though, so feel free to just keep it here.

cmd/sops/common/common.go
dnozay: this is used in 3 places, for cli `--input-type`, cli `--output-type`, and in `decrypt.File`.

decrypt/decrypt.go
autrilla: Do you think `DataWithFormat` would be a more useful API? Considering `Store` is not really meant to be used by third parties. And I think we should re-export `Format` in this file as well. Keep in mind only the `decrypt` code should be public to external projects, and right now we're requiring them to access `go.mozilla.org/sops`.

autrilla:
By the way, sorry this initially simple PR has turned into a somewhat major refactor of how we handle formats :(

autrilla:
@dnozay looks like you forgot to commit the formats package though

dnozay:
@autrilla - sorry about that; it is there now.

autrilla:
No worries! All good from my end, I'll merge now, and if @ajvb has any concerns he can raise them later and we can revert.

PR #553, opened 2019-10-23
version-based tags for Docker Images in Docker hub

### LINKED ISSUE #542: Consider using version-based tags for Docker Images in Docker hub
Currently, it appears the only tags pushed to Dockerhub are the `latest` tag and the git commit sha.

For example, when the newest `3.4.0` image was tagged, it was tagged as

- latest
- 14a22d7a7060a0fba06c5dd716898d47728f0fce

It's often useful to have more meaningful tags created that match the actual version.  With the latest `3.4.0` release it would preferred that it also be tagged with the version.  So the following docker tags.

- latest
- 14a22d7a7060a0fba06c5dd716898d47728f0fce
- 3.4.0

Additionally, it's not to have simplified major/minor tags.  So the ideal set of tags would be  

- latest
- 14a22d7a7060a0fba06c5dd716898d47728f0fce
- 3.4.0
- 3.4
- 3

imskr:
Fixes #542

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,6 @@ jobs:
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
                 ${GOPATH}/src/go.mozilla.org/sops/bin/ci/deploy_dockerhub.sh "latest"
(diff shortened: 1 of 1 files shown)


imskr:
@ajvb Please review this!

imskr:
@autrilla I think I should replace 3.4.0 with `$DOCKER_TAG`

PR #554, opened 2019-10-24
Fix --encrypted-regex K8 Secrets Example in README

devstein:
I am using `SOPS` to encrypt our `K8` secrets and would only like to encrypt `data` and `stringData` fields. The regex in the current example also encrypts the secret `metadata` field, so I fixed it to only match the exact `data` and `stringData` keys


autrilla:
@devstein this needs to be changed to merge to `develop` instead of `master`. Could you do that?

devstein:
@autrilla Looks like its fixed on `deverlop` already!

PR #558, opened 2019-10-27
Sanitize hostname used for AWS STS role session name

### LINKED ISSUE #441: hostname in STS role session name may cause an error in the AssumeRole call
when assuming a role for KMS decryption, if a hostname contains characters that are not allowed by STS in a role session name, the operation will fail.

in code:
https://github.com/mozilla/sops/blob/a2d0328e35e6e37b51f3ad468dc6f213c7b44014/kms/keysource.go#L169

amazon reference:
https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html#options

this happened to us for a hostname with invalid characters in it, e.g. "hostname-(123)"

imsky:
From official docs for --role-session-name (https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html):
> The regex used to validate this parameter is a string of characters consisting of upper- and lower-case alphanumeric characters with no spaces. You can also include underscores or any of the following characters: =,.@-

This fixes #441, which occurs when the hostname includes spaces and parentheses

### DIFF
--- a/kms/keysource.go
+++ b/kms/keysource.go
@@ -167,7 +167,7 @@ func (key MasterKey) createStsSession(config aws.Config, sess *session.Session) 
 	}
 	stsRoleSessionNameRe, err := regexp.Compile("[^a-zA-Z0-9=,.@-]+")
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Failed to compile STS role session name regex: %v", err)
 	}
 	sanitizedHostname := stsRoleSessionNameRe.ReplaceAllString(hostname, "")
 	stsService := sts.New(sess)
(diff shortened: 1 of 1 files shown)


kms/keysource.go
```diff
@@ -165,8 +165,13 @@ func (key MasterKey) createStsSession(config aws.Config, sess *session.Session)
 	if err != nil {
 		return nil, err
 	}
+	stsRoleSessionNameRe, err := regexp.Compile("[^a-zA-Z0-9=,.@-]+")
+	if err != nil {
+		return nil, err
```
autrilla: Please return a more meaningful error here. Wrap it and add some context so users know what it’s about.

imsky: @autrilla please take another look

PR #531, opened 2019-09-14
Upgrade to go 1.13

chenrui333:
Relates to https://github.com/Homebrew/homebrew-core/pull/44191

Issue #540, opened 2019-09-26
Tag releases with v prefix


carlpett:
For importing sops as a Go module in other code, the release tags should be prefixed with a v, eg `v3.4.0` rather than `3.4.0`.

See https://github.com/golang/go/wiki/Modules#modules (emphasis mine):
>Modules must be semantically versioned according to semver, usually in the form v(major).(minor).(patch), such as v0.1.0, v1.2.3, or v1.5.0-rc.1. **The leading v is required.** If using Git, tag released commits with their versions.

If this sounds acceptable, could you also add a v3.4.0 tag "on top" of 3.4.0?

autrilla:
I pushed the v3.4.0 tag for now, and I guess we'll need to push an additional v tag every release, or stop pushing non-v tags.

ajvb:
Thanks @autrilla - I can come back to this before the next release and update the travis config and such to support it.

davidovich:
I believe this can be closed by #566

PR #566, opened 2019-11-06
Upgrade sops to go 1.13

ajvb:
Upgrade sops to go 1.13, making proper use of go modules internally - https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

fyi @g-k - Until this is merged into master and released, `sops` cannot be imported using go modules in go 1.13

Supersedes #531 
Relates to https://github.com/Homebrew/homebrew-core/pull/44191


README.rst
```diff
@@ -33,7 +33,7 @@ For the adventurous, unstable features are available in the `develop` branch, wh
         $ git checkout develop
         $ make install
 
-(requires Go >= 1.12)
```
autrilla: Also update the go get a few lines above

autrilla:
My knowledge about Go modules is pretty much zero, but is there a way we could separate the versioning of the Go codebase from the versioning of the sops binary, which is effectively the only thing we actually version with semver? If we do that, could we call this 'v1' of the codebase and not have it in the import path? Or do you have the have the 'v1' there anyway?

ajvb:
> My knowledge about Go modules is pretty much zero, but is there a way we could separate the versioning of the Go codebase from the versioning of the sops binary, which is effectively the only thing we actually version with semver? If we do that, could we call this 'v1' of the codebase and not have it in the import path? Or do you have the have the 'v1' there anyway?

The thing I'm not really sure on is how changing the `go.mod` at the root of the repo exactly impacts the sops binary. Using `go 1.13`, `go get -u go.mozilla.org/sops/cmd/sops` works exactly as you'd expect it to right now. I'm curious if with this change we would actually need to update it to `go get -u go.mozilla.org/sops/v3/cmd/sops` or whether `go get -u go.mozilla.org/sops/cmd/sops` would continue to work as expected.


Additionally, forgot about https://github.com/mozilla/sops/issues/540, will want to include that in this PR.

autrilla:
I see. I have to think about this a bit more, I'll try to do that over the weekend, but feel free to merge to develop for now.

Basically, I think there's two things we need to version here:

1) `go.mozilla.org/sops/cmd/sops`
2) `go.mozilla.org/sops/decrypt`

And everything else should be unversioned if we can get away with it. As I said, I'll look and see if this is actually possible. A quick read of go modules makes me think it's not.

README.rst
```diff
@@ -28,12 +28,12 @@ For the adventurous, unstable features are available in the `develop` branch, wh
 
 .. code:: bash
 
-	$ go get -u go.mozilla.org/sops/cmd/sops
+	$ go get -u go.mozilla.org/v3/sops/cmd/sops
```
jvehent: typo

ajvb:
@autrilla From what I can tell from my experiments it seems like this might cause only minor breakage for some users.

One concern that I had was whether this could require all users using the `sops/decrypt` module within their codebases to start using go modules. This does not seem to be the case, surprisingly:

```bash
root@18591ccbf56d:/go/t# go version
go version go1.9.7 linux/amd64

root@18591ccbf56d:/go/t# cat ../src/github.com/ajvb/modules-testing-sops/go.mod              
module github.com/ajvb/modules-testing-sops/v2

go 1.13

root@18591ccbf56d:/go/t# cat ../src/github.com/ajvb/modules-testing-sops/anotherfile.go 
package modules_testing_sops

import "fmt"

func AnotherExampleFunc(msg, secondMsg string) {
        fmt.Println(msg)
        fmt.Println(secondMsg)
}

root@18591ccbf56d:/go/t# cat ../src/github.com/ajvb/modules-testing-sops/submodule/example.go 
package submodule

import "github.com/ajvb/modules-testing-sops/v2"

func SubmoduleFunc(msg string) {
        modules_testing_sops.AnotherExampleFunc(msg, msg)
}

root@18591ccbf56d:/go/t# cat main.go 
package main

import "github.com/ajvb/modules-testing-sops/submodule"

func main() {
submodule.SubmoduleFunc("test")
}

root@18591ccbf56d:/go/t# go run main.go 
test
test
```

Though the above is seemingly using code with `import "github.com/ajvb/modules-testing-sops/v2"` and is running on `go 1.9`, it seems to work as expected without having the specify the v2 itself.

As well, `go get` to run the binary (or more explicitly, outside the context of a go module) cannot make use of versions, so you should be able to still run `go get -u go.mozilla.org/sops/cmd/sops` and get `master`.

I think where this might cause stuff to break is that people who have sops with a commit pin in their `go.mod` like [`go.mozilla.org/sops v0.0.0-20190611200209-e9e1e87723c8`](https://github.com/mozilla-services/autograph/blob/2530d5e7f376b66ca39d58eed22bc4fd96286e9e/go.mod#L38) and want to use it with `go 1.13` will need to change it to `go.mozilla.org/sops/v3 v3.4.1` or similar.

Please feel free to double check me as I still feel very new to go modules, but I think we are in the relative clear for breakage.

Lastly, after thinking about it a bit, I think that versioning the `decrypt` module and the sops binary separately would be a major change in how we structure the sops project and I don't quite see the value as long as we can reduce the breakage for our users. Keeping the binary and modules in sync version wise allows for very clear expectations on what versions of the binary or various modules will work with each other. Having a case where versions 1.X of the decrypt module work with 3.X through 4.4 of the binary or something could create a lot of confusion, both for us and for our users.

go.mod
```diff
@@ -19,18 +17,17 @@ require (
...
 	github.com/lib/pq v1.2.0
 	github.com/mitchellh/go-wordwrap v1.0.0
-	github.com/mozilla-services/yaml v0.0.0-20180922153656-28ffe5d0cafb
+	github.com/mozilla-services/yaml v0.0.0-20191106225358-5c216288813c
 	github.com/pkg/errors v0.8.1
 	github.com/sirupsen/logrus v1.4.2
-	github.com/smartystreets/goconvey v0.0.0-20190710185942-9d28bd7c0945 // indirect
 	github.com/stretchr/testify v1.3.0
 	go.mozilla.org/gopgagent v0.0.0-20170926210634-4d7ea76ff71a
+	go.mozilla.org/sops v0.0.0-20190912205235-14a22d7a7060 // indirect
```
carlpett: This seems worrying? Is there any remaining references to the old path, or does it go away with `go mod tidy`?

autrilla:
@ajvb thanks a lot for explaining. I am very surprised things still work for older versions of Go, so since that's the case, I have no objections to merging this whatsoever!

go.mod
ajvb: @carlpett This is interesting. Running `go mod tidy` does remove it, but then `go build` adds it back. I think the changes in this PR might need to be added to master and tagged before this can be properly fixed.

ajvb:
@autrilla Of course! Same hehe and sounds good.

Makefile
```diff
@@ -10,7 +10,7 @@ all: test vet generate install functional-tests
 origin-build: test vet generate install functional-tests-all
 
 install:
-	$(GO) install go.mozilla.org/sops/cmd/sops
+	$(GO) install go.mozilla.org/sops/v3/cmd/sops
```
ajvb: @carlpett Not changing this was the cause of the addition of the old sops path to `go.mod`.

PR #569, opened 2019-11-09
NACL BOX master key support

jvehent:
This is work in progress and not yet functional.

The goal is to provide a new type of master key using the NaCl Box protocol that uses local private keys. It's meant to provide an alternative to PGP for those who want to encrypt files locally, without relying on hosted KMS.

I added a new cmdline tool called `makenaclboxkey` that generates a keypair under `$HOME/.sops/naclbox/`.

Most of the logic is in place, but I need to finish the integration with the store. Data key encryption works, but `nonce` and `ephemeralpubkey` aren't captured in the file metadata.

Early review/feedback much appreciated.


cmd/makenaclboxkey/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"golang.org/x/crypto/nacl/box"
+)
+
+// KeyPair is a public and private key usable with NACL BOX
+type KeyPair struct {
+	PublicKey, PrivateKey string
+}
+
+func main() {
```
ajvb: Why not have this as a subcommand under the `sops` binary? Could be `sops naclbox-gen` or similar?

cmd/makenaclboxkey/main.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	keypair := KeyPair{
+		PublicKey:  base64.StdEncoding.EncodeToString(pub[:]),
+		PrivateKey: base64.StdEncoding.EncodeToString(priv[:]),
+	}
+	out, err := json.MarshalIndent(keypair, "", "    ")
+	if err != nil {
+		panic(err)
+	}
+	os.MkdirAll(os.Getenv("HOME")+"/.sops/naclbox", 0750)
```
ajvb: We will need to make sure this is compliant with Windows before we ship.

jvehent:
I wasn't sure if you wanted this in a subcommand, so thanks for that piece of information.
I also want to implement private key wrapping with a passphrase entered via pinentry before shipping this. Cleartext private keys are bad.

cmd/makenaclboxkey/main.go
autrilla: Agreed, but I'd prefer to keep the longer `makenaclboxkey` name.

naclbox/keysource.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// Encrypt takes a sops data key, encrypts it with NACL BOX
+// and stores the result in the EncryptedKey field
+func (key *MasterKey) Encrypt(dataKey []byte) error {
+	// NACL BOX uses two keys pairs to seal secret data: sender and recipient.
+	// When used in Sops, we don't care much about the sender keypair, but it is
+	// required for the protocol to work, so we issue an ephemeral keypair instead
+	// and store its pubkey alongside the encrypted data key, so it can later
+	// be decrypted.
```
autrilla: Should this say "so it can later be reencrypted"? Unless I'm misunderstanding, you're generating a key pair and throwing away the priv key (since it's not going to decrypt anything), and we need to keep it so we can reencrypt the data key in the future if needed without generating a completely new keypair.

naclbox/keysource.go
```diff
@@ -0,0 +1,219 @@
...
+
+// Decrypt decrypts the EncryptedKey field with NACL BOX private key and returns the result.
+func (key *MasterKey) Decrypt() ([]byte, error) {
+	// All we know at this stage is the public key used to encrypt the data key.
+	// We need to find the corresponding private key to perform the decryption.
+	// As a convention, we require that the private key is located under
+	// $HOME/.sops/naclbox/<sha256(pubkey)>.key
+	// where <sha256(pubkey)> is the sha256 hash of the raw public key
+	h := sha256.Sum256(key.pub[:])
+	path := fmt.Sprintf("%s/.sops/naclbox/%x.key", os.Getenv("HOME"), h[:])
```
autrilla: I'd split this into several functions to make it a bit easier to understand. There seems to be a few steps going on, e.g. loading the keys, validating them (base64 decoding and such), and actually decrypting

jvehent:
@autrilla Quick question about the protobuf of keyservice: I noticed it hasn't been updated in a while, and when I added the new key type and regenerated the pb.go file, it changed it dramatically. In particular, `Awsprofile` was gone. This seems to break tests in CI
`keyservice/server_test.go:60:5: unknown field 'AwsProfile' in struct literal of type KmsKey (but does have Awsprofile)`

Do you know what's going on there? Was it manually updated at some point?

autrilla:
@jvehent I think it's kind of the opposite? Looks like the generated Go code was checked in, but the updated proto definition was not. 4 is the tag previously used for that field, and it's the one you're adding in this PR, so we're good on that side. If you make the name: 

```diff
-string awsprofile = 4;
+string aws_profile = 4;
```

It'll also go back to the old generated code name, which I think we should keep.

Mic92:
Hopefully this will create a lot less headaches than gnupg's over-complicated design.

naclbox/keysource.go
Mic92: Please also add an environment variable that allows to relocate they key to a different directory than `$HOME`.

jvehent:
deprecated in favor of age

PR #573, opened 2019-11-15
Fix failing Vault test; update aws-sdk-go to v1.23.13

### LINKED ISSUE #572: Need updated aws-sdk-go for EKS IAM Role for Service Accounts
Per https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html the minimum `aws-sdk-go` version to support the newer EKS IRSA feature (which allows pods to assume an IAM role with a Kubernetes Service Account) is `1.23.13`.

The current version in `go.mod` is `1.21.6`. Is there any reason we can't update to at least `1.23.13`? I don't think any breaking changes were introduced between those versions.

The primary use case for this is using SOPS inside a Kubernetes pod where it needs to assume the correct pod role. Newer versions of the `aws-sdk-go` support this natively, but the old version doesn't support it at all, thus making it impossible for us to properly use roles with SOPS in our Kubernetes cluster.

jessebye:
Resolves #572

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ go: 1.12
 go_import_path: go.mozilla.org/sops/
 
 env:
-  - VAULT_VERSION=1.1.3 VAULT_TOKEN=root VAULT_ADDR='https://127.0.0.1:8200'
+  - VAULT_VERSION=1.1.3 VAULT_TOKEN=root VAULT_ADDR='http://127.0.0.1:8200'
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ go: 1.12
 go_import_path: go.mozilla.org/sops/
 
 env:
-  - VAULT_VERSION=1.1.3 VAULT_TOKEN=root VAULT_ADDR='http://127.0.0.1:8200'
+  - VAULT_VERSION=1.1.3 VAULT_TOKEN=root VAULT_ADDR='https://127.0.0.1:8200'
 
 addons:
   apt:
--- a/README.rst
+++ b/README.rst
@@ -996,14 +996,14 @@ Below is an example of publishing to Vault (using token auth with a local dev in
 .. code:: bash
 
    $ export VAULT_TOKEN=...
-   $ export VAULT_ADDR='http://127.0.0.1:8200'
+   $ export VAULT_ADDR='https://127.0.0.1:8200'
    $ sops -d vault/test.yaml
    example_string: bar
    example_number: 42
    example_map:
        key: value
    $ sops publish vault/test.yaml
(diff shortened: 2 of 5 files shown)


autrilla:
```
--- FAIL: TestLoadConfigFileWithVaultDestinationRules (0.00s)
    config_test.go:326: 
        	Error Trace:	config_test.go:326
        	Error:      	Not equal: 
        	            	expected: "https://127.0.0.1:8200/v1/secret/data/foobar/barfoo"
        	            	actual  : "http://127.0.0.1:8200/v1/secret/data/foobar/barfoo"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-https://127.0.0.1:8200/v1/secret/data/foobar/barfoo
        	            	+http://127.0.0.1:8200/v1/secret/data/foobar/barfoo
        	Test:       	TestLoadConfigFileWithVaultDestinationRules
    config_test.go:330: 
        	Error Trace:	config_test.go:330
        	Error:      	Not equal: 
        	            	expected: "https://127.0.0.1:8200/v1/kv/barfoo/barfoo"
        	            	actual  : "http://127.0.0.1:8200/v1/kv/barfoo/barfoo"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-https://127.0.0.1:8200/v1/kv/barfoo/barfoo
        	            	+http://127.0.0.1:8200/v1/kv/barfoo/barfoo
        	Test:       	TestLoadConfigFileWithVaultDestinationRules
FAIL
```

heh.

jessebye:
@autrilla darn. So something between how the tests are run in CI and how they are run locally is not consistent.

autrilla:
Yes, I think I figured out what! CI sets `export VAULT_ADDR='http://127.0.0.1:8200'`, which the Vault SDK then uses for the new connections, even in tests.

jessebye:
Aha, I see that now. I updated the URL in the CI config... does that sound OK?

autrilla:
> Aha, I see that now. I updated the URL in the CI config... does that sound OK?

I think that's going to fail the tests as the Vault tests server isn't set up for https. I think you should just revert any of those changes so CI still passes, and only make them locally to make sure your tests pass there. We'll figure out what to do to fix this inconsistency later.

jessebye:
Aaaaand it passes!

jessebye:
@autrilla what's the release cycle look like for a change like this? we're blocked by this currently, would be awesome even if we had to pull in a beta build of SOPS to keep moving.

autrilla:
> @autrilla what's the release cycle look like for a change like this? we're blocked by this currently, would be awesome even if we had to pull in a beta build of SOPS to keep moving.

We don't have a set release cycle. For now the quickest way is probably for you to build your own binary and distribute it. It might be good to set up CI to automatically publish assets for develop and master.

PR #576, opened 2019-11-16
Set up Github actions

autrilla:
My main reason to switch to Github actions is that it's much easier for us to push artifacts for every build (which is useful for people who don't want to wait for us to make a release). We could also create actions to automate creating releases by tagging. Overall, I think it's best to keep as much as we can in Github, and if we can stop using TravisCI and CircleCI, that's two fewer services we rely on.

This is missing the environment variable setup, but I wanted an initial review before that. We would need to pass it AWS credentials and `FUNCTIONAL_TEST_KMS_ARN`, as well as dockerhub credentials. Does Mozilla support builds from Github these days, or do we have to keep that in CircleCI?

Also fixes a bug with the Go tests, where they only worked because CI set a Vault env var for functional tests.

You can see these checks in action here: https://github.com/autrilla/sops/commit/f4763172a26ce2d03f5e66d6ba9fcfcbfca93320/checks

I suggest we keep TravisCI and CircleCI for a bit.


ajvb:
@autrilla Before I review this, I'm going to wait for the internal Mozilla sec audit of Github Actions to be completed (https://bugzilla.mozilla.org/show_bug.cgi?id=1594423). As well, as of now Github Actions are disabled within the `mozilla` Github org.

autrilla:
Oh woops, I didn’t realize Mozilla didn’t use Github actions yet. My bad!

autrilla:
We're unlikely to do this due to security considerations, AIUI, so closing.

PR #597, opened 2019-12-19
Feature/add php

sushant-115:
Add logic to encrypt and decrypt the simple php configuration files

### DIFF
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -18,6 +18,7 @@ import (
 	"go.mozilla.org/sops/v3/stores/dotenv"
 	"go.mozilla.org/sops/v3/stores/ini"
 	"go.mozilla.org/sops/v3/stores/json"
+	"go.mozilla.org/sops/v3/stores/php"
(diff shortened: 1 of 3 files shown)


autrilla:
This looks like a verbatim copy of the json store. I’m not sure what the
point of this is. It’s going to have the same effect as treating it as a
binary file, which is the current behavior.

sushant-115:
Yes, I just copied the json store and made the changes to fulfil my requirements. The binary store (or any other store) was not able to partially encrypt the php file with --encrypted-regex, with this we can partially encrypt the php files.

kanapuli:
@sushant-115  Can you describe what was the exact issue that you faced with php file ? Can you give us an example and ways to reproduce ?

autrilla:
The changes you've made seem pretty arbitrary. Feel free to expand on what issues you faced as @kanapuli asked, but either way, I think PHP support is unlikely to ever be added to SOPS.

We want to support encryption of _data_ formats. PHP is certainly not a data format. It is not well-suited to SOPS model of the world at all.

PR #591, opened 2019-12-10
Update the PGP Key for all Examples

apeschel:
Commit e9b9f7aeef15870454c8b788e87964efd62ec4c5 generated new PGP keys
for this repository, but failed to update the keys used by the examples.
As a result, the documentation for testing with the dev pgp key does not
work.

This commit rekeys the examples using the newly generated examples,
which allows the testing to work again.

PR #599, opened 2019-12-23
Update example files to use latest gpg key

pysysops:
Hey 👋 

I came across an issue when trying to decode the example files using the dev gpg keys. It appears that the files weren't updated with the new key introduced in #512 . I grabbed the old keys, updated `.sops.yaml` with one of the new keys and ran a `sops updatekeys -y example.<<ext>>` to fix the issue.


autrilla:
Thanks for the PR! This was fixed on #591 in the develop branch. @ajvb, could we cherry-pick that into master so people aren't confused?

pysysops:
Ah, that's great news 👍 - I didn't check the develop branch.

PR #607, opened 2020-01-15
Added version to docker tag #542

eddiewebb:
This change accesses the environment variable CIRCLE_TAG to address #542 

Changes:

1. Break docker push into its own job, runs only on tag creation
2. Use machine executor for docker push job (a little faster since it avoids shipping docker context around)
3. use the CIRCLE_TAG variable provided on tag builds

Options:

1. We can merge jobs back AND/OR
2. We can still publish SHA on master and just AMJOR/MINOR/PATCH, etc on tags

I ran in my own circle spae to test tagging.  I obviously dont have your docker creds, so the push fails, but the tagging works.

Run:
``` bash
#!/bin/bash -eo pipefail
docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"
# Update all semvers affected
echo "Tag mozilla/sops:$MAJOR"
docker tag mozilla/sops "mozilla/sops:$MAJOR"
echo "Tag mozilla/sops:$MAJOR.$MINOR"
docker tag mozilla/sops "mozilla/sops:$MAJOR.$MINOR"
echo "Tag mozilla/sops:$MAJOR.$MINOR.$PATCH"
docker tag mozilla/sops "mozilla/sops:$MAJOR.$MINOR.$PATCH"
docker images
```

Output
```
Tag mozilla/sops:0
Tag mozilla/sops:0.0
Tag mozilla/sops:0.0.5
REPOSITORY          TAG                                        IMAGE ID            CREATED                  SIZE
mozilla/sops        0                                          c533f52a658b        Less than a second ago   2.3GB
mozilla/sops        0.0                                        c533f52a658b        Less than a second ago   2.3GB
mozilla/sops        0.0.5                                      c533f52a658b        Less than a second ago   2.3GB
mozilla/sops        57f8a1fef66eb850dff6e43f06b1def877cff6af   c533f52a658b        Less than a second ago   2.3GB
mozilla/sops        latest                                     c533f52a658b        Less than a second ago   2.3GB
golang              1.12                                       8639ad3fc384        6 days ago               810MB
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,29 +74,29 @@ jobs:
           name: Tag & Push containers
           command: |
             #latest
-            src/go.mozilla.org/sops/bin/ci/deploy_dockerhub.sh "latest"
+            bin/ci/deploy_dockerhub.sh "latest"
 
             # by sha
             echo "Tag and push mozilla/sops:$CIRCLE_SHA1"
             docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"
-            src/go.mozilla.org/sops/bin/ci/deploy_dockerhub.sh "$CIRCLE_SHA1"
+            bin/ci/deploy_dockerhub.sh "$CIRCLE_SHA1"
 
             # by semver
             # v1.2.3
             if [ ! -z $PATCH ];then
               echo "Tag and Push mozilla/sops:v$MAJOR.$MINOR.$PATCH"
               docker tag mozilla/sops "mozilla/sops:v$MAJOR.$MINOR.$PATCH"
-              src/go.mozilla.org/sops/bin/ci/deploy_dockerhub.sh "v$MAJOR.$MINOR.$PATCH"
+              bin/ci/deploy_dockerhub.sh "v$MAJOR.$MINOR.$PATCH"
             fi
             # v1.2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -12,10 +24,65 @@ jobs:
...
+
+  push:
+    machine: true
+    resource_class: large
+    steps:
+      - checkout
+      - run: 
+          name: semver check
+          command: |
+            curl "https://raw.githubusercontent.com/fsaintjacques/semver-tool/d66113a6803dbffca2a3830e207a6795a8516cd1/src/semver" > semver.bash
```
ajvb: Pulling in an external tool for this feels a bit heavy weight. Can this be changed to just using `cut`?

eddiewebb: I agree with the concern, I locked into a hash to avoid surprises but I understand the reluctance.

1. We could commit the referenced bash script into source.  
1. We could re-write it in bash ourselves (dislike this)
1. if this project won't ever use classififers and just 1.2.3. I could do some sed manipulation.

eddiewebb: Updated to use cut, push step will ignore missing minor/patch versions, but requires major version.

PR #609, opened 2020-01-18
Alpine-based docker image

ikedam:
Replaces #604 .

I want alpine-based docker images for sops, as it's small and easy to use:

* The current image uses 2 GB+:
    ```
    $ docker images mozilla/sops
    REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
    mozilla/sops        latest              6a8e325d2d11        5 weeks ago         2.3GB
    ```
* The alpine-based image uses only 65MB:
    ```
    $ docker build -t sops -f Dockerfile.alpine .
    ...
    $ docker images sops
    REPOSITORY          TAG                 IMAGE ID            CREATED              SIZE
    sops                latest              c4f6175388ce        About a minute ago   64.4MB
    ```

And this pull request also changes `.circleci/config.yml` to deploy alpine based images as `mozilla/sops:3.5.0-alpine` and `mozilla/sops:alpine`.
I believe we don't need alpine images for every commits and only images for tags are enough useful.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,16 +56,20 @@ jobs:
           name: Build containers
           command: |
             docker build -t mozilla/sops .
+            docker build -f Dockerfile.alpine -t mozilla/sops:alpine .
       - run:
           name: Tag & Push containers
           command: |
             #latest
             bin/ci/deploy_dockerhub.sh "latest"
+            bin/ci/deploy_dockerhub.sh "alpine"
 
             # by sha
             echo "Tag and push mozilla/sops:$CIRCLE_SHA1"
             docker tag mozilla/sops "mozilla/sops:$CIRCLE_SHA1"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -12,10 +12,25 @@ jobs:
...
+                ${GOPATH}/src/go.mozilla.org/sops/bin/ci/deploy_dockerhub.sh "alpine"
+            fi
+workflows:
+  version: 2
+  build:
+    jobs:
+      - build:
+          filters:
+            tags:
+              only: /.*/
```
ikedam: I need to use workflow to have circleci run also for tags.

ikedam:
@autrilla Thank you, I rebased this to the latest develop and reworked based on #607.

autrilla:
Thanks for working on this :)

ikedam:
Thanks!

PR #615, opened 2020-01-26
Support compact-json format

ikedam:
"compact-json" is json format without insignificant white space letters:

```
/workspace # sops -d --output-type json test.yaml
{
        "key1": "value1",
        "key2": {
                "nestedkey1": "value1",
                "nestedkey2": "value2"
        },
        "key3": [
                "value1",
                "value2",
                "value3"
        ]
}/workspace #
/workspace # sops -d --output-type compact-json test.yaml
{"key1":"value1","key2":{"nestedkey1":"value1","nestedkey2":"value2"},"key3":["value1","value2","value3"]}/workspace #
```

It would be nice if sops supports output with that format as values without line breaks are often useful to use in command parameters and values for environment variables.

### DIFF
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -51,7 +51,15 @@ func newIniStore() Store {
 }
 
 func newJsonStore() Store {
-	return &json.Store{}
+	return &json.Store{
+		Compact: false,
+	}
+}
+
+func newCompactJsonStore() Store {
+	return &json.Store{
+		Compact: true,
+	}
 }
 
 func newYamlStore() Store {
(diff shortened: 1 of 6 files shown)


ajvb:
At the current moment, I am against merging this. I appreciate the PR @ikedam and don't want your work to be dismissed, but I am unconvinced at the moment that the value this adds is worth the additional options, code, and maintenance cost.

I do agree with @autrilla that the maintenance cost is small, but it adds up over time. In recent refactors and feature additions, I have noticed the size that sops is currently at combined with a lack of full coverage test suite, and it has caused problems.

With that in mind, I am against merging features that don't provide a lot of value until we can improve the quality of our test suite to a state where small features like this won't break after a year or so.

ikedam:
OK. Thanks for reviewing!
I’ll recreate this request when test codes of sops get improved.

PR #616, opened 2020-01-28
update aws-sdk-go dependency

joshua-rutherford:
There are numerous updates to the aws-sdk-go libraries and this upgrades those dependencies only.


autrilla:
Is there any change in particular you care about?

joshua-rutherford:
The real driver here is that the regions supported by the AWS SDK are enumerated in the source code.  As a result, to use new ore modified regions you must upgrade the SDK.

ajvb:
@joshua-rutherford I mistakenly merged this into master, but this PR needs to be made against `develop`. Would you mind recreating it against the `develop` branch and I can pull it right in?

scjudd:
Also, can we have a more descriptive commit message than "f", maybe one that captures some of the context given here?

PR #618, opened 2020-01-30
Fix exec-env with multiline env vars

scjudd:
When using the `exec-env` subcommand, SOPS does not unescape newline
characters. Causing the following behavior:

```console
$ echo 'FOO=foo\nbar\nbaz' > plaintext.env
$ sops -e --output ciphertext.env plaintext.env
$ sops exec-env ciphertext.env 'env | grep FOO | xxd'
00000000: 464f 4f3d 666f 6f5c 6e62 6172 5c6e 6261  FOO=foo\nbar\nba
00000010: 7a0a                                     z.
```

This commit unescapes the newlines when constructing the environment to
be passed to the given command, producing the expected output:

```console
$ sops exec-env ciphertext.env 'env | grep -A2 FOO | xxd'
00000000: 464f 4f3d 666f 6f0a 6261 720a 6261 7a0a  FOO=foo.bar.baz.
```

### DIFF
--- a/cmd/sops/subcommand/exec/exec.go
+++ b/cmd/sops/subcommand/exec/exec.go
@@ -91,6 +91,7 @@ func ExecWithEnv(opts ExecOpts) {
 		if line[0] == '#' {
 			continue
 		}
+		line := bytes.Replace(line, []byte("\\n"), []byte("\n"), -1)
 		env = append(env, string(line))
 	}
 


scjudd:
For context:
* Dotenv files are newline-delimited and must therefore escape newlines in variable values.
* Environment variables in a running process are null-byte-delimited and can therefore contain newlines.

The expectation when a person encodes a newline in a dotenv file is that it translates to an actual newline when injected into a process' environment.

scjudd:
Actually, this seems to break in the "a person wished to include a literal `\n`" case: instead of `\\n` becoming `\n` (`0x5c6e`), it becomes `0x5c0a` (a backslash and a newline).

I'm going to close this in pursuit of a solution that addresses the above.

PR #622, opened 2020-02-04
Add standard newline/quoting behavior to dotenv store

scjudd:
Rationale
=========

The dotenv store as it exists right now performs splitting on newlines to determine where a new key-value pair or comment begins. This works remarkably well, up until you need to handle values that contain newlines.

While I couldn't find an offical dotenv file format spec, I sampled a number of open-source dotenv parsers and it seems that they typically apply the following rules:

Newline handling:

* If a value is unquoted and contains a literal `\n` (`0x5c6e`), it is interpretted literally and NOT converted to an actual newline (`0x0a`).

* If a value is single-quoted and contains a literal `\n` (`0x5c6e`), it is interpretted literally and NOT converted to an actual newline (`0x0a`).

* If a value is double-quoted and contains a literal `\n` (`0x5c6e`), it is converted to an actual newline (`0x0a`).

* If a value is either single- or double-quoted, it may contain an actual newline (`0x0a`).

Whitespace trimming:

* If a value is unquoted and contains any leading or trailing whitespace, it is trimmed.

* If a value is either single- or double-quoted and contains any leading or trailing whitespace, it is left untrimmed.

Quotation handling:

* Because quotations around values have special meaning, they are interpretted and are not included in the parsed value. Literal quotes may be included within a quoted string either by escaping them or using the opposite quotation mark.

Because single- and double-quoted values may contain actual newlines, we cannot split our input data on newlines as this may be in the middle of a quoted value. This, along with the other rules around handling quoted values, prompted me to try and implement a more robust parsing solution. This commit is my first stab at that.

Special Considerations
======================

This is _not_ a backwards-compatible change:

* The `dotenv` files produced by this version of SOPS _cannot_ be read by an earlier version.

* The `dotenv` files produced by an earlier version of SOPS _can_ be read by this version, with the understanding that the semantics around quotations and newlines have changed.

Examples
========

The below examples show how double-quoted values are passed to the running environment:

```console
$ echo 'FOO="foo\\nbar\\nbaz"' > plaintext.env
$ sops -e --output ciphertext.env plaintext.env
$ sops exec-env ciphertext.env 'env | grep FOO | xxd'
00000000: 464f 4f3d 666f 6f5c 6e62 6172 5c6e 6261  FOO=foo\nbar\nba
00000010: 7a0a                                     z.
```

```console
$ echo 'FOO="foo\nbar\nbaz"' > plaintext.env
$ sops -e --output ciphertext.env plaintext.env
$ sops exec-env ciphertext.env 'env | grep -A2 FOO | xxd'
00000000: 464f 4f3d 666f 6f0a 6261 720a 6261 7a0a  FOO=foo.bar.baz.
```

### DIFF
--- a/cmd/sops/subcommand/exec/exec.go
+++ b/cmd/sops/subcommand/exec/exec.go
@@ -1,13 +1,14 @@
 package exec
 
 import (
-	"bytes"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"runtime"
 	"strings"
 
 	"go.mozilla.org/sops/v3/logging"
+	"go.mozilla.org/sops/v3/stores/dotenv"
 
 	"github.com/sirupsen/logrus"
 )
@@ -83,15 +84,18 @@ func ExecWithEnv(opts ExecOpts) {
 	}
 
 	env := os.Environ()
-	lines := bytes.Split(opts.Plaintext, []byte("\n"))
-	for _, line := range lines {
-		if len(line) == 0 {
+	store := dotenv.Store{}
+
+	branches, err := store.LoadPlainFile(opts.Plaintext)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for _, item := range branches[0] {
+		if item.Value == nil {
 			continue
 		}
-		if line[0] == '#' {
-			continue
-		}
-		env = append(env, string(line))
+		env = append(env, fmt.Sprintf("%s=%s", item.Key, item.Value))
 	}
 
 	cmd := BuildCommand(opts.Command)
--- a/stores/dotenv/parser.go
+++ b/stores/dotenv/parser.go
@@ -1,0 +1,311 @@
+package dotenv
+
+// The dotenv parser is designed around the following rules:
+//
+// Comments:
+//
+// * Comments may be written by starting a line with the `#` character.
+//   End-of-line comments are not currently supported, as there is no way to
+//   encode a comment's position in a `sops.TreeItem`.
+//
+// Newline handling:
+//
+// * If a value is unquoted or single-quoted and contains the character
+//   sequence `\n` (`0x5c6e`), it IS NOT decoded to a line feed (`0x0a`).
+//
+// * If a value is double-quoted and contains the character sequence `\n`
+//   (`0x5c6e`), it IS decoded to a line feed (`0x0a`).
+//
+// Whitespace trimming:
+//
+// * For comments, the whitespace immediately after the `#` character and any
+//   trailing whitespace is trimmed.
+//
+// * If a value is unquoted and contains any leading or trailing whitespace, it
+//   is trimmed.
+//
+// * If a value is either single- or double-quoted and contains any leading or
+//   trailing whitespace, it is left untrimmed.
+//
+// Quotation handling:
+//
+// * If a value is surrounded by single- or double-quotes, the quotation marks
+//   are interpreted and not included in the value.
+//
+// * Any number of single-quote characters may appear in a double-quoted
+//   value, or within a single-quoted value if they are escaped (i.e.,
+//   `'foo\'bar'`).
+//
+// * Any number of double-quote characters may appear in a single-quoted
+//   value, or within a double-quoted value if they are escaped (i.e.,
+//   `"foo\"bar"`).
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+
+	"go.mozilla.org/sops/v3"
+)
+
+var KeyRegexp = regexp.MustCompile(`^[A-Za-z_]+[A-Za-z0-9_]*$`)
+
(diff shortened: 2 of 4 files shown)


scjudd:
Alright, this is now to a point where I'm content with the quality of it and I promise to not push anymore unless someone requests changes. :smile:

scjudd:
Okay so I lied. I added whitespace trimming to comments.

cmd/sops/subcommand/exec/exec.go
```diff
@@ -83,15 +84,18 @@ func ExecWithEnv(opts ExecOpts) {
...
-	for _, line := range lines {
-		if len(line) == 0 {
-			continue
-		}
-		if line[0] == '#' {
+	store := dotenv.Store{}
+
+	branches, err := store.LoadPlainFile(opts.Plaintext)
+	if err != nil {
+		log.Fatal(err)
```
ajvb: Could we include a more descriptive error message alongside the error value?

stores/dotenv/parser.go
```diff
@@ -0,0 +1,271 @@
+package dotenv
+
```
ajvb: The write-up in the top-level comment in this PR is awesome, could a version of that be placed as a comment here so that future generations understand how the parser is supposed to work and why?

stores/dotenv/parser.go
```diff
@@ -0,0 +1,271 @@
...
+			break
+		}
+
+		builder.WriteByte(b)
+	}
+
+	key = builder.String()
+
+	if !KeyRegexp.MatchString(key) {
+		err = fmt.Errorf("invalid dotenv key: %q", key)
```
ajvb: I know the original version didn't include one, but since you're in here, could you include a test case with an "invalid dotenv key"?

scjudd: Absolutely—great idea.

stores/dotenv/parser.go
scjudd: For sure. :+1:

cmd/sops/subcommand/exec/exec.go
scjudd: I agree in spirit, but that seems to deviate from the norm established by the other functions in this file. Any suggestions?

ajvb: Hmm, yeah I see your point. In that case, let's just leave it as-is for now.

scjudd: Sounds good!

scjudd:
Hey @autrilla, I can't remember all of the different parsers I looked at, but the rules defined for https://github.com/motdotla/dotenv#rules are what I based a lot of this on.

As far as breaking this out into a separate repo, I'm not opposed to that, but right now it is very purpose-built for SOPS. If you have any suggestions on what changes you'd like to see, I'm down to push this forward a little further. _Something_ along these lines needs to land to fix the issue for newlines and exec-env.

scjudd:
Hey @autrilla, just checking in on this. Is there anything I can do to help get this landed?

autrilla:
Sorry, I forgot about this. LGTM, thanks for the patch!

mltsy:
Turns out this is a pseudo-breaking change for people using the dotenv format (including me), because the previous version of sops doesn't know what to do with the new quoted values (so in our build pipeline, which uses sops 3.5.0, it's unable to parse the new format coming from our local environments as people start upgrading to 3.6.0)

I'm not sure if there's a good place to put these notes.  Ideally they wouldn't have been pointed out more prominently in the changelog - maybe I'll put in an issue to request a flag for --omit-dotenv-quotes or something in case others are encountering this issue...

patricknelson:
Is this at all related to my issue at #784?

In my use case, I was loading environment variables which contain multiple lines (stored in encrypted `JSON` and `YAML`). It appears to be interpreted properly when viewing/editing directly within sops itself but not when loaded into environment variables via `exec-env`.

mltsy:
Don't think it's related to this (this was only present in 3.5.0) - except that it's probably related to the dotenv encoder.  Commented on your issue.

patricknelson:
I think you're right @mltsy, I see [the operative line](https://github.com/mozilla/sops/blob/master/stores/dotenv/store.go#L122) that affects my use case were reverted in 5d32d9a. However, I will say that investigating my issue which is related via the shared use of the `dotenv` output formatting, I found a possible issue with how the `dotenv` output format encodes/decodes environment variables (i.e. it'll add `\n`'s even though the variable in it's *raw* form may already contain `\n's`, meaning it is mixing unencoded content with encoded content and potentially mangling things). I have more details in my other comment. 

Basically, I recommend using something like [joho/godotenv](https://github.com/joho/godotenv) to encode/decode `dotenv` values and then separating **that** process from the process used for calling `os.StartProcess` which _may_ accept an entirely different environment variable encoding (something I don't know, since I'm totally new to golang).

PR #507, opened 2019-08-07
Adding Hashicorp Vault KMS

gitirabassi:

PR #623, opened 2020-02-05
Followup: Adding Hashicorp Vault KMS

ldue:
Required improvements for [PR 507](https://github.com/mozilla/sops/pull/507).


ldue:
I‘ll try to add some more tests, but please feel free to inspire me for that 🙂

ldue:
I noticed during testing that the Readme maybe should contain some more notes what to do for the functional testing (e.g. enable the vault kv store).  @gitirabassi  has done this nicely for this PR, but maybe we should do this reworks in a separate PR and maybe also decide with that one, if the dependency for vault while unit testing should be kept.

ajvb:
@ldue I completely agree on the lack of docs for functional tests. If you're willing to create an issue with any notes on what specifically was hard to find I'd be happy to work on it.

ldue:
@ajvb yeah sure, I think I will get to this in the next week.
@autrilla thanks for the review! So then it's ready to merge?

autrilla:
Ideally I’d want @ajvb to give it a look as well before merging. I’d also really appreciate the notes on what kind of tests and docs you find lacking!

README.rst
```diff
@@ -290,6 +290,66 @@ And decrypt it using::
...
+	Success! Data written to: sops/keys/thirdkey
+
+	$ sops --hc-vault $VAULT_ADDR/v1/sops/keys/firstkey vault_example.yml
+
+	$ cat <<EOF >.sops.yaml
+	creation_rules:
+		- path_regex: \.dev\.yaml$
+		hc_vault_uris: "$VAULT_ADDR/v1/sops/keys/secondkey"
+		- \.prod\.yaml$
+		hc_vault_uris: "$VAULT_ADDR/v1/sops/keys/thirdkey"
```
mmorev: Seems should look like this
```suggestion
	creation_rules:
		- path_regex: \.dev\.yaml$
		  hc_vault_uris: "$VAULT_ADDR/v1/sops/keys/secondkey"
		- path_regex: \.prod\.yaml$
		  hc_vault_uris: "$VAULT_ADDR/v1/sops/keys/thirdkey"
```

hcvault/keysource.go
```diff
@@ -0,0 +1,274 @@
...
+		return err
+	}
+	return nil
+}
+
+// ToMap converts the MasterKey to a map for serialization purposes
+func (key MasterKey) ToMap() map[string]interface{} {
+	out := make(map[string]interface{})
+	out["vault_address"] = key.VaultAddress
+	out["keyname"] = key.KeyName
```
mmorev: just for uniformity with config snake case variant
```suggestion
	out["key_name"] = key.KeyName
```

hcvault/keysource_test.go
```diff
@@ -0,0 +1,168 @@
...
+		CreationDate: time.Date(2016, time.October, 31, 10, 0, 0, 0, time.UTC),
+		VaultAddress: "http://127.0.0.1:8200",
+		BackendPath:  "foo",
+		KeyName:      "bar",
+		EncryptedKey: "this is encrypted",
+	}
+	assert.Equal(t, map[string]interface{}{
+		"vault_address": "http://127.0.0.1:8200",
+		"backend_path":  "foo",
+		"keyname":       "bar",
```
mmorev:
```suggestion
		"key_name":       "bar",
```

config/config_test.go
```diff
@@ -195,12 +209,14 @@ func TestLoadConfigFile(t *testing.T) {
...
 				PGP:       "2",
 				GCPKMS:    "3",
+				Vault:     "http://4:8200/v1/4/keys/4",
 			},
 			{
 				PathRegex: "",
 				KMS:       "foo",
 				PGP:       "bar",
 				GCPKMS:    "baz",
+				Vault:     "http://127.0.1.1/v1/baz/keys/baz",
```
mmorev:
```suggestion
				Vault: []vaultKey{
					{
						VaultAddress: "http://127.0.1.1",
						BackendPath:  "baz/keys",
						KeyName:      "baz",
					},
				},
```

config/config.go
```diff
@@ -174,6 +186,13 @@ func getKeyGroupsFromCreationRule(cRule *creationRule, kmsEncryptionContext map[
 		for _, k := range azureKeys {
 			keyGroup = append(keyGroup, k)
 		}
+		vaultKeys, err := hcvault.NewMasterKeysFromURIs(cRule.Vault)
```
mmorev:
```suggestion
		for _, k := range cRule.Vault {
			keyGroup = append(keyGroup, hcvault.NewMasterKey(k.VaultAddress, k.BackendPath, k.KeyName))
		}
		vaultKeys, err := hcvault.NewMasterKeysFromURIs(cRule.VaultUri)
```

vault_example.yml
```diff
@@ -0,0 +1,48 @@
+hello: ENC[AES256_GCM,data:1DWfYHFHwafRZSOSuMy8rHvF3Da+9NFjcwPdFPmepZl2CKHL76L38CC5H2NzIQ==,iv:AhSDXStvMXp6ulPwFWW8+izJo6l5Zem9Xh6+I7/GKys=,tag:+nkbrSzLFm1VXiggb5GknA==,type:str]
```
mmorev: What is this file here for?

config/config_test.go
```diff
@@ -195,12 +209,14 @@ func TestLoadConfigFile(t *testing.T) {
 				KMS:       "1",
 				PGP:       "2",
 				GCPKMS:    "3",
+				Vault:     "http://4:8200/v1/4/keys/4",
```
mmorev:
```suggestion
				VaultUri:  "http://4:8200/v1/4/keys/4",
```

vault_example.yml
autrilla: As an example, for consistency with all the other _example.yml files we have for all stores.

autrilla: Actually, my bad, this is not a store, hah. Yeah, I don't think we need this example file. The docs should be enough.

README.rst
autrilla: Good catch!

autrilla:
> I think it would be more clear and definite to use separate parameters in sops.yml, just like these you offer in `key_groups.hc_vault`.
> 
> However, single-string URI is still very convenient for command line options like `--add-hc-vault`.

I kind of disagree here, I think consistency is best and this is pretty similar to what we do for AWS: we take an ARN. We do the same for GCP, and probably Azure (but I haven't checked). The URI is a "standard" way to refer to things in Vault, no?

> Also I would suggest rename `hc-vault` to `vault-transit`, because it is common name for this key management service type.

I completely agree with this. Probably even `hc-vault-transit`. Maybe we'll need different backends that are also provided by Hashicorp Vault in the future, and the concern about a different product also named Vault (by another company) still applies.

Thanks for giving this a look too @mmorev!

config/config.go
```diff
@@ -110,6 +118,7 @@ type creationRule struct {
 	PGP               string
 	GCPKMS            string     `yaml:"gcp_kms"`
 	AzureKeyVault     string     `yaml:"azure_keyvault"`
+	Vault             string     `yaml:"hc_vault_uris"`
```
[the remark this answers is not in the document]
autrilla: As mentioned elsewhere, I think we should keep the URI, but actually, please include Hashicorp and Transit in the name somehow. `HcVaultTransit`, for example. I'm open to alternatives, but let's keep it consistent everywhere.

mmorev:
> I kind of disagree .here, I think consistency is best and this is pretty similar to what we do for AWS: we take an ARN.
ARN is much more definite because it uses colons as field separator, so if any required field is missing or incorrect, we can easily find out, which exactly. Even we can safely skip unnecessary field.
URIs are much more ambiguous. Vault can be behind reverse proxy and/or WAF, or path to secret store can contain slashes (can not as of now, but other stores can) - voila, URI is not consistent.
Also mentioned `/v1/store_path/keys/key_name` is not exact URI path used when calling Vault API. It can be `v1/store_path/encrypt/key_name` for encrypting data, for example.
So I suggested to implement both methods - URI and separate fields.

Also @ldue please rename backend_path to secret_path or something else in code and parameters, because backend in Vault terms is actually the storage for all configured vault secrets,.

> Probably even `hc-vault-transit`.
`hc-` prefix wasn't used when Vault publish was implemented, why should here?

config/config.go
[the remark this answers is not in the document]
ldue: I've adapted you suggestions thanks! I also changed the plural `hc_vault_uris` to singular `hc_vault_uri`

ldue:
Thanks for your great feedback! Sorry it took me so long to get back to it.

@autrilla i also aggree with @mmorev  it makes sense to implement both ways, uri based and "object" based config. I do not see to much disadvantage.

@mmorev I renamed `BackendPath` to `EnginePath` as I find secrets somehow ambiguous and as the vault term is "secrets engine" i think it fits.

I think the `hc-` prefix came exactly from some confusion around AzureKeyVault, the general concept of a vault for keys and (hashicorp) Vault, when implementing the publish feature. 
I think it should be refactored for the publish feature, but not with this PR.
You can check the original [PR 507](https://github.com/mozilla/sops/pull/507) for that.

autrilla:
> ARN is much more definite because it uses colons as field separator, so if any required field is missing or incorrect, we can easily find out, which exactly. Even we can safely skip unnecessary field.
> URIs are much more ambiguous. Vault can be behind reverse proxy and/or WAF, or path to secret store can contain slashes (can not as of now, but other stores can) - voila, URI is not consistent.
> Also mentioned `/v1/store_path/keys/key_name` is not exact URI path used when calling Vault API. It can be `v1/store_path/encrypt/key_name` for encrypting data, for example.

It is in no way different than an ARN. It's a URI, but not _any_ URI. It must match a very specific format. Just like an ARN.

The issue with reverse proxies is irrelevant here. The URI is just a way to specify the resource. It does not imply we're going to make HTTP requests to that URI directly. In fact, we're parsing it and passing it off to the Vault library.

An argument against this that I would accept is that this "URI" is not really an exclusive identifier Vault uses for the resource. It's just the way to access it through the HTTP API, but there might be other ways to access it in the future. So maybe it does make sense to take explicitly structured data instead of this URI.

> So I had suggested to implement both methods - URI and separate fields.

This is absolutely what I do not want. I do not want two ways of doing the same thing. It makes the code more complicated and a pain to maintain. I don't care super strongly about what is done regarding URI vs more structured data in the config file, but let's do only one thing. The added value is tiny and the maintenance cost adds up very quickly. 

> `hc-` prefix wasn't used when Vault publish was implemented, why should here?

I think it's the right thing to do regardless of what we've done in the past. Clearly we've made some mistakes with naming. AWS KMS being named just "kms" is probably the biggest one.

ldue:
> An argument against this that I would accept is that this "URI" is not really an exclusive identifier Vault uses for the resource. It's just the way to access it through the HTTP API, but there might be other ways to access it in the future. So maybe it does make sense to take explicitly structured data instead of this URI.

I think in that case URI would be more consistent to the other providers.

> The added value is tiny and the maintenance cost adds up very quickly.

I get that, I do not have any experience with maintaining code for that kind of config.

So I'll remove the structured config and leave VaultURI in and then we can bring this to an end? 😄

cmd/sops/main.go
```diff
@@ -501,6 +523,11 @@ func main() {
 			Usage:  "comma separated list of Azure Key Vault URLs",
 			EnvVar: "SOPS_AZURE_KEYVAULT_URLS",
 		},
+		cli.StringFlag{
+			Name:   "hc-vault--transit",
```
vaibhavk:
```suggestion
			Name:   "hc-vault-transit",
```

autrilla:
> I think in that case URI would be more consistent to the other providers.
> I get that, I do not have any experience with maintaining code for that kind of config.
> So I'll remove the structured config and leave VaultURI in and then we can bring this to an end? 😄

Sounds good to me!

mmorev:
@autrilla IMO what really makes code more complicated and harder to read/maintain is a function with two unobvious regular expressions which is needed to parse an URI in unobvious way and then, anyway, store result as three different values in sops metadata.
Maybe, my suggestion doesnt add value, but it does add logic and readability to source code and to users configuration files.

Anyway, it's just a suggestion, you are maintainer, you are decision maker.

autrilla:
We wouldn't store the three values in SOPS metadata, we'd store just the URI.

Like I said, I'd be open to storing only the three values and not dealing with URIs at all. What I don't want is both things.

mmorev:
Oh, got it. @ldue What do you think about removing URI and regex parsing function and leaving just separate parameter set?

cmd/sops/main.go
```diff
@@ -961,12 +1012,21 @@ func keyGroups(c *cli.Context, file string) ([]sops.KeyGroup, error) {
 			azkvKeys = append(azkvKeys, k)
 		}
 	}
+	if c.String("hc-vault-transit") != "" {
+		hcVaultKeys, err := hcvault.NewMasterKeysFromURIs(c.String("vault"))
```
vaibhavk:
```suggestion
		hcVaultKeys, err := hcvault.NewMasterKeysFromURIs(c.String("hc-vault-transit"))
```

cmd/sops/main.go
```diff
@@ -961,12 +1012,21 @@ func keyGroups(c *cli.Context, file string) ([]sops.KeyGroup, error) {
...
+			hcVaultMkKeys = append(hcVaultMkKeys, k)
+		}
+	}
 	if c.String("pgp") != "" {
 		for _, k := range pgp.MasterKeysFromFingerprintString(c.String("pgp")) {
 			pgpKeys = append(pgpKeys, k)
 		}
 	}
-	if c.String("kms") == "" && c.String("pgp") == "" && c.String("gcp-kms") == "" && c.String("azure-kv") == "" {
+	if c.String("kms") == "" && c.String("pgp") == "" && c.String("gcp-kms") == "" && c.String("azure-kv") == "" && c.String("vault") == "" {
```
vaibhavk:
```suggestion
	if c.String("kms") == "" && c.String("pgp") == "" && c.String("gcp-kms") == "" && c.String("azure-kv") == "" && c.String("hc-vault-transit") == "" {
```

binlab:
Great work, thanks a lot!

jvehent:
@ajvb @autrilla can this be merged?

autrilla:
There's still some review comments that should be addressed, mainly only allowing one of specifying the Vault URI or specifying each value (hostname, key, etc) individually.

vnzongzna:
Would it be acceptable to make a new PR since we need vault feature in SOPS and this PR has been inactive for more than 2 months now.
cc @ldue @jvehent @autrilla @mmorev

autrilla:
> Would it be acceptable to make a new PR since we need vault feature in SOPS and this PR has been inactive for more than 2 months now.
> The new PR would be third iteration for the same feature but I'll be working on it continuously.
> cc @ldue @jvehent @autrilla @mmorev

Yes, that is acceptable as long as you keep authorship of the original commits. I think this PR itself started that way as well...

mykter:
This can be closed now that #655 has been merged

PR #627, opened 2020-02-10
Fixes #626 return exit code with exec-env and exec-file

### LINKED ISSUE #626: sops exec-file does not propagate exit codes
`sops exec-file` is a really cool feature, but I find it hard to use in CI pipelines because of one issue. Exit code of the executed command is not propagated, for example:
```bash
$ sops exec-file secret.enc.yaml 'exit 1'
$ echo $?
0
```

It would be great if the exit codes could be propagated or extracted somehow after running `sops exec-file` (and probably `exec-env` too).
Or maybe I am missing something?

alx13:
now returning exit code with exec-env and exec-file

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	osExec "os/exec"
 	"path/filepath"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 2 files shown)


alx13:
I found these commands extremely useful (exec-env with terraform and exec-file with helm),
but I don't see much point in having --background flag:

1. there is no way of saying if something went wrong
2. for exec-file it would just not work well because of race-condition of executable reading file and sops deleting it (`sops exec-file --background secrets.env 'sleep 5; cat {} > /tmp/xxx.env'; sleep 10; cat /tmp/xxx.env` — gives you empty file)

ajvb:
@alx13 re: `--background` - yeah, I completely agree. Created https://github.com/mozilla/sops/issues/629

ajvb:
Thanks @alx13 !

PR #630, opened 2020-02-11
Fix #629 : removing --background flag for exec-env and exec-file

### LINKED ISSUE #629: Consider deprecating/removing sops exec --background flag
Initially raised by @alx13 in https://github.com/mozilla/sops/pull/627#issuecomment-584328589

The `--background` flag for the `sops exec-env` and `sops exec-file` is not particularly useful and backgrounding commands should really be done using your shell environment.

alx13:


autrilla:
This is going to break people :( if we want to remove it, it should be done in two stages. First, deprecate and warn, and some time later, actually remove it.

PR #638, opened 2020-03-07
decrypt master key in parallel

SinisterMinister:
Currently, Sops tries to decrypt the master key by iterating over each one in series. In practice, this can lead to significant performance issues if the working key is at the end of the list. The example case we encountered was using Sops with Ansible and the upcoming Sops plugin for it. During execution, Ansible calls sops for every task to decrypt the variables. In our case, we used KMS and PGP for our keys, and when running on a machine that relied on PGP, it would take over 30 seconds to decrypt a file as it iterated through providers. This would make PGP based builds take hours where it would take minutes for KMS based builds as it would fail the KMS key look up thousands of times throughout the run.

This change makes the decryption efforts happen in parallel, returning the first to succeed, making the speed faster than KMS based builds without affecting the KMS based build times.

### DIFF
--- a/kms/keysource.go
+++ b/kms/keysource.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"go.mozilla.org/sops/v3/logging"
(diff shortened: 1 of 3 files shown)


kms/keysource.go
```diff
@@ -92,17 +96,26 @@ func (key *MasterKey) Decrypt() ([]byte, error) {
 		log.WithField("arn", key.Arn).Info("Decryption failed")
 		return nil, fmt.Errorf("Error base64-decoding encrypted data key: %s", err)
 	}
+
+	// Capture service locally to prevent data races
+	kmsSvcMtx.RLock()
+	svc := kmsSvc
```
SinisterMinister: These changes were necessary to deal with the concurrent access of the `kmsSvc` by the decrypt calls.

PR #641, opened 2020-03-09
Add support for decoding JSON arrays of arrays

### LINKED ISSUE #640: Support JSON arrays of arrays
Currently, JSON arrays of arrays cannot be parsed by sops, leading to an unmarshaling error. Given the following input, for example:

```json
{
	"ary": [[]]
}
```

sops will produce the following error:

```
[CMD]    ERRO[0094] Could not load tree, probably due to invalid syntax. Press a key to return to the editor, or Ctrl+C to exit.  error="Could not unmarshal input data: Expected JSON object key, got ] of type json.Delim instead"
```

The same input treated as YAML works, so the current workaround at least appears to be to use YAML.

nilium:
Add support for decoding JSON arrays of arrays by handling, during
slice decoding, when the next token is an array opening. This produces
nested []interface{} slices.

Closes #640.

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -79,6 +79,12 @@ func (store Store) sliceFromJSONDecoder(dec *json.Decoder) ([]interface{}, error
 				return slice, err
 			}
 			slice = append(slice, item)
+		} else if ok && delim.String() == "[" {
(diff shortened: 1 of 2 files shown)


nilium:
This was supposed to be against develop. Fixing that.

PR #647, opened 2020-03-30
Add a new YamlFile function to decrypt package that uses keyservices

### LINKED ISSUE #643: Support keyservice in decrypt package
I'm writing a tool that uses `sops` as a library via the `decrypt` package, and I would like to use that tool in a workspace with keygroups. Currently, `decrypt.File` does not make use of keyservices and thus decryption fails when more than one keygroup is required.

Is it just a matter of implementing https://github.com/mozilla/sops/blob/master/cmd/sops/decrypt.go#L22 in the decrypt package?

ni? @autrilla

jvehent:
fixes #643

This doesn't break any existing code, it's just a new function in the decrypt package that makes a few assumptions, like the presence of a keyservice unix socket at /tmp/sops.sock.

Perhaps we could call this WorkspaceYamlFile so it's clear it's specific to us?

### DIFF
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -11,8 +11,72 @@ import (
 
 	"go.mozilla.org/sops/v3/aes"
 	"go.mozilla.org/sops/v3/cmd/sops/common"
-	. "go.mozilla.org/sops/v3/cmd/sops/formats" // Re-export
+	"go.mozilla.org/sops/v3/cmd/sops/formats"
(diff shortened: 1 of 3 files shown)


jvehent:
Yeah, after letting this sit for a couple hours, I do think it's too specific to my use case to live in the sops codebase. I'll just put it into the tool I'm writing directly.

ajvb:
I agree with both of you. I think a more generic option could be to add a keyservice option to `decrypt.DataWithFormat` or similar new function.

PR #672, opened 2020-05-07
Fix #671: `updatekeys` checks for config file flag

### LINKED ISSUE #671: `updatekeys` appears to ignore the config file switch
Probably somewhere around this line needs to also check if the `--config` file switch was used and change accordingly https://github.com/mozilla/sops/blob/master/cmd/sops/main.go#L421

Our workaround is to rotate the key: `sops -i -r <file>` which does honour the `--config` switch.

TuhinNair:
The 'updatekeys' subcommand did not check for the config global string flag.
 Add that check and if found use it to set configPath.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -477,8 +477,8 @@ func main() {
 				var err error
 				var configPath string
 				// Check if config is provided though the command line
-				if c.String("config") != "" {
-					configPath = c.String("config")
+				if c.GlobalString("config") != "" {
+					configPath = c.GlobalString("config")
 				} else {
 					configPath, err = config.FindConfigFile(".")
 					if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -474,9 +474,16 @@ func main() {
 				},
 			}, keyserviceFlags...),
 			Action: func(c *cli.Context) error {
-				configPath, err := config.FindConfigFile(".")
-				if err != nil {
-					return common.NewExitError(err, codes.ErrorGeneric)
+				var err error
+				var configPath string
+				// Check if config is provided though the command line
```
autrilla, line 479:
```suggestion
```

It's pretty much self-explanatory, we don't need a comment here

Issue #679, opened 2020-06-10
AWS Profiles not properly resolving the .aws/config file


gl-ahiggins:
# AWS Profile not correctly resolving .aws/config file

When attempting to use a profile in a .aws/config file, sops does not properly resolve the credentials. Example sops and .aws config files are provided below. I believe I have a fix working, with a PR coming shortly.

## SOPS file
```yaml
hello: ENC[...]
creds:
    user: ENC[...]
    password: ENC[...]
sops:
    kms:
    -   arn: arn:aws:kms:us-east-1:<account_id>:key/<key_id>
        created_at: '2020-06-09T16:55:32Z'
        enc: "..."
        aws_profile: my-account # This profile exists in .aws/config
    gcp_kms: []
    azure_kv: []
    lastmodified: '2020-06-09T16:57:40Z'
    mac: "..."
    pgp: []
    unencrypted_suffix: _unencrypted
    version: 3.5.0
```

## .aws/config file

Here we're defining a profile where we assume a role into a separate account. While assuming roles is possible with sops using the `kms_arn+role_arn` configuration, our organization uses dozens of roles to manage AWS access. Different teams accessing the same KMS key are likely to have different roles. Ideally this is solved through normal behavior of the AWS sdk, rather than be reliant on sops' implementation.

```ini
[profile my-account]
source_profile = my-main-account
role_arn = arn:aws:iam::<account_id>:role/my-automation-role
```

## Existing problems

SOPS does not currently resolve the profile correctly. This prevents standard usage where profiles defined in the .aws/config file assume AWS credentials before executing commands. The following command works with the aws cli, setting a profile through the environment variables:

```bash
# Properly assumes role through .aws/config and prints buckets
AWS_PROFILE=my-account aws s3 ls
```

In sops, the following is returned:

```
AWS_PROFILE=my-accounts sops my-file.yaml
# ...
Group 0: FAILED
  arn:aws:kms:us-east-2:<account_id>:key/<key_id>: FAILED
    - | Error decrypting key: SharedCredsLoad: failed to get profile

Recovery failed because no master key was able to decrypt the file. In
order for SOPS to recover the file, at least one key has to be successful,
but none were.
```

Similarly, if no AWS_PROFILE is provided, the one from the sops file is used. This similarly fails to resolve:

```
sops my-file.yml
# ...
Failed to get the data key required to decrypt the SOPS file.

Group 0: FAILED
[...]

PR #680, opened 2020-06-10
AWS Profile not correctly resolving .aws/config file - #679

gl-ahiggins:
Possible fix for issue #679 . After building, I am successfully able to run the following:

```bash
# Can read the profile in the sops file from .aws/config
sops my-secrets.yaml
# If profile is not in the sops file, can still set AWS_PROFILE directly
AWS_PROFILE=my-account sops my-secrets.yaml

# If profile is set in BOTH env and the sops file, the sops file still takes priority to avoid breaking change.
# Here AWS_PROFILE is ignored if profile is defined in my-secrets.yaml
AWS_PROFILE=my-account sops my-secrets.yaml
```

### DIFF
--- a/kms/keysource.go
+++ b/kms/keysource.go
@@ -195,13 +195,11 @@ func (key MasterKey) createSession() (*session.Session, error) {
 
 	config := aws.Config{Region: aws.String(matches[1])}
 
-	if key.AwsProfile != "" {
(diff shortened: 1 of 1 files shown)


ajvb:
@autrilla After reading the docs, this feels safe. It may cause confusion to existing users if they had some breaking difference between their `~/.aws/credentials` and `~/.aws/config` files, but then the `aws` cli would also be busted for them.

PR #683, opened 2020-06-22
Add support for OpenStack Barbican

rochaporto:
Adds a new keysource storing master keys under OpenStack Barbican:
https://wiki.openstack.org/wiki/Barbican

As described in the commit message, the logic is different from the other keysources as Barbican does not directly offer encrypt/decrypt logic, but works instead as a secret storage service. In this keysource we rely on Barbican to store and retrieve the master keys.

The PR is built on top of the Vault one, i'm happy to rebase if necessary against current master.


jvehent:
Please submit your patch against the `develop` branch and not the master branch.

rochaporto:
@jvehent done

ajvb:
@rochaporto Thank you for the PR.

I must say, I am currently against merging this unless:

1. There is a sizeable enough portion of our user base that would like this
1. We can work out a detailed testing guide within this PR
1. There is full coverage integration tests that can be ran locally and in CI within this PR

rochaporto:
Hi @ajvb .

Thanks for the feedback.

I'll work on the testing guide and integration tests, and in parallel inquire about interest by other people within openstack.

Thanks for checking this out.

fnpanic:
We use Barbican in our enviroment and it would be great to use upstream sops instead of custom cern version +1

barbican/keysource.go
```diff
@@ -0,0 +1,282 @@
...
+	} else {
+		provider, err = openstack.AuthenticatedClient(*opts)
+		if err != nil {
+			log.WithError(err).Error("failed to create authenticated client")
+			return nil, err
+		}
+	}
+
+	client, err := openstack.NewKeyManagerV1(provider,
+		gophercloud.EndpointOpts{Region: "cern"})
```
skuethe, line 213: @rochaporto This breaks usage outside of cern ;)
As the region is hardcoded and not using env variables, it results in

```bash
[BARBICAN]       ERRO[0000] failed to create key manager                  error="No suitable endpoint could be found in the service catalog."
[BARBICAN]       ERRO[0000] Encryption failed                             SecretHref="[REMOVED]"
Could not generate data key: [failed to encrypt new data key with master key "[REMOVED]": Cannot create KeyManager service: No suitable endpoint could be found in the service catalog.]
```

Error message

fnpanic: I agree. It would be better to source it from the OS_REGION_NAME and pass it to gophercloud like it was done with all other config vars.

https://github.com/mozilla/sops/pull/683/commits/e8d541db0ce426778ad8740e29c4984f82bbafcc#diff-772389010f5fe2d4bd05f8aed0cc0a4dd204857414fdc83611067b7aa73045adR162-R168 #

pelme:
FWIW, our organization is using sops with AWS KMS and plan to change to a provider that with OpenStack Barbican.

stmps:
is there any chance of getting this merged?

We use `sops` + `barbican` in our environment and it would be great to not rely on a fork!

Nils98Ar:
@rochaporto Do you plan to continue working on this?

rochaporto:
I'm afraid we moved on to other options of managing sensitive data, i won't be working on this any longer.

Issue #624, opened 2020-02-07
Remove temporary files, please.


xtrasolver:
SOPS is very useful tool, but there is one issue preventing it to use in production. In does not delete temporary files.

sops --verbose test.enc.yaml

[PGP]    time="2020-02-07T17:58:34+01:00" level=info msg="Decryption succeeded" fingerprint=XXX...redacted....XXX
[SOPS]   time="2020-02-07T17:58:34+01:00" level=info msg="Data key recovered successfully"
[SOPS]   time="2020-02-07T17:58:34+01:00" level=debug msg="Decrypting tree"
[CMD]    time="2020-02-07T17:58:45+01:00" level=info msg="File written successfully"

It will left clear text file in temp folder e.g.:

C:\Users\Somebody\AppData\Local\Temp\433471151

sops -version
sops 3.5.0 (latest)

echo %EDITOR%
code --wait

systeminfo (redacted)
OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.19041 N/A Build 19041

dnicholas2001:
This is reproducible on Windows 10 and Git-Bash:

1. using default vim editor - if no changes, then the clear text file is not removed
1. using custom editor, EDITOR="atom -w", if no changes, the clear text file is not removed
1. Also, if there's a problem opening the editor, such as using sublime, the clear text file is not removed
1. However, if using vim or atom, if there's changes, then, the clear text file is removed

ajvb:
Fixed with #685

PR #685, opened 2020-06-28
Close tmpfile after writing

lbonanomi:
This PR closes https://github.com/mozilla/sops/issues/624

Windows will not allow for deletion of a file that has an open handle. 

Close tmpfile after writing to prevent unencrypted tmpfiles out-living
the execution of sops.

### DIFF
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -134,7 +134,7 @@ func editTree(opts editOpts, tree *sops.Tree, dataKey []byte) ([]byte, error) {
 	}
 	
 	// Close temporary file, since Windows won't delete the file unless it's closed beforehand
-	tmpfile.Close()
+	defer tmpfile.Close()
 
 	// Compute file hash to detect if the file has been edited
 	origHash, err := hashFile(tmpfile.Name())
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -133,7 +133,7 @@ func editTree(opts editOpts, tree *sops.Tree, dataKey []byte) ([]byte, error) {
 		return nil, common.NewExitError(fmt.Sprintf("Could not write output file: %s", err), codes.CouldNotWriteOutputFile)
 	}
 	
-	// Close temporary file to placate Windows Delete call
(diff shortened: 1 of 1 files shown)


autrilla:
Does it actually fix the issue completely? We should probably delete the temp file entirely.

lbonanomi:
> Does it actually fix the issue completely? 

I think so, but I didn't find the original problem and would like to get a confirmation from yourself (as a maintainer) and @xtrasolver (who opened the issue) before I say "done" with certainty. 

> We should probably delete the temp file entirely.

Yes indeed! Here's my 3-cent analysis: Linux lets you delete a file from disk while its open in an editor but Windows protects a file while it is open in an editor. Adding a close statement lets the downstream `os.RemoveAll` [here](https://github.com/mozilla/sops/blob/4bc27f6eb72b1b4090753e9f3dba1d094544e1c3/cmd/sops/edit.go#L114) to run to completion in git bash.

cmd/sops/edit.go
```diff
@@ -132,6 +132,9 @@ func editTree(opts editOpts, tree *sops.Tree, dataKey []byte) ([]byte, error) {
 	if err != nil {
 		return nil, common.NewExitError(fmt.Sprintf("Could not write output file: %s", err), codes.CouldNotWriteOutputFile)
 	}
+	
+	// Close temporary file to placate Windows Delete call
```
autrilla, line 136:
```suggestion
	// Close temporary file, since Windows won't delete the file unless it's closed beforehand
```

autrilla:
Thanks for your explanation, I missed that we were actually deleting the entire directory.

However, you're currently closing the file before we're done using it. I'd add a `defer` before closing so the file is closed after the function has fully run. LGTM otherwise.

lbonanomi:
> Thanks for your explanation, I missed that we were actually deleting the entire directory.
>
> However, you're currently closing the file before we're done using it. I'd add a `defer` before closing so the file is closed after the function has fully run. LGTM otherwise.

Per your instruction I added a `defer`. Just to be complete: this didn't influence my test results in git-bash. Thank you!

autrilla:
Thanks!

Issue #684, opened 2020-06-23
AWS KMS asymmetric key support


james-callahan:
I'd like to be able to allow *anyone* to encrypt secrets to me; but only allow decryption through sops/KMS.
AWS support this with key_usage of `ENCRYPT_DECRYPT`. https://docs.aws.amazon.com/kms/latest/developerguide/symm-asymm-concepts.html#asymmetric-cmks

When I attempted to just use an RSA_4096 KMS key with sops I got:
>  Failed to call KMS encryption service: InvalidKeyUsageException: Algorithm SYMMETRIC_DEFAULT is incompatible with key spec RSA_4096.

reegnz:
This feature ist still desired.

I'd also add one more aspect to it, which is offline encryption (without API access to the KMS key):
https://aws.amazon.com/blogs/security/how-to-use-aws-kms-rsa-keys-for-offline-encryption/

beachygreg:
+1 on this feature. It would be very advantageous to not have to have AWS credentials and roles in order to encrypt.

seanorama:
Commenting to keep this from going stale.

This would be a big win.

The need to have the key eliminates KMS as a choice in many scenarios, especially between departments where one team, without privileged access, needs to pass credentials to an operations team.

It also prevents the ability, to work "offline" to encrypt secrets.

nathanwebsterdotme:
This would be very useful

peter-leonov-ch:
Something similar to [ejson](https://github.com/Shopify/ejson) would we awesome.

PR #689, opened 2020-07-06
Add AWS KMS asymmetric keys support

cousine:
* Update aws-go to version 1.33.1
* Add AWS KMS Asymmetric Key Support
* Update README

Addresses #684


autrilla:
AFAICT (please correct me if I'm wrong) this lets folks use some key types they otherwise wouldn't be able to use with SOPS, but it doesn't add full asymmetric key support, right? If I were to export the pubkey from AWS and give it to someone, how could they create a SOPS file and send it to me?

cousine:
@autrilla you are correct this doesn't add full asymmetric key support, it will only allow SOPS to use the asymmetric key type for AWS KMS

cousine:
I'm curious though if there was any design thought regarding full asymmetric support, would love to take a stab at it

cousine:
Just thinking out loud, but one way full asymmetric support in the way you described is to provide the public key and the key resource (in the case of AWS the ARN) of the key, the arn would then be used for metadata, while the public key is used for encryption.

Looking at SOPS architecture, I can see that happen in two places, either we create another keysource that implements the encrypt function but composes other keysources for decryption depending on the decryption source provided, or we implement the public key encryption in the base package while building the tree.

Another question would rise though, and I can imagine this is a more common use case, if someone wants to add a secret to an already encrypted sops file, and they only have a public key, how can they do that? is there a way where the meta can be recalculated? maybe layering the encryption while keeping the structure of the store? WDYT?

roshan1988:
Will this fix me merged anytime in the roadmap ? Is there any workaround for using asymmetric keys from kms in sops ?

PR #690, opened 2020-07-08
Fix endless loop in x/crypto/openpgp func ReadMessage

uwehdaub:
I wasn't able to write tests for it, because for this problem to occur, I had to produce a slightly inconsistent state 
with private keys. I guess this only occurs if some old private key (GPG 1?) are migrated to GPG 2.
But this is only guessing.

Nevertheless I was able to test it locally with my GPG key, which produced this problem, before I fixed it.

### DIFF
--- a/pgp/keysource.go
+++ b/pgp/keysource.go
@@ -323,7 +323,7 @@ func (key *MasterKey) passphrasePrompt() func(keys []openpgp.Key, symmetric bool
 	maxCalls := 3
 	return func(keys []openpgp.Key, symmetric bool) ([]byte, error) {
 		if callCounter >= maxCalls {
-			return nil, fmt.Errorf("function passphrasePrompt called more than once")
+			return nil, fmt.Errorf("function passphrasePrompt called too many times")
 		}
 		callCounter++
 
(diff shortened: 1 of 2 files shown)


uwehdaub:
I fixed the test, so that they run locally, but this now breaks circleCI.
Is this a well-known problem with the tests?

pgp/keysource.go
```diff
@@ -318,39 +318,48 @@ func (key *MasterKey) fingerprintMap(ring openpgp.EntityList) map[string]openpgp
...
-			return nil, err
-		}
-		for _, k := range keys {
-			k.PrivateKey.Decrypt(pass)
+func (key *MasterKey) passphrasePrompt() func(keys []openpgp.Key, symmetric bool) ([]byte, error) {
+	callCounter := 0
+	maxCalls := 3
+	return func(keys []openpgp.Key, symmetric bool) ([]byte, error) {
+		if callCounter >= maxCalls {
+			return nil, fmt.Errorf("function passphrasePrompt called more than once")
```
autrilla, line 326:
```suggestion
			return nil, fmt.Errorf("function passphrasePrompt called too many times")
```

I have no idea what the cause of the issue is, but if it's known, we could hint at the user what it is.

autrilla:
> I fixed the test, so that they run locally, but this now breaks TravisCI.
> Is this a well-known problem with the tests?

It is. https://github.com/mozilla/sops/pull/576 had a fix for it, but we never ended up merging it.

autrilla:
Thanks!

Issue #652, opened 2020-04-14
Can't use --input-type for exec-file


Xartos:
## Describe the problem
There is no way to force subcommand exec-file to use a specific input format with the `--input-type`.

## How to reproduce
```
$ cat test.yaml
superSecret: ENC[AES256_GCM,data:NtahnsRDqbVULO3sH2jPnw==,iv:caX9CGdMKnDukieQTRfa52QFgNvaf86GALZR67uolX8=,tag:Yni0LEXnSM3BpWhC8lt+MA==,type:str]
sops:
    kms: []
    gcp_kms: []
    azure_kv: []
    lastmodified: '2020-04-14T15:01:38Z'
    mac: ENC[AES256_GCM,data:l2TQUc+PkDKwv+cb1RTo7v8C/4ej5X+byT7DYVowIm13GvGgET05V/R/LadjrZya/aGnSBTUZ87iitiFQvmQmpelxcvoxYOE5fh/Me7uNV7hWRJk8qilHL0zZleApuS32KXK0QVXtgKJ911AQ4tcfeaBz8RdneJb8p31lD7iR3M=,iv:GhLe9YQYP7b2hIP2nzD2wflP+u21t6mhjJldFiM9pYs=,tag:l1IxV/U3T7tU4M4WjGglhA==,type:str]
    pgp:
    -   created_at: '2020-04-14T15:01:37Z'
        enc: |-
            -----BEGIN PGP MESSAGE-----
            <REMOVED>
            -----END PGP MESSAGE-----
        fp: <REMOVED>
    unencrypted_suffix: _unencrypted
    version: 3.5.0

$ cp test.yaml test

$ sops exec-file test.yaml 'cat {}'
superSecret: superSecretValue

$ sops --input-type yaml exec-file test 'cat {}'
Error unmarshalling input json: invalid character 's' looking for beginning of value
```

## Expected behaviour
That the flag `--input-type` should force sops to decrypt as yaml format.

Xartos:
```
$ sops --version
sops 3.5.0 (latest)
```

PR #699, opened 2020-07-22
Support --input-type for exec-file

lbonanomi:
This PR resolves https://github.com/mozilla/sops/issues/652, adding support for an `--input-type` argument that will be honored by `sops exec-file`.

Until the argument for `func outputStore` was reversed (to use `input-type` instead of `output-type`) sops kept crashing with the message: 

> Error dumping file: No binary data found in tree

This is certainly a debatable fix, but results were positive.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -190,7 +190,7 @@ func main() {
                                 },
 				cli.StringFlag{
 					Name:  "output-type",
-					Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the binary output format",
+					Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the input file's extension to determine the output format",
 				},
 			}, keyserviceFlags...),
 			Action: func(c *cli.Context) error {
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -184,6 +184,10 @@ func main() {
 					Name:  "user",
 					Usage: "the user to run the command as",
 				},
+				cli.StringFlag{
+                                        Name:  "input-type",
+                                        Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the file's extension to determine the type",
+                                },
 			}, keyserviceFlags...),
 			Action: func(c *cli.Context) error {
 				if len(c.Args()) != 2 {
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -966,8 +970,9 @@ func inputStore(context *cli.Context, path string) common.Store {
 	return common.DefaultStoreForPathOrFormat(path, context.String("input-type"))
 }
 
+// SCARY! SHOULD THIS BE USING INPUT-TYPE?
```
autrilla, line 973: No, using output-type is correct. Please revert me, or show why you think I'm wrong :D

lbonanomi: @autrilla I freely-admit that I'm probably poking the wrong part of this code. Here was my (simplistic) test.

With the file test containing only: "`superSecret: superSecretValue`"

and a sops/v3/cmd/sops/main.go containing:

```
func outputStore(context *cli.Context, path string) common.Store {
        //fmt.Println("Doing the right thing")
        //return common.DefaultStoreForPathOrFormat(path, context.String("output-type"))

        fmt.Println("Doing the wrong thing")
        return common.DefaultStoreForPathOrFormat(path, context.String("input-type"))
}
```

I get the following results:

```
# sops exec-file --input-type yaml /tmp/test 'cat {}'
Doing the right thing
Error dumping file: No binary data found in tree
```

```
# sops exec-file --input-type yaml /tmp/test 'cat {}'
Doing the wrong thing
superSecret: superSecretValue
```

Thank you for your time.

autrilla: I see! I think the right thing to do here would be adding support for --output-type too. The issue is that SOPS tries to decide on the output store to use based on the file name, and assumes it should be binary (since `/tmp/test` does not have an extension that indicates a known file type). You'll need to make your calls to sops be of the form `sops exec-file --input-type yaml --output-type yaml /tmp/test 'cat {}'` afterwards.

lbonanomi: > I think the right thing to do here would be adding support for --output-type too. 

Well that's fair. PR is updated for same.

cmd/sops/main.go
```diff
@@ -184,6 +184,14 @@ func main() {
 					Name:  "user",
 					Usage: "the user to run the command as",
 				},
+				cli.StringFlag{
+                                        Name:  "input-type",
+                                        Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the file's extension to determine the type",
+                                },
+				cli.StringFlag{
+					Name:  "output-type",
+					Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the binary output format",
```
autrilla, line 193:
```suggestion
					Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the input file's extension to determine the output format",
```

What you wrote is partially incorrect. SOPS will use the input file's extension. If the extension is not one of the known types, binary will be used.

lbonanomi: Okay yeah, I see what you mean now. Thank you for the patch!

PR #708, opened 2020-07-24
Fixes #707 Update Config Tests

### LINKED ISSUE #707: Failing Make Install (wrong test inputs)
`make install` is failing due to failing test of [sops/config/config_test.go](https://github.com/mozilla/sops/blob/master/config/config_test). Lines [375](https://github.com/mozilla/sops/blob/master/config/config_test.go#L375) and [379](https://github.com/mozilla/sops/blob/master/config/config_test.go#L379) of the [TestLoadConfigFileWithVaultDestinationRules](https://github.com/mozilla/sops/blob/5f7d32404639855ee4f35f187714d71fc43aa4fa/config/config_test.go#L371) are erroneous.

The test expects the output of `conf.Destination.Path` to have an `http` protocol but the function returns the output URL with `https` protocol which is causing the tests to fail.

RishabhBhatnagar:
- Update the test to use dynamically fetched url from the library
   instead of a statically typed url (uses https instead of http)

Signed-off-by: RishabhBhatnagar <bhatnagarrishabh4@gmail.com>


ajvb:
@RishabhBhatnagar Thank you for this PR and the original issue. I decided to merge a quick fix that doesn't rely on the hostname, as that part is not really needed for the test. This fix should get released (merged into master) soon.

PR #710, opened 2020-07-25
azkv: attempt Azure MSI before falling back to CLI

### LINKED ISSUE #704: azure: try MSI without requiring explicit user opt-in
So, I thought this was going to be blocked on Azure/go-autorest#541 as I assumed you were using their `GetAuthorizer` but I've just noticed that you have it as well.

Is there a reason we don't go ahead and try Azure MSI as an authentication mechanism automatically?

As of right now, I have to go out of my way to set `AZURE_AUTH_MODE` when actually using this in real deployment scenarios to get the VMs to pull from KV automatically (any other method requires round-tripping high-value creds into the VM).

Further, it is VERY rare that there is a situation where MSI and CLI creds are available and a supposed user would be surprised to find MSI being used automatically.

This change would allow the existing developer workflow to remain the same. Since MSI is not present on developer machines (or if it is, it's serving the user creds), so you will continue to use CLI creds as normal.

colemickens:
This will let sops work out of the box on Azure MSI-enabled VMs, rather than requiring `AZURE_AUTH_MODE` to get set to MSI.

This actually is slightly closer to what the SDK has for it's default behavior in `GetAuthorizer` (where they were already falling back to MSI). I almost refactored this a bit more to use `GetAuthorizer` for the bulk of the work, but I didn't want to break existing handling of `AZURE_AUTH_MODE`.

This is the cleanest I could get the change. We need the explicit call to `ServicePrincipalToken()` to exercise the MSI-path and ensure we get a token back. But after that, we want to use the `Authorizer()`'s return value (that doesn't give us easy access to a token refresh function... hence the manual check ahead of time) so that later refreshes are automatic, etc.

And our `go-autorest` packages were quite out of date, and I need a function from a newer version, so I updated it.

This fixes #704.

**PR NOTE**: I'm leaving this as a draft until I complete another day of e2e testing (on both encryption(cli)/decryption(msi) sides). And then I'll delete this line and mark it ready.


colemickens:
This is basically blocked by: https://github.com/Azure/go-autorest/issues/542

ajvb:
@colemickens ping me or autrilla when this PR is ready for review.

ajvb:
@colemickens Going to close this for now, but feel free to ping me if you wanna reopen it.

colemickens:
This basically should all be removed and replaced by their single call Authorizer thing now, but I'm not really investing energy in this area, right now.

PR #722, opened 2020-08-20
one line fix for install in readme

tcbtcb:
To do the `git checkout development` you need to be one extra level deep in the hierarchy.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,7 +29,7 @@ For the adventurous, unstable features are available in the `develop` branch, wh
 .. code:: bash
 
 	$ go get -u go.mozilla.org/sops/v3/cmd/sops
-        $ cd $GOPATH/src/go.mozilla.org/sops/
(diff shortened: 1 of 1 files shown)


ajvb:
That's not how those version numbers work.

felixfontein:
@ajvb that's how go works apparently :) If I repeat the steps, the current instructions fail because
```
~/go/src/go.mozilla.org/sops$ git branch
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
On the other hand,
```
~/go/src/go.mozilla.org/sops/v3$ git branch
* develop
  master
```

felixfontein:
(This should be a PR against `develop` though...)

ajvb:
Well then lol...

Issue #751, opened 2020-09-29
stores/ini - Characters # and ; get parsed as comments


enote-kane:
There are many cases, where characters such as `#` (hash) and `;` (semicolon) are required inside values and the target parser does not support the value quoting via ` (backticks).

Examples:
```ini
Password = UJ&J#22222
PostUp = iptables -A FORWARD -i tun0 -j ACCEPT; iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
use = egg:Paste#http
```

PR #752, opened 2020-09-29
[stores/ini] Relax ini parser

enote-kane:
Specifically: Allow hash/semicolon in values and Python multi-line indented values - refs #751


felixfontein:
Looking at this in the context of #1597, I think this is a breaking change. Adding it to #1435.

PR #756, opened 2020-10-07
Added a way to ignore MAC validation to the decrypt package

ricje:
The CLI provides a way to skip MAC validation which was not mirrored to the decrypt package.

Simply adding a way to duplicate what can be done using the sops CLI with the `--ignore-mac` flag.


ricje:
Wrong target branch, sorry

Issue #384, opened 2018-10-04
Sops swallows YAML comment of each first key:value pair


perry123456:
I'm used to extensively describe the contents of my YAML files with comments. But with sops I have the problem, that each comment line, which is located right above of a key-value-pair which stands on top of a hierarchy level, disappears when the yaml file gets encrypted with sops.

1) My environment:
Windows 10
Sops version 3.1.1

2) My document in clear text:
```
# level 1, pos 1 comment
key1: value
# level 1, pos 2 comment
key2: value
# level 1, pos 3 comment
key3:
    # level 2, pos 1 comment
    key1: value
    # level 2, pos 2 comment
    key2: value
```

3) The encrypted document (without sops meta data):
```
key1: ENC[AES256_GCM,data:UwPZMFg=,iv:xPdFZTYT6sYQn6OCPKeqw00EcqAif9VUuBPz6VEb+vU=,tag:0ciGE608lqo+AcrM63zz+A==,type:str]
#ENC[AES256_GCM,data:bAnAxvnHLi4UoPBALf9S/piTuJg7vko=,iv:kEoo5qeUnO0q9gYdx34kUSryuOfCG0BbTVVVkRVjeq8=,tag:YHhn21H+AQPTzV400+p1Fg==,type:comment]
key2: ENC[AES256_GCM,data:kRPrHrE=,iv:+Q1byOFYvVM5rjzZYOGldCRpWPv7r+Ih4H/xsO+u9FI=,tag:U0u0zq6kJjjPbwSnkLkQYg==,type:str]
#ENC[AES256_GCM,data:zO9n8/aQ2qhyhMWRvcZfB632pUOoEjk=,iv:Ah8H79x3eVJbRtCBwvrzDSoYCzy2WCcuJEJ5B1/EQN0=,tag:dEtOiCgw8sD0eeHn/BkVmA==,type:comment]
#ENC[AES256_GCM,data:9HDfsvGdJQAW9qw6Z7t6x8yTT5LFgCQ=,iv:sALPLPo4IZrKfA2X4VqSPlnSppOfou5IW3/2+IQac3U=,tag:7xzhSZgKKMAkU+DN4Eo6lw==,type:comment]
key3:
    key1: ENC[AES256_GCM,data:lHvpmKc=,iv:a5eaxNvCWDAlyXT3yf3c/7sP3WVm7GjwwfKQ14NgGDY=,tag:yCEeeji7+5D2ySlzWCA8RA==,type:str]
[...]

Issue #695, opened 2020-07-16
sops/3.6.0 encrypted yaml file is broken (even to sops)


icy:
I have the following yaml file:

```
## Configuration for prometheus-node-exporter subchart
##
prometheus-node-exporter:
  podLabels:
    ## Add the 'node-exporter' label to be used by serviceMonitor to match standard common usage in rules and grafana dashboards
    ##

    jobLabel: node-exporter
  extraArgs:
    - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+)($|/)
    - --collector.filesystem.ignored-fs-types=^(autofs|binfmt_misc|cgroup|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|mqueue|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|sysfs|tracefs)$
```

When it's encrypted with `sops`, I get the output as below

```
prometheus-node-exporter:
## Add the 'node-exporter' label to be used by serviceMonitor to match standard common usage in rules and grafana dashboards
##  podLabels:
        jobLabel: node-exporter
    extraArgs:
    - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+)($|/)
    - --collector.filesystem.ignored-fs-types=^(autofs|binfmt_misc|cgroup|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|mqueue|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|sysfs|tracefs)$
sops:
    kms:
    -   arn: arn:aws:kms:eu-central-1:0000000:key/75f9aedd-ce9e-4b85-b112-00000000
        created_at: '2020-07-16T17:43:07Z'
        enc: AQICAHgquprv1l5xoZPFQCgP7aJyPm2SG/VQw5mKqJ1aWMOo1wGMTOvYqCZquWh0DzogTo7uAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMgkcK1GZYgX5WXfFJAgEQgDsRmTsCzr7vNGSLvP0sZ7zGQcXqg+5pEHNOp3H7jjVdyUAFY/LowjnEmWzWGdQCku4hljPHv/N3MXrjMw==
        aws_profile: ""
    gcp_kms: []
    azure_kv: []
    hc_vault: []
[...]

Issue #374, opened 2018-08-30
Encryption process strips leading comments


grantbachman:
I'm using Sops 3.05, and found that if a file contains a leading comment, like in the example below, it is completely absent in the encrypted version of the file. Is there a reason for this or is this just a bug?

```yaml
# Comment
key: value
# Another Comment
```

autrilla:
This is a known bug (see https://github.com/mozilla/sops/issues/300#issuecomment-367942757). The reason for this is that it was too complicated to support this in [our YAML parser](https://github.com/mozilla-services/yaml).

grantbachman:
Thanks for the quick reply! Completely understandable. It seems more and more tools are written in Go, which I don't yet know...

Issue #300, opened 2018-02-23
Panic when encrypting yaml file with document separator followed by comment


mccutchen:
I've found a panic when encrypting a yaml document that begins with a document separator (i.e. `---`) followed by a comment.

### Demonstration

```
$ cat testcase-panic.yml
---
# comment
key: value
```
```
sops-3.0.2.darwin --encrypt testcase-panic.yml
panic: expected end of document event but got 9 [recovered]
	panic: expected end of document event but got 9

goroutine 1 [running]:
go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml.handleErr(0xc420162738)
	/home/travis/gopath/src/go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml/yaml.go:192 +0x9f
panic(0x1622580, 0xc420049bc0)
	/home/travis/.gimme/versions/go1.9.linux.amd64/src/runtime/panic.go:491 +0x283
go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml.(*parser).document(0xc420152900, 0x1028689)
	/home/travis/gopath/src/go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml/decode.go:144 +0x2d1
go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml.(*parser).parse(0xc420152900, 0x174af90)
	/home/travis/gopath/src/go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml/decode.go:117 +0x6b
go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml.CommentUnmarshaler.Unmarshal(0xc4201e0000, 0x19, 0x219, 0x16016e0, 0xc420149b60, 0x0, 0x0)
	/home/travis/gopath/src/go.mozilla.org/sops/vendor/github.com/mozilla-services/yaml/yaml.go:53 +0x161
go.mozilla.org/sops/stores/yaml.Store.Unmarshal(0xc4201e0000, 0x19, 0x219, 0x1010b88, 0xa0, 0x16bd660, 0x1, 0xc4201be460)
	/home/travis/gopath/src/go.mozilla.org/sops/stores/yaml/store.go:39 +0x88
go.mozilla.org/sops/stores/yaml.(*Store).Unmarshal(0x1b18cf8, 0xc4201e0000, 0x19, 0x219, 0x219, 0x0, 0x0, 0x1c45b48, 0x0)
	<autogenerated>:1 +0x5a
[...]

PR #757, opened 2020-10-07
Do not discard comments at beginning of YAML documents

felixfontein:
I couldn't find documentation mentioning that the top-of-document comments are discarded (that needs updating). @autrilla you mentioned that this behavior is documented in https://github.com/mozilla/sops/issues/695#issuecomment-659920093; can you point me to where?

Also, I couldn't run the functional tests since I don't have Rust set up right now, but I guess they'll run in CI :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Develop branch
+--------------
+
+Changes:
+
+    * Comments at the beginning of YAML documents are now no longer discarded (#756)
+
+Bug fixes:
+
+    * Fix several problems with YAML comment handling described in #695 (#754)
+
 3.6.1
 -----
 Features:
--- a/functional-tests/res/comments.enc.yaml
+++ b/functional-tests/res/comments.enc.yaml
@@ -1,3 +1,4 @@
+#ENC[AES256_GCM,data:IYA+b4ORDq8u9CBQolipWD4HRqoZyA==,iv:F8ldQqGng+WptHuBkFtjrGM+7sRZCsvd0FHq98lrpAE=,tag:ZHbLU9+CELinf5PhhuIzSQ==,type:comment]
 lorem: ENC[AES256_GCM,data:PhmSdTs=,iv:J5ugEWq6RfyNx+5zDXvcTdoQ18YYZkqesDED7LNzou4=,tag:0Qrom6J6aUnZMZzGz5XCxw==,type:str]
 #ENC[AES256_GCM,data:HiHCasVRzWUiFxKb3X/AcEeM,iv:bmNg+T91dqGk/CEtVH+FDC53osDCEPmWmJKpLyAU5OM=,tag:bTLDYxQSAfYDCBYccoUokQ==,type:comment]
 dolor: ENC[AES256_GCM,data:IgvT,iv:wtPNYbDTARFE810PH6ldOLzCDcAjkB/dzPsZjpgHcko=,tag:zwE8P+AwO1hrHkgF6pTbZw==,type:str]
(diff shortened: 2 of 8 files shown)


autrilla:
I don't think it's in the readme, but it's in several issues, e.g. #300, #374, and probably others. I've found a few instances scattered in comments on several other issues.

This also fixes https://github.com/mozilla/sops/issues/384

autrilla (approved):

autrilla:
Thanks again!

felixfontein:
@autrilla thanks a lot for reviewing and merging! :)

BTW, is there a schedule/... for when a new release will happen?

patricknelson:
Since this hasn't been released yet: Bump 😄 

Also, since I cannot test this, I want to confirm: Will this retain or discard comments throughout the document? I have a particular use case for `ytt` (YAML templating) such as an [encrypted values file](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/docs/data-values-vs-overlays.md): 

```yaml
#@data/values
---
testing: 'foobar'
```

And since there are no values above that document divider `---`, it could still end up being discarded, but not sure.

patricknelson:
p.s. If anyone lands here from a google search who _also_ happens to be using both `sops` and `ytt`, here's my current hack/workaround:

```bash
echo -e "#@data/values\n---\n$(sops -d encrypted-values.yml)" | ytt -f - -f template-file.yml
```

This prepends the necessary `ytt` comment and document separator and then uses the STDIN syntax `-f -`.

felixfontein:
@patricknelson mozilla/sops#791 together with go-yaml/yaml#684 now preserves constructs such as
```.yaml
#@data/values
---
testing: 'foobar'
```

patricknelson:
Oh ok cool @felixfontein! I commented here mainly since I _assumed_ that this merged PR would already perform that function. Is there a difference between your example and this new one below when it comes to this PR and the one you created?

```yaml
# comment
testing: 'foobar'
```

vs.

```yaml
# comment but as separate document
---
testing: 'foobar'
```

I only have to ask since I cannot test this since I'm unable to compile this and get it to work on my machine. So my gual is for this to hopefully get **released** soon 😄 (pretty please @autrilla 🙏).

autrilla:
I'm not a project owner, so I can't create releases unfortunately.

felixfontein:
@patricknelson this PR mainly bumps the version of the modified copy of yaml.v2 that is used by sops, which improves comment parsing. Since comment parsing is only part of the modified copy of yaml.v2, and not of the official yaml.v2, this is kind of a maintenance nightmare, since only sops developers are using this modified copy of yaml.v2 (at least that I'm aware of).

My new PR switches this modified copy of yaml.v2 by the official [yaml.v3](https://github.com/go-yaml/yaml/tree/v3) API, which does support comments. Unfortunately yaml.v3 is a work in progress itself, has bugs (w.r.t. comment processing) and not that actively developed (just compare https://github.com/go-yaml/yaml/commits/v3 with https://github.com/go-yaml/yaml/issues?q=is%3Aissue+is%3Aopen+v3). But still, since this is used by other projects, the chances of yaml.v3 being improved are higher than our modified copy of yaml.v2 being improved.

I've also created a PR for yaml.v3 which fixes some comment problems, including a bug which causes yaml.v3 to read your construct
```.yaml
#@data/values
---
testing: 'foobar'
```
as a single document
```.yaml
#@data/values
testing: 'foobar'
```
(which thus won't allow sops to treat it correctly). So when the yaml.v3 PR is merged (and a new version released), and the sops PR for using yaml.v3 is updated to use the latest yaml.v3 version, my new sops PR should be ready to go.

It's probably better to wait for the yaml.v3 version of sops, instead of releasing a new version with this PR included, to make sure that there's only one breaking change coming up (see https://github.com/mozilla/sops/pull/791#issuecomment-752548622).

@autrilla is @ajvb the only project owner at the moment?

autrilla:
As far as I know, yes, but I think technically anyone in the @mozilla org has permission to do everything in this repo.

patricknelson:
Just confirming that I'm using version 3.7.1 (released April 2021) and this feature still isn't yet available after a full year. I know it was merged into `develop`, but still... this isn't a breaking or backward incompatible, is it? Technically it does change output but I can't see how emitting a comment at the beginning of a YAML document would break things.

@ajvb just curious: what is the process/workflow for incorporating changes from `develop` into `master` so that they're ultimately released as an executable that we can use?

felixfontein:
@patricknelson for me it's working with 3.7.1.

patricknelson:
@felixfontein Got a _little_ egg on my face here; turns out I was actually using 3.**6**.1 and not 3.7.1. So the good news is that thankfully I can put comments at the top of YAML files which will help a little bit, however...

I just upgraded and found that my multi-document format (the format I still need it to support from `ytt`) is doing what you were describing in your last comment, specifically: The comment above the document separator `---` ends up coalescing incorrectly into a single document, which isn't valid (as `ytt` parses these comments and they still have semantic meaning).

> ```yaml
> #@data/values
> ---
> testing: 'foobar'
> ```
> 
> as a single document
> 
> ```yaml
> #@data/values
> testing: 'foobar'
> ```

So for now I'll still be using [my old polyfill script above](https://github.com/mozilla/sops/pull/757#issuecomment-748322761) in the meantime until a solution is found for that. I suppose _that_ is a separate issue unrelated to this ticket?

patricknelson:
Seems like a separate issue (comments as their own document in a YAML stream), so I created #936 to track/discuss.

PR #761, opened 2020-10-23
add filename to exec-file

cgroschupp:
Some tools (e.g. Terraform) require the right file extension to recognize the file type.

see more details for terraform at https://www.terraform.io/docs/configuration/variables.html#variable-definitions-tfvars-files

i get this error
```sh
~/.go/bin/sops exec-file test.json "terraform plan -var-file={}"

Error: Argument or block definition required

  on /tmp/.sops640510710/tmp-file line 1:
   1: {

An argument or block definition is required here.

exit status 1
```
works for me:
```sh
~/.go/bin/sops exec-file --filename test.json test.json "terraform plan -var-file={}"

Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.
...
```

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -132,6 +132,11 @@ func main() {
 					Name:  "user",
 					Usage: "the user to run the command as",
 				},
+				cli.StringFlag{
+					Name:  "filename",
+					Usage: "filename for the temporarily file (default: tmp-file)",
+				},
+
 			}, keyserviceFlags...),
 			Action: func(c *cli.Context) error {
 				if len(c.Args()) != 2 {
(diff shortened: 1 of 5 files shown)


cgroschupp:
@ajvb @autrilla can you please review?

autrilla:
Before we add a new feature that we'll have to maintain forever, can you explain why this is useful for you? The file name gets passed to the command anyway, and there's no way for the command to know what it'll be ahead of time because the directory will have a random name.

cgroschupp:
@autrilla I have updated the description.

cmd/sops/main.go
```diff
@@ -211,12 +216,18 @@ func main() {
 					return toExitError(err)
 				}
 
+				filename := c.String("filename")
+				if filename == "" {
+					filename = "tmp-file"
+				}
```
autrilla, line 222: This is broken, since you're not using `filename` again

autrilla (changes requested):
I'm OK with adding the feature, but other than my code comment, can you please:

1. Document this in README.md (similar to how e.g. `--user` is documented).
2. Modify the description with a more concrete example of a tool (actual commands, terraform works for me) that needs this to work properly? It would be useful for other people searching for a solution when they encounter the same problem you did.

autrilla (approved):
Thanks! We have some CI issues, but I'll make sure this PR will make it into the next release.

renannprado:
Looking forward to use this feature :smile:

autrilla:
Can you please rebase so this branch picks up the CI fixed that are now in develop?

cgroschupp:
@autrilla done

autrilla:
Thanks!

PR #762, opened 2020-10-29
Returning stderr when gpg encrypt fails

cloudstarz:
We have been having a lot of trouble investigating some failures that seem to be very flaky on a few of our projects where we use sops as a library. All we see is a message that looks like the following:

```
failed to encrypt new data key with master key "681E3A89EB1DAFD36EB883120A73BB48E26694D8": could not encrypt data key with PGP key: golang.org/x/crypto/openpgp error: key with fingerprint 681E3A89EB1DAFD36EB883120A73BB48E26694D8 is not available in keyring and could not be retrieved from keyserver; GPG binary error: exit status 2]
```

This verbose error message will hopefully be more helpful to investigate and fix the underlying issue. The current error `GPG binary error: exit status 2` does not give too many details today and has been making it virtually impossible to understand the failure on a CI machine that we might or might not have access to.


autrilla:
Can you make this merge to `develop` please? Looks good to me otherwise.

cloudstarz:
@autrilla done. thanks for taking a look.

ajvb (approved):
lgtm

PR #788, opened 2020-12-17
Use age/armor for encrypted data key

kouk:
Currently the encrypted data key is stored as a binary value, and this
results in SOPS encrypted DOTENV files having weird binary characters.

This changes the encrypt/decrypt methods to use the armor reader writer
provided by: filippo.io/age/armor

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"filippo.io/age"
+	"filippo.io/age/armor"
 	"github.com/sirupsen/logrus"
 	"go.mozilla.org/sops/v3/logging"
 )
@@ -45,8 +46,9 @@ func (key *MasterKey) Encrypt(datakey []byte) error {
 
 		key.parsedRecipient = parsedRecipient
 	}
+	armorWriter := armor.NewWriter(buffer)
 
-	w, err := age.Encrypt(buffer, key.parsedRecipient)
+	w, err := age.Encrypt(armorWriter, key.parsedRecipient)
 	if err != nil {
 		return fmt.Errorf("failed to open file for encrypting sops data key with age: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


autrilla:
We can't accept this change as is since it's not backwards compatible, unfortunately.

kouk:
@autrilla given that the support for age isn't yet released I would think that backwards compatibility isn't a blocker and perhaps it would be preferable to correct the age functionality which in some sense is a bit broken right now (unless it's OK for committing binary files to git). In any case, would you accept it if it was behind an opt-in feature flag?

autrilla:
> @autrilla given that the support for age isn't yet released I would think that backwards compatibility isn't a blocker and perhaps it would be preferable to correct the age functionality which in some sense is a bit broken right now (unless it's OK for committing binary files to git).

Ah, then yeah, this should be fine. Let's see what @jimmycuadra thinks of it, who did much of the work to implement the feature in the first place.

> In any case, would you accept it if it was behind an opt-in feature flag?

I'd rather not have yet another knob. Given it's not released yet, we should be able to just change the behavior.

jimmycuadra:
Using armor over binary makes sense to me. No objections here.

autrilla (approved):
Looks good, thanks!

autrilla:
Needing CI to pass before merging is going to be annoying for this PR in particular...

kouk:
Seems like it timed out fetching some pgp key. I forced push in order to retry, but I'm wondering if there an easier way for me to retry the build if this one also fails.

autrilla:
Unfortunately the issue is that the GPG key server SOPS uses by default is flaky and/or completely broken, so it's unlikely that'll work...

kouk:
i pushed a new version of the commit (changed my email), but now the build has failed due to other test suite errors, which unfortunately I can't reproduce locally (tests pass fine). Doesn't seem to be related to the keyserver. I have `rustc 1.43.0`, while travis seems to be installing 1.48. Is there a chance that's the issue?

autrilla:
Unlikely. I don't have permission to just force merge unfortunately, but I know the issues here are unrelated to your code. Given that we'll have to move off Travis CI soon, we might take the chance to fix this stuff then :). Maybe @ajvb can merge this in the meantime.

autrilla:
We're definitely going to need this PR before releasing age support, because currently JSON files are broken if the age data is stored as a binary blob instead of armored.

kouk:
is there a way forward here? cc @ajvb @jvehent @autrilla

autrilla:
> is there a way forward here? cc @ajvb @jvehent @autrilla

Once the CI issues are solved, we can merge this. It shouldn’t be very long.

ajvb:
@kouk Could you please merge `mozilla:develop` into your branch and push up the merge? This should fix CI for this PR.

tullo:
Is this still blocked b/c of the move off Travis CI?

Looks like @kouk is having a break from GH and this PR is not going anywhere soon.

kouk:
sorry for the delay in responding, I see that someone has submitted another PR which got merged. So closing this one..

Issue #774, opened 2020-11-25
Errors with make test


ursuad:
Hi, when I run `make test` some of the tests are failing because of an issue with the PGP key, and if I comment that out I get a different error, for another test. 
Any idea why this is happening? 

The Go version I use is `go version go1.15.5 darwin/amd64`
```
$ make test
GO15VENDOREXPERIMENT=1 GO111MODULE=on GOPROXY=https://proxy.golang.org go mod tidy
GO15VENDOREXPERIMENT=1 GO111MODULE=on GOPROXY=https://proxy.golang.org go mod vendor
gpg --import pgp/sops_functional_tests_key.asc 2>&1 1>/dev/null || exit 0
gpg: key 3D16CEE4A27381B4: "SOPS Functional Tests Key 1 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key 3D16CEE4A27381B4: secret key imported
gpg: key D8720D957C3D3074: "SOPS Functional Tests Key 2 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key D8720D957C3D3074: secret key imported
gpg: key 3D16CEE4A27381B4: "SOPS Functional Tests Key 1 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key D8720D957C3D3074: "SOPS Functional Tests Key 2 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key 19F9B5DAEA91FF86: "SOPS Functional Tests Key 3 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: Total number processed: 5
gpg:              unchanged: 5
gpg:       secret keys read: 2
gpg:  secret keys unchanged: 2
./test.sh
ok      go.mozilla.org/sops/v3  0.372s  coverage: 51.7% of statements
ok      go.mozilla.org/sops/v3/aes      0.390s  coverage: 76.9% of statements
?       go.mozilla.org/sops/v3/audit    [no test files]
ok      go.mozilla.org/sops/v3/azkv     0.171s  coverage: 25.5% of statements
?       go.mozilla.org/sops/v3/cmd/sops [no test files]
?       go.mozilla.org/sops/v3/cmd/sops/codes   [no test files]
?       go.mozilla.org/sops/v3/cmd/sops/common  [no test files]
[...]

PR #793, opened 2020-12-29
Remove dead broken code

felixfontein:
The function `encodeMetadataItem` in stores/ini/store.go is not used anywhere, and fails compilation (see #774).

The failure might be Go version dependent (my `go version` says `go version go1.15.6 linux/amd64`), but since the code isn't used, I guess it's best to just remove it.

### DIFF
--- a/stores/ini/store.go
+++ b/stores/ini/store.go
@@ -252,56 +252,6 @@ func (store *Store) encodeMetadataToIniBranch(md stores.Metadata) (sops.TreeBran
 	return branch, nil
 }
 
-func encodeMetadataItem(prefix string, kind reflect.Kind, field reflect.Value) (map[string]interface{}, error) {
-
-	result := make(map[string]interface{}, 0)
-
-	switch kind {
-	case reflect.Slice:
-		slf := field
-		for j := 0; j < slf.Len(); j++ {
-			item := slf.Index(j)
-			p := fmt.Sprintf("%s[%d]", prefix, j)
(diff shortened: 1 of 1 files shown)


felixfontein:
The only use of that function was removed in dc4a6975000071b507b199ccbb2f2726c63ddc95, the second commit touching stores/ini/store.go.

autrilla (approved):

autrilla:
It does seem version dependant, since it works for me on 1.13. But yeah, let's get rid of it.

felixfontein:
It seems to be more a warning with go 1.15 (golang/go#32479, golang/go#3939), but that seems to make `go test` fail. I don't know go well enough, maybe `go test` compiles with warnings as errors. In any case, simply removing the code block (since the corresponding decoding block is already long gone) really seems like the best choice :)

felixfontein:
@autrilla @ajvb I've rebased my two PRs to latest `develop`, but they still try to run tests with Travis. I think the commit which removes Travis and adds GitHub Actions needs to get pulled into `master` before CI works again in PRs.

felixfontein:
@ajvb it seems that Travis is still active for this repository, which causes it to show up as a CI step which never runs and blocks completion of CI (especially since it is marked as "Required"). I guess someone with admin rights needs to disable / remove it.

ajvb:
@felixfontein Thanks, fixed.

felixfontein:
@autrilla @ajvb thanks a lot for reviewing and merging!

PR #796, opened 2021-01-06
Use rustup directly instead of actions-rs and run Go unit tests in Github Actions

autrilla:


ajvb (approved):
Awesome!

From looking at https://github.com/mozilla/sops/runs/1659479127?check_suite_focus=true - where are we running our go unit tests?

autrilla:
Good point, looks like I forgot about that. Now they should run on all the OSs we build on.

autrilla:
It's fixed now. Running tests on a cross-compiled binary doesn't work very well.

ajvb:
Awesome!

PR #817, opened 2021-02-17
Add wrap verb to formatted error outputs

mkatychev:
* Replaces `%v` formatting of errors using `fmt.Errorf` with `%w`.
* In cases where an `Unwrap` method is not used, `%w` behaves in the _exact_ same way as `%v`.

[Reference:](https://blog.golang.org/go1.13-errors#TOC_3.3.)
> In Go 1.13, the fmt.Errorf function supports a new `%w` verb. When this verb is present, the error returned by `fmt.Errorf` will have an Unwrap method returning the argument of `%w`, which must be an error. In all other ways, `%w` is identical to `%v`.

This would allow returned errors such as this one to check and handle cases when the underlying error is `os.ErrNotExist` using `os.IsNotExist(err)`:
https://github.com/mozilla/sops/blob/8e21de8dbc750f050a54d63209f2f1f93477168c/decrypt/decrypt.go#L21-L24

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -48,17 +48,17 @@ func (key *MasterKey) Encrypt(datakey []byte) error {
 
 	w, err := age.Encrypt(buffer, key.parsedRecipient)
 	if err != nil {
-		return fmt.Errorf("failed to open file for encrypting sops data key with age: %v", err)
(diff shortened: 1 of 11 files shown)


autrilla (approved):
Thanks!

PR #819, opened 2021-02-20
Use age/armor for encrypted data key

tullo:
Currently the encrypted data key is stored as a binary value, and this
results in SOPS encrypted DOTENV files having weird binary characters.

This changes the encrypt/decrypt methods to use the armor reader writer
provided by: filippo.io/age/armor

Signed-off-by: Andreas Amstutz <tullo@users.noreply.github.com>

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"filippo.io/age"
+	"filippo.io/age/armor"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


tullo:
I run into the same problem as pointed out in https://github.com/mozilla/sops/pull/788

With the current version I'm not able to decrypt age encrypted DOTENV files. With the patch in this PR the files get decrypted as expected.

autrilla (approved):

PR #827, opened 2021-03-03
Match file path relative to config file's directory

### LINKED ISSUE #480: Behavior of path_regex changed in patch release and broke some usecases
Since release 3.3.1 (mozilla/sops#478) path regexes check against the full path as implemented in PR mozilla/sops#466 (issue mozilla/sops#465).  This solved the usecase mentioned in the issue but broke another one.

Consider this config (taken from mozilla/sops#466):

```yaml
creation_rules:
    - path_regex: .*/development/.*
      kms: '…'

    - path_regex: .*/production/.*
      kms: '…'

    - kms: '…'
```

If the repo that contains the secrets is located under a path like `/home/user/development/secrets/`, the first rule will always match.  I have no control over the location where other developers store their repos.

We tried to prevent this by using `^(./)?development/.+\.yaml` as a `path_regex` and calling sops always from the project root, but this no longer works with sops 3.3.1 and fails with *error loading config: no matching creation rules found*.  It even fails for existing encrypted files, despite the irrelevance of creation rules for them.

**Possible solution**

Instead of simply converting the path passed to sops to an absolute path, the absolute path should be made relative to the config file (if one exists).

This should support both the use case from mozilla/sops#465 and ours.

paulolieuthier:
Fixes #480.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -650,7 +650,7 @@ and its KMS and PGP keys are used to encrypt the file. It should be noted that
 the looking up of ``.sops.yaml`` is from the working directory (CWD) instead of
 the directory of the encrypting file (see `Issue 242 <https://github.com/mozilla/sops/issues/242>`_).
 
-The path_regex checks the full path of the encrypting file. Here is another example:
+The path_regex checks the path of the encrypting file relative to the config file. Here is another example:
 
 * files located under directory **development** should use one set of KMS A
 * files located under directory **production** should use another set of KMS B
--- a/README.rst
+++ b/README.rst
@@ -650,7 +650,7 @@ and its KMS and PGP keys are used to encrypt the file. It should be noted that
 the looking up of ``.sops.yaml`` is from the working directory (CWD) instead of
 the directory of the encrypting file (see `Issue 242 <https://github.com/mozilla/sops/issues/242>`_).
 
-The path_regex checks the path of the encrypting file relative to the config file. Here is another example:
+The path_regex checks the path of the encrypting file relative to the .sops.yaml config file. Here is another example:
 
 * files located under directory **development** should use one set of KMS A
 * files located under directory **production** should use another set of KMS B
--- a/config/config.go
+++ b/config/config.go
@@ -67,7 +67,6 @@ func FindConfigFile(start string) (string, error) {
 type configFile struct {
 	CreationRules    []creationRule    `yaml:"creation_rules"`
 	DestinationRules []destinationRule `yaml:"destination_rules"`
-	Path             string
 }
 
 type keyGroup struct {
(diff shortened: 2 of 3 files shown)


config/config.go
```diff
@@ -334,6 +336,14 @@ func LoadCreationRuleForFile(confPath string, filePath string, kmsEncryptionCont
...
 		return nil, err
 	}
+
+	configDir, err := filepath.Abs(filepath.Dir(confPath))
+	if err != nil {
+		return nil, err
+	}
+
+	// compare file path relative to path of config file
+	filePath = strings.TrimPrefix(filePath, configDir + "/")
```
autrilla, line 346: Please use `filepath.Separator` instead.

autrilla (changes requested):
Thanks for your PR. Before feeling comfortable merging this, I would like to see the documentation mention how the creation rules path is computed, and some tests as well.

paulolieuthier:
Hey @autrilla, please tell me if it's ok now.

README.rst
```diff
@@ -623,7 +623,7 @@ and its KMS and PGP keys are used to encrypt the file. It should be noted that
 the looking up of ``.sops.yaml`` is from the working directory (CWD) instead of
 the directory of the encrypting file (see `Issue 242 <https://github.com/mozilla/sops/issues/242>`_).
 
-The path_regex checks the full path of the encrypting file. Here is another example:
+The path_regex checks the path of the encrypting file relative to the config file. Here is another example:
```
autrilla, line 626:
```suggestion
The path_regex checks the path of the encrypting file relative to the .sops.yaml config file. Here is another example:
```

autrilla (dismissed):
Looks good, thanks!

autrilla:
Oh, can you make this merge against `develop` please?

paulolieuthier:
Rebased to develop!

config/config.go
```diff
@@ -65,6 +67,7 @@ func FindConfigFile(start string) (string, error) {
 type configFile struct {
 	CreationRules    []creationRule    `yaml:"creation_rules"`
 	DestinationRules []destinationRule `yaml:"destination_rules"`
+	Path             string
```
autrilla, line 70: Sorry I missed this in the last review, but this struct is used to parse the .sops.yaml file, and I don't want this exposed in there. Can you move this out of the struct and just pass it in as a parameter to the functions?

autrilla (changes requested):
Just one thing I missed in the original review, and then we're good to go.

config/config.go
paulolieuthier: No problem. Tell me if it's ok now.

paulolieuthier:
@autrilla I just saw you reverted the merge. Was that because of the failing tests (I just saw that as well). Do you have anything against merging a new PR (with passing tests)?

autrilla:
Yes, it was because of failing tests after merging. Nothing against merging
a new PR that merges with passing tests.

Issue #826, opened 2021-03-02
Support more complicated path_regexp


sfrevel:
Is it possible to support more complicated regexp in path_regex? Like mentioned in the documentation here https://golang.org/pkg/regexp/#MatchString

"More complicated queries need to use Compile and the full Regexp interface"

so the match has to be implemented more like this

validRegExp = regexp.MustCompile(path_regexp)
match = validRegExp.MatchString(filePath))

I think this is the reason for issues like this one https://github.com/mozilla/sops/issues/480

We also had problems to specify a path which is including the filename and not just a filename. Especially if there are '/' in the path it's not working at all and we get the message 'no matching creation rules found'

Example not working (no matter if '/' are escaped or not!):

path_regexp: stages/dev/secrets.env$

Examples working:

path_regexp: .*dev.*
path_regexp: secrets\.env$

autrilla:
We would be happy to take a patch for this. An accompanying test to show the improvement would be excellent too.

bcolucci:
https://github.com/mozilla/sops/pull/829

PR #829, opened 2021-03-05
feat: Support more complicated path_regexp (issues/826)

bcolucci:
Ref: https://github.com/mozilla/sops/issues/826

This should works. But @sfrevel, if you have better examples so I can complete the tests? (but I guess it could be okay like that already...).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -10,7 +10,6 @@ import (
 	"path"
 	"regexp"
 
-	"gopkg.in/yaml.v3"
 	"github.com/sirupsen/logrus"
 	"go.mozilla.org/sops/v3"
 	"go.mozilla.org/sops/v3/age"
@@ -21,6 +20,7 @@ import (
 	"go.mozilla.org/sops/v3/logging"
 	"go.mozilla.org/sops/v3/pgp"
 	"go.mozilla.org/sops/v3/publish"
+	"gopkg.in/yaml.v3"
 )
 
 var log *logrus.Logger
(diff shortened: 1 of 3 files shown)


sfrevel (approved):

sfrevel (dismissed):
Looks exactly like I thought it should be... Never used the Destination rule, but there's also just the simple way for patterns used. Maybe it's also worth to take a look...

config/config.go
```diff
@@ -306,11 +306,9 @@ func parseCreationRuleForFile(conf *configFile, filePath string, kmsEncryptionCo
 			rule = &r
 			break
 		}
-		if r.PathRegex != "" {
-			if match, _ := regexp.MatchString(r.PathRegex, filePath); match {
-				rule = &r
-				break
-			}
+		if regexp.MustCompile(r.PathRegex).MatchString(filePath) {
```
felixfontein, line 309: The `if r.PathRegex != "" {` is missing.

Also, I'm not sure whether using `MustCompile` here is a good idea in terms of error handling. It will panic if `r.PathRegex` does not compile. The old code was simply ignoring errors, which might not be the be solution, but panicing instead is not a good idea IMO.

felixfontein (commented):
You need to rebase the PR against the `develop` branch.

config/config.go
bcolucci: The `r.PathRegex != ""` is not required because the opposite case is already tested before, right? ^^
I'll modify the code to handle the compilation error.

bcolucci: > The `if r.PathRegex != "" {` is missing.
> 
> Also, I'm not sure whether using `MustCompile` here is a good idea in terms of error handling. It will panic if `r.PathRegex` does not compile. The old code was simply ignoring errors, which might not be the be solution, but panicing instead is not a good idea IMO.

The `if r.PathRegex != "" {` is useless because we check if the string is empty before, and if it's the case, then  it breaks. So the string is not empty at this point.
I update the code so I return an error instead of panicing.

autrilla (approved):
Thank you!

PR #842, opened 2021-03-30
Remove unused age keysource code

hiddeco:
This removes two pieces of code in the age keysource that are not
actually used.

The `parsedIdentity` is technically a candidate to stay, but should
then be changed to a `[]*age.X25519Identity` type and be lazy-loaded
by `Decrypt` (with the result of `age.ParseIdentities`).


ajvb (approved):

ajvb:
Thank you @hiddeco !

PR #849, opened 2021-04-05
config: support setting age and PGP keys as a strings array

### LINKED ISSUE #845: Feature request: support arrays for `age` and `pgp` fields in configuration
Hey.

It would be great to be able to set age and PGP keys as arrays of strings in the configuration. This would be more in phase with YAML style and would allow putting comments, e.g.:

```
creation_rules:
  - path_regex: \.sops\.yaml$
    […]
    age:
      - age1xxx # Foo
      - age1yyy  # Bar
    pgp:
      - XXX, # Foo
      - YYY  # Bar
```

I believe sops already turns the comma separated value into an array internally, so that shouldn’t change the logic too much.

johanfleury:
Resolves #845

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"filippo.io/age"
 	"filippo.io/age/armor"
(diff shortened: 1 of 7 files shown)


johanfleury:
Ok, so I’m not a Go expert and I’m not sure this is the best way to implement this.

Unit tests pass locally, but I’d appreciate a review on this before going further.

autrilla:
I don't think this is backwards compatible, right? Functional tests are failing because of that.

johanfleury:
I can’t really tell why the functional tests are failing and when I tried to run them manually (i.e. executing sops with the same args) it seemed to work.

I can work on making them pass, but as I said, I’m not sure this is the correct way to implement this so I’ve not focused on that yet.

lastthyme (approved):

felixfontein:
Closing since this now has been implemented in #1880. Thanks a lot for your contribution!

Issue #840, opened 2021-03-25
exec-env/exec-file signal handling might be broken


si14:
Hey folks,

We've encountered some issues after we started using `exec-file` for our software. For example, `py.test` being run with `exec-file` inside a docker container would terminate on sigterm immediately, whereas normally it prints test failures first, which is quite annoying.

I'm sorry that I can't provide a minimal reproducing case as we just started exporting secrets before running our process "naked", which solved those problems for us, but after looking through the sources my theory to why that's happening is as follows:

* docker starts a docker container with `sops` as its main process
* `sops` spawns our process in a subprocess here https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/cmd/sops/subcommand/exec/exec_unix.go#L15
* the subprocess is in the same process group, which means both `sops` and the subprocess will receive incoming signals
* we ask Docker to send sigterm (or sighup)
* Docker sends sigterm (or sighup) to that process group
* `sops` exits immediately
* docker container is terminated immediately and no stdout is sent to the host

…which sounds similar to the problem someone was having in this blog post https://bigkevmcd.github.io/go/pgrp/context/2019/02/19/terminating-processes-in-go.html

If my understanding is correct, it's a subtle problem for a lot of software using sops. The solution would be for sops to spawn subprocess into a separate process group and pass signals on manually, waiting for the process to finish.

Thank your for your awesome piece of software!

autrilla:
Your assesment seems correct to me. We'd be happy to take a PR for this issue. Thanks for putting in the time to report it even if your team has found a workaround!

PR #880, opened 2021-05-31
Add same process option for exec-env

ricmatsui:
When running `sops exec-env`, the command is currently launched in a child process, signals are not forwarded, and the command continues to run even if the `sops` process is killed. This prevents `exec-env` from being usable as a wrapper for launching commands such as servers where signals and process exiting needs to be handled correctly, as mentioned in #840. 

This PR adds an option to launch the command in the same process via the `execve` system call, which causes the command to run with the new environment in the same process, immediately exiting from `sops`. This means signals and process exiting are handled correctly since the command is in control.

The option has been added to `exec-env` only, so it does not fully address #840 and more may be needed to properly handle temporary files created with `exec-file`. The option is also incompatible with the background option.

Related #1209


README.rst
```diff
@@ -1052,6 +1052,11 @@ written to disk.
     $ echo your password: $database_password
     your password:
 
+If you want process signals to be sent to the command, for example if you are
+running ``exec-env`` to launch a server and your server handles sigterm, then the
```
felixfontein, line 1056: I think signal names are usually capitalized:
```suggestion
running ``exec-env`` to launch a server and your server handles SIGTERM, then the
```

README.rst
```diff
@@ -1052,6 +1052,11 @@ written to disk.
...
+``--same-process`` flag can be used to instruct ``sops`` to start your command in
+the same process instead of a child process. This uses the ``execve`` system call
+and is supported on unix-like systems.
```
felixfontein, line 1059:
```suggestion
and is supported on Unix-like systems.
```

cmd/sops/subcommand/exec/exec.go
```diff
@@ -134,6 +135,14 @@ func ExecWithEnv(opts ExecOpts) error {
 
 	env = append(env, opts.Env...)
 
+	if opts.SameProcess {
+		if opts.Background {
+			log.Fatal("background is not supported for same-process")
+		}
+
+		return ExecSyscall(opts.Command, env)
```
felixfontein, line 143:
```suggestion
		// Note that the call does NOT return, unless an error happens.
		return ExecSyscall(opts.Command, env)
```

cmd/sops/subcommand/exec/exec.go
```diff
@@ -134,6 +135,14 @@ func ExecWithEnv(opts ExecOpts) error {
 
 	env = append(env, opts.Env...)
 
+	if opts.SameProcess {
+		if opts.Background {
+			log.Fatal("background is not supported for same-process")
```
felixfontein, line 140: I think this should also be checked in cmd/sops/main.go, so that the program exists with an error message and never gets to this point. (Please keep this check here so that if someone calls this function with an invalid combination, there's still some warning.)

cmd/sops/subcommand/exec/exec.go
```diff
@@ -134,6 +135,14 @@ func ExecWithEnv(opts ExecOpts) error {
```
felixfontein, line 135: At the top of this function, please check whether if `runtime.GOOS == "windows" && opts.SameProcess`, and either return a nice `error` in that case, or print a warning and set `opts.SameProcess = false` (similar to https://github.com/getsops/sops/blob/08476d1bed45d7b95ec7482fdaf47baf65bb258a/cmd/sops/subcommand/exec/exec.go#L63-L66).

Right now the caller can only see from the log output that something went wrong and the command wasn't run at all; the exit code will be 0 since no error is reported.

felixfontein (commented):
Thanks for your contribution! I think the PR looks good; I've mainly added some comments on the docs and error reporting.

ricmatsui:
Thanks for reviewing! I've add the docs changes and improved error handling, trying to match existing conventions.

felixfontein (approved):

felixfontein:
@ricmatsui thanks for your contribution!

Issue #882, opened 2021-06-02
Can't pass golang unit tests


lavrxxx:
Hello, I want to add something to project but I can't pass unit tests in develop branch. What am I doing wrong?
```
vr-Work-MacBook sops % make test
GO15VENDOREXPERIMENT=1 GO111MODULE=on GOPROXY=https://proxy.golang.org go mod tidy
GO15VENDOREXPERIMENT=1 GO111MODULE=on GOPROXY=https://proxy.golang.org go mod vendor
gpg --import pgp/sops_functional_tests_key.asc 2>&1 1>/dev/null || exit 0
gpg: key 3D16CEE4A27381B4: "SOPS Functional Tests Key 1 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key 3D16CEE4A27381B4: secret key imported
gpg: key D8720D957C3D3074: "SOPS Functional Tests Key 2 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key D8720D957C3D3074: secret key imported
gpg: key 3D16CEE4A27381B4: "SOPS Functional Tests Key 1 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key D8720D957C3D3074: "SOPS Functional Tests Key 2 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: key 19F9B5DAEA91FF86: "SOPS Functional Tests Key 3 (https://github.com/mozilla/sops/) <secops@mozilla.com>" not changed
gpg: Total number processed: 5
gpg:              unchanged: 5
gpg:       secret keys read: 2
gpg:  secret keys unchanged: 2
./test.sh
ok      go.mozilla.org/sops/v3  0.412s  coverage: 51.7% of statements
ok      go.mozilla.org/sops/v3/aes      0.449s  coverage: 76.9% of statements
ok      go.mozilla.org/sops/v3/age      0.224s  coverage: 49.3% of statements
?       go.mozilla.org/sops/v3/audit    [no test files]
ok      go.mozilla.org/sops/v3/azkv     0.231s  coverage: 25.5% of statements
?       go.mozilla.org/sops/v3/cmd/sops [no test files]
?       go.mozilla.org/sops/v3/cmd/sops/codes   [no test files]
?       go.mozilla.org/sops/v3/cmd/sops/common  [no test files]
[...]

PR #887, opened 2021-06-08
Fix #881 (#464): Added JSON encoder to disable escaping html symbols

### LINKED ISSUE #464: json decoding does not seem to properly respect character encoding
Hello,
I'm using sops 3.2.0 with gcp kms. I noticed that decoding using json input type does not respect (decode) the keys and values' text encoding.

It is easy to reproduce for the '&' character (https://www.fileformat.info/info/unicode/char/0026/index.htm):

'testing1' file contents
```
{
  "key&?abcd":"value?p1=true&p2=false"
}
```

Encrypt as a json input type file and a binary input file
```
sops --encrypt --input-type=json --output-type=json --gcp-kms projects/someproject/locations/global/keyRings/sops/cryptoKeys/sops-key testing1 > testing1.json.enc  

sops --encrypt --input-type=raw --output-type=raw --gcp-kms projects/someproject/locations/global/keyRings/sops/cryptoKeys/sops-key testing1 > testing1.raw.enc
```

Decrypt the raw encrypted version (ok)
```
sops -d --input-type=raw --output-type=raw testing1.raw.enc
```
```
{
  "key&?abcd":"value?p1=true&p2=false"
}
```

Decrypt the json encrypted version (not ok)
```
sops -d --input-type=json --output-type=json testing1.json.enc
```
```
{
	"key\u0026?abcd": "value?p1=true\u0026p2=false"
}
```

### LINKED ISSUE #881: Encoding of JSON file contains unicode representation instead of symbol
## Relations

Possible related to issues: #464 & #740 

### Inputs
Currently, having simple JSON file, like (`test.json`):
```json
{
    "TestValue": "test <test@test.test>",
    "TestValue2": "&",
    "TestValue3": "@"
}
```

Using the `.sops.yaml` file with KMS defined and command:

```bash
sops -e -i test.json
```

And decrypt:
```bash
sops -d --input-type=json -i test.json

```

### Results

After decrypting the input file doesn't decode all symbols, like (this is founds only, might be more):
- `&` [Unicode Character 'AMPERSAND' (U+0026)](https://www.fileformat.info/info/unicode/char/0026/index.htm)
- `<` [Unicode Character 'LESS-THAN SIGN' (U+003C)](https://www.fileformat.info/info/unicode/char/003c/index.htm)
- `>` [Unicode Character 'GREATER-THAN SIGN' (U+003E)](https://www.fileformat.info/info/unicode/char/003e/index.htm)

and result is the following:

```json
{
    "TestValue": "test \u003ctest@test.test\u003e",
    "TestValue2": "\u0026",
    "TestValue3": "@"
}
```

It might be a huge issue, while the example in the first value should represent the proper email info, like:
`User Name <user@example.org>` instead of `User Name \u003cuser@example.org\u003e`.

## Environment
- SOPS version: sops 3.7.1 (latest)
- Cloud: AWS / KMS
- OS: MacOS Big Sur (11.3.1)

## Expected results

The JSON values should be decoded in the same manner for all symbols to avoid data discrepancy.

Moskovych:
Fixes #881
Fixes #464

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -149,6 +149,16 @@ func (store Store) treeBranchFromJSONDecoder(dec *json.Decoder) (sops.TreeBranch
 	}
 }
 
+// Encoder to disable escaping html symbols
+// See: https://github.com/getsops/sops/issues/881
+func jsonMarshal(v interface{}) ([]byte, error) {
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(v)
+	return buffer.Bytes(), err
+}
+
 func (store Store) encodeValue(v interface{}) ([]byte, error) {
 	switch v := v.(type) {
 	case sops.TreeBranch:
@@ -156,7 +166,7 @@ func (store Store) encodeValue(v interface{}) ([]byte, error) {
 	case []interface{}:
 		return store.encodeArray(v)
 	default:
-		return json.Marshal(v)
+		return jsonMarshal(v)
 	}
 }
 


Moskovych:
@felixfontein , @autrilla , could you please review PR and approve builds?
Any additional changes are required? More tests?
Functionality is not changed - as you said, output file is semantically the same.

autrilla (approved):
LGTM. Because this is technically a breaking change, I'd like @ajvb's LGTM too. Like I've mentioned before, I think this kind of breaking change is not really breaking, since we make no guarantees about the syntax of the output, and only about the semantics.

autrilla:
A test for this would be nice to have

Moskovych:
> @autrilla A test for this would be nice to have

Ok, will work on it.

autrilla:
Hmmm, looks like the tests are broken

Moskovych:
@autrilla , and looks like it's pgp one, which I didn't changed.

Moskovych:
I can't proceed with adding new tests as existing are not working properly (see #882)

tanandy:
hi guys we have the same issue, any updates here ?

Moskovych:
@tanandy , sorry for delaying, but this PR still require unit test to be covered. I'll try to add them asap.

akshaypatidar1999:
@ajvb @autrilla Can you please release this? We are also facing similar issue

Moskovych:
Still blocked by #977.

hiddeco:
Have you tried one of the keys from this history tree? https://github.com/getsops/sops/commits/main/pgp/sops_functional_tests_key.asc

Moskovych:
@hiddeco , thanks for pointing me out, but correct me if I'm wrong:
to fix the tests: [stores/json/store_test.go](https://github.com/getsops/sops/blob/1ea4a74f658105ddd0b76ff5a33fef55f3eca23b/stores/json/store_test.go)
I need to decrypt the file: [stores/json/test_resources/example.json](https://github.com/getsops/sops/blob/1ea4a74f658105ddd0b76ff5a33fef55f3eca23b/stores/json/test_resources/example.json)
where it contains aws kms key (for which I don't have access) or gpg `E5297818703249D0C60E19E6824612478D1A4CCD`

From what I see, there are no such key (the last one, gpg), even in history of repository.

Did I missed something?

felixfontein:
[stores/json/test_resources/example.json](https://github.com/getsops/sops/blob/1ea4a74f658105ddd0b76ff5a33fef55f3eca23b/stores/json/test_resources/example.json) isn't used in the tests.

stores/json/store.go
```diff
@@ -149,14 +149,24 @@ func (store Store) treeBranchFromJSONDecoder(dec *json.Decoder) (sops.TreeBranch
...
 
+// Encoder to disable escaping html symbols
+// See: https://github.com/getsops/sops/issues/881
+func jsonMarshal(v interface{}) ([]byte, error) {
+    buffer := &bytes.Buffer{}
+    encoder := json.NewEncoder(buffer)
+    encoder.SetEscapeHTML(false)
+    err := encoder.Encode(v)
+    return buffer.Bytes(), err
+}
```
felixfontein, line 160: Please run gofmt to reformat this file (you use spaces instead of tabs for indentation). Also you need to sign off the commit (for DCO). Besides that, I think the PR looks good.

Moskovych:
@felixfontein , github workflows are not running without any approvals. Can you help with it?

stores/json/store.go
```diff
@@ -149,14 +149,24 @@ func (store Store) treeBranchFromJSONDecoder(dec *json.Decoder) (sops.TreeBranch
 	}
 }
 
+// Encoder to disable escaping html symbols
+// See: https://github.com/getsops/sops/issues/881
+func jsonMarshal(v interface{}) ([]byte, error) {
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(v)
```
felixfontein, line 158: `Encode` appends a newline (https://pkg.go.dev/encoding/json#Encoder.Encode), which `json.Marshal` apparently does not do. (This makes the unit tests fail.) This is apparently because `Encoder` is for JSON streams, while `Marshal` produces a single JSON file.

felixfontein: Simply adding code which removes a trailing newline (if exists) probably suffices to fix this. I don't see another way to do this with Go's API.

Moskovych: Shouldn't we stick with newline in new files instead?

felixfontein: This function is called from `encodeValue` and used during serialization of any JSON value, so if you use this to create JSON of a map, you get someting like
```json
{
  "foo"
: "bar"
,
  "baz"
: 42
,
  "bam"
: null
}
```

bscaleb:
Any updates here?

popsikle:
Is it possible to get this merged now that #977 is merged?

PR #893, opened 2021-06-30
Update Golang to 1.16

czunker:
Golang 1.13 has reached EOL:
https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance

Tests should parse, as soon as #882 is fixed.


ajvb:
Closed with #1012

Issue #692, opened 2020-07-09
Adding support for ssh keys for encryption.


Mic92:
SSH keys could be also used for encryption. They are already in well-known locations i.e. `/etc/ssh/ssh_host_rsa_key`. Most developer/servers have already ssh keys.
Since most users also have ways of trusting those ssh keys (TOFU, DNS, hosters like github/gitlab or certificates) importing them should be easy. For my [project](https://github.com/Mic92/sops-nix) I am thinking about converting ssh keys to gpg keys. However it would be probably also beneficial for sops itself to have this supported out-of-the box. I think usability of ssh-keygen over gpg is out of question.

Mic92:
Here is some boilercode to convert ssh keys to gpg: https://gist.github.com/Mic92/24c40996cd97cb8edd53fd688c60ab6f

jvehent:
I agree with the initial request of using SSH keys for encryption, and #688 will give us that because `age` supports ssh keys.

The PGP key storage format is an abomination and it would be a mistake to add more complexity. Ultimately, we need to steer people away from PGP entirely.

gzm55:
@jvehent unfortunately, #688 missing ssh keys supports

Mic92:
I will built a tool eventually to convert ed25519 as well to age keys, like I did for gpg.

D3vl0per:
Bump?

Mic92:
I have built: https://github.com/Mic92/ssh-to-age https://github.com/Mic92/ssh-to-pgp/ for use with sops. ssh-to-age can be used for ed25519 keys and ssh-to-pgp for rsa-based ssh keys.

childnode:
@Mic92 does your implementation literaly the same as https://github.com/FiloSottile/age/blob/main/agessh/agessh.go#L190 << https://blog.filippo.io/using-ed25519-keys-for-encryption age internally does on providing an ssh-ed25519 wrapping it into a X25519 curve encoded as Bech32 with HRP `AGE-SECRET-KEY-`?! looks like you copied some code and redistributed it under MIT while age provided it under [BSD](https://github.com/FiloSottile/age/blob/main/LICENSE) 🤔

Mic92:
BSD and MIT are compatible with each other.

PR #898, opened 2021-07-01
Add support for Age SSH

iamd3vil:
Currently sops only supports age with age keys. This PR adds support for using SSH keys for encryption, decryption using age.

#### Encryption

```bash
./sops --ssh /home/user/.ssh/id_ed25519.pub -e -i secrets.yaml
```

`--ssh` flag here denotes the ssh keys for encrypting.

#### Decryption

```bash
./sops -d enc.json
```

If there is no `SOPS_AGE_SSH_PRIVATE_KEY` env variable given, sops will check `~/.ssh/id_ed25519` and fallbacks to `~/.ssh/id_rsa`.

Let me know if I need to do any changes.

P.S: I have updated to the latest version as well.


itscaro:
Any news on this PR?

cedi (approved):

hikhvar:
This fixes #692 and would allow a really streamlined process combined with gitops tools like argocd or flux.

agessh/keysource.go
```diff
@@ -0,0 +1,241 @@
...
+		userConfigDir, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("user config directory could not be determined: %w", err)
+		}
+
+		// Check if there is a Ed25519 Key
+		ageKeyFilePath = filepath.Join(userConfigDir, ".ssh", "id_ed25519")
+		_, err = os.Stat(ageKeyFilePath)
+		if err != nil {
+			if err == os.ErrNotExist {
```
overfl0, line 113: This seems to always fail and the code never falls back to looking for `id_rsa`.
Changing the check to `if os.IsNotExist(err)` makes it work correctly.
@iamd3vil

hiddeco:
GitOps tool maintainer here (Flux). Thanks a lot for this contribution, I have no doubt this will be extremely useful to quite some users. :1st_place_medal: 

I have assigned this to myself to review, but am waiting for #1064 to land first as I have a gut feeling it would be better to merge the two key source into one than to introduce an additional one. For this, I need a clear view on the state of `develop` post-merge in combination with this PR. Please hold for a tiny bit longer :bow:

saimonn:
@hiddeco #1064 is now merged, is there some way we could help to merge this current PR as well ?

hiddeco:
Sorry for the wait folks, this has been on my to-do list for some time but #1072 and #1085 had a bit more priority.

Based on a quick study of the [current key source implementation in `develop`](https://github.com/mozilla/sops/blob/develop/age/keysource.go) and age itself, I am wondering if the current `age.X25519Recipient` in this implementation could be replaced with a more generic `age.Recipient`. We could then load decryption keys (identities) from all known files (either X25519 or SSH), and re-use most logic already there with some minor changes to deal with different receiver string types. WDYT?

Sebor:
Hello @hiddeco!
Any news about the review?

neongreen:
Would also love to have this. Perhaps there is somebody else who can pick up the review?

neongreen:
Or should it be closed in favor of https://github.com/getsops/sops/pull/1134? @hiddeco

hiddeco:
While I really do appreciate your work here @iamd3vil, I am going to close this in favor of #1134 which incorporates the feedback I gave in https://github.com/getsops/sops/pull/898#issuecomment-1176522057. Thank you very much nonetheless! :bow:

PR #912, opened 2021-07-28
introduce a --pristine flag to sops exec-env

e-nikolov:
The purpose of this flag is to modify the exec-env's behaviour to only insert decrypted secrets into the environment of the supplied command, without forwarding any of the existing environment variables. 

This is useful in cases where the environment needs to be sanitized from the existing environment variables.

This feature is inspired by [envconsul](https://github.com/hashicorp/envconsul)'s --pristine flag

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -137,6 +137,10 @@ func main() {
 					Name:  "background",
 					Usage: "background the process and don't wait for it to complete",
 				},
+				cli.BoolFlag{
+					Name:  "pristine",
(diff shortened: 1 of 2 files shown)


felixfontein:
@e-nikolov could you please rebase, sign-off on the commit, force-push the result? I think this would be a useful addition.

e-nikolov:
@felixfontein Done

felixfontein (approved):
Looks good to me.

hiddeco (approved):

felixfontein:
@e-nikolov thanks for your contribution!

PR #915, opened 2021-08-09
Redirect signals to a child process

deliro:
Hi!

Currently, SOPS doesn't redirect any signals to the child process when running in `exec-file` or `exec-env` modes so the child process can't react to a signal (e.g. `SIGHUP` for reloading or `SIGTERM`/`SIGINT` to shutdown graceful). Instead, SOPS shuts down without giving any chance to the child process.

This PR causes SOPS to wait for the child process to finish first and redirect all the signals it receives to the child process. Also, child processes are created with different process group ID which means that only SOPS receives signals, and not both.

Fixes https://github.com/mozilla/sops/issues/840
Related https://github.com/mozilla/sops/pull/880

### DIFF
--- a/cmd/sops/subcommand/exec/exec.go
+++ b/cmd/sops/subcommand/exec/exec.go
@@ -2,8 +2,11 @@ package exec
 
 import (
 	"bytes"
+	"context"
 	"io/ioutil"
 	"os"
+	"os/exec"
+	"os/signal"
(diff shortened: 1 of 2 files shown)


dmedvinsky:
Hate to be that guy, but kinda bumping this. Would love to see this PR progress. Can anyone take a look?

PR #925, opened 2021-08-29
switch to openpgp fork

Mic92:
upstream no longer maintains this library:
https://github.com/golang/go/issues/44226
Hence bugfixes that are not security relevant are unlikely to get fixed
as well as new features.
Hence this commit switches to protonmail's fork, which is still
regularly updated: https://github.com/ProtonMail/go-crypto

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,8 +40,8 @@ require (
 	github.com/smartystreets/goconvey v0.0.0-20190710185942-9d28bd7c0945 // indirect
 	github.com/stretchr/testify v1.5.1
 	go.mozilla.org/gopgagent v0.0.0-20170926210634-4d7ea76ff71a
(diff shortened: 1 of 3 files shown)


Mic92:
I don't expect this to have an impact on binary size since it just replaces the original openpgp library.

Enrico204:
Any improvement changing the library? I mean, is `sops` having bugs caused by the openpgp library? The current library will be updated in case of any security issue, and it seems quite stable

Mic92:
I am not following development too closely, as I am preparing to switch to age. There are a bunch of bugfixes as you can see here: https://github.com/ProtonMail/go-crypto/pulls?page=1&q=is%3Apr+is%3Aclosed

PR #946, opened 2021-10-07
Allow age decryption directly from environment variable

rnijveld:
One of the things we noticed when using age for encryption/decryption is that the current sops implementation requires us to first store the private key on disk before we can decrypt secrets. Unfortunately with modern file systems and on SSDs we cannot ensure that the secret key data is fully removed if it was ever on disk. 

This pull request adds some code that allows specifying the contents of the age keyfile as an environment variable, thus preventing secret keys from ever having to be stored on disk.

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -96,27 +96,36 @@ func (key *MasterKey) SetEncryptedDataKey(enc []byte) {
 
 // Decrypt decrypts the EncryptedKey field with the age identity and returns the result.
 func (key *MasterKey) Decrypt() ([]byte, error) {
-	ageKeyFilePath, ok := os.LookupEnv("SOPS_AGE_KEY_FILE")
+	var ageKeyData io.Reader
+	ageKeyText, ok := os.LookupEnv("SOPS_AGE_KEY")
+	ageKeyFilePath := "SOPS_AGE_KEY"
+	if ok {
+		ageKeyData = strings.NewReader(ageKeyText)
+	} else {
+		var ok bool
+		ageKeyFilePath, ok = os.LookupEnv("SOPS_AGE_KEY_FILE")
 
-	if !ok {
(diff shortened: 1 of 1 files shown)


p0bailey:
I would love this feature, especially when it comes to automation. Having the secret key exposed as an environment variable would simplify the workflow and especially won't leave the key.txt file hanging around. Great stuff @rnijveld rnijveld 

    env:
      SOPS_SECRET_KEY: ${{ secrets.sops_secret_key }}

p0bailey:
Hi @rnijveld 
your patch works fine within sops itself opening an encrypted file ( sops secrets.yaml ) where SOPS_AGE_KEY is exported as env variable as shown below.

export SOPS_AGE_KEY="AGE-SECRET-KEY-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"

However, when I use sops terraform provider https://github.com/carlpett/terraform-provider-sops the process fails 
to read the same secrets.yaml file. Quite surprisingly it works using export SOPS_AGE_KEY_FILE="${PWD}/key.txt"

│ Error: Error getting data key: 0 successful groups required, got 0
│
│   with data.sops_file.secrets,
│   on locals.tf line 2, in data "sops_file" "secrets":
│    2: data "sops_file" "secrets" {
│

I hope my feedback will help others struggling with the same problem.

Phillip

ajvb:
This was resolved with #1006. Thank you though!

p0bailey:
Thanks,

great stuff, really appreciated.

Phillip

p0bailey:
Hi @ajvb 

any plans when this would be merged into master ?

Phillip

felixfontein:
`develop` is merged into `master` when a release is made.

p0bailey:
Thanks @felixfontein 

any plans for a new release?

Phillip

Issue #681, opened 2020-06-16
Read GCP key content from environment?


reidab:
The GCP KMS implementation uses Google's Application Default Credentials scheme, which requires the path to a key file to be passed set as `GOOGLE_APPLICATION_CREDENTIALS`.

We've run into a few places in our build process where we'd like to use sops with the GCP KMS, but where it's not feasible for us to write credentials to a file beforehand. Would it make sense for sops to provide an environment variable to accept Google credentials as content instead of as a path?

The Google API client offers an initialization method using credentials content: https://pkg.go.dev/google.golang.org/api/option?tab=doc#WithCredentialsJSON

autrilla:
This functionality should probably be added to the GCP SDK instead.

Regardless, how are you able to get the sops file into the build
environment, but not a GCP credentials file?

reidab:
It's definitely something that should be added to the GCP SDK, but every discussion I've ever seen about it has ended with "we implement Application Default Credentials, that's just how it works". 😞 

The specific use case that I was trying to make work is:
- We're using Terraform to manage infrastructure and the hosted Terraform Cloud environment to apply changes.
- We have existing secrets managed in sops files that I'd like to read as part of the Terraform run. These sops files would be checked in to git alongside the terraform config.
- https://github.com/carlpett/terraform-provider-sops uses sops as a library to decrypt sops files and expose them to Terraform.
- In the context of the Terraform Cloud runner, I can set environment vars and terraform input vars, but I can't execute arbitrary setup code (e.g. to write the contents of an environment variable to a file).
- I'd like to patch `terraform-provider-sops` so that it can accept credentials directly in a terraform input var and pass them on to sops. I don't think I can do that without a change to sops to add an option that will initialize a Google client with explicit params instead of the defaults here:

  https://github.com/mozilla/sops/blob/4bc27f6eb72b1b4090753e9f3dba1d094544e1c3/gcpkms/keysource.go#L134

farant:
I might be wrong but my understanding is that GOOGLE_APPLICATION_CREDENTIALS can either be a path to the json credentials or the actual json credential contents?

Could be worth testing if that would help your use case.

reidab:
The Application Default Credentials spec says it's a file path, and that's how the Go library implements it:

https://github.com/golang/oauth2/blob/5d25da1a8d43b66f2898c444f899c7bcfd6a407e/google/default.go#L78-L85

[...]

PR #953, opened 2021-10-27
Support for GCP Service Account as JSON or Path in Default Application Credentials

joshkaplinsky:
There are many use cases where systems require mounting a GCP Service Account JSON as an environment variable directly - and cannot support mounting a file and using the default application credentials as a path to the mounted file. 

This is common in many CICD pipelines, where secrets are managed by the system but passed to Docker Containers as  environment variables. 

These changes support a custom lookup to `GOOGLE_APPLICATION_CREDENTIALS` and will run `os.Stat()` to check if the environment variable resolves as a path to a file. 
- If `err == nil` - the content of `GOOGLE_APPLICATION_CREDENTIALS` is a path to a file
    - Will use `ioutil.ReadFile` and return the contents of the file as `[]byte` or an error (if one occurs)
- if `err != nil` - an error is dropped, and will use the content of the `os.GetEnv("GOOGLE_APPLICATION_CREDENTIALS")`, converting the returned string to `[]byte`  
    - If the environment variable is not present, an empty string would be returned - and that error would be handled when trying to create the client. 

Changes in creating the client, remove the deprecated `cloudkms.New()` in favor of `cloudkms.NewService()` this allows for passing an optional `client.Option` - in the changes here - it passes `option.WithCredentialsJSON()` and the returned `creds` as `[]byte` from `getDefaultApplicationCredentials()` described above.  

This resolves issue: https://github.com/mozilla/sops/issues/681


multani:
@jkapdev Would it make sense to use another dedicated variable instead of piggy backing on the variable from the Google SDK?

Something like `GOOGLE_CREDENTIALS`.

This would make the behavior less surprising and would make it more compatible with other systems which are doing the same (*cough* terraform *cough*)

jkapdev:
> This would make the behavior less surprising and would make it more compatible with other systems which are doing the same (_cough_ terraform _cough_)

Seems like that would kill 2 birds with 1 stone. Will get that updated when I have a chance!

jkapdev:
Sorry to bug @autrilla - anyway we can get some eyes on this PR? Or get the community to help out in reviewing some changes to sops, moving forward?

MikeSchlosser16:
We are also looking for this functionality

joshkaplinsky:
@ajvb Has been merged - let me know if there's anything else I need to do.

gcpkms/keysource.go
```diff
@@ -151,3 +153,19 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	out["created_at"] = key.CreationDate.UTC().Format(time.RFC3339)
 	return out
 }
+
+// getDefaultApplicationCredentials allows for passing GCP Service Account
+// Credentials as either a path to a file, or directly as an environment variable
+// in JSON format.
+func getDefaultApplicationCredentials() (token []byte, err error) {
```
ajvb, line 160: Why use this when https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials exists? Shouldn't we check if it returns an error first and then do this custom implementation?

gcpkms/keysource.go
```diff
@@ -151,3 +153,19 @@ func (key MasterKey) ToMap() map[string]interface{} {
...
+	var defaultCredentials = os.Getenv("GOOGLE_CREDENTIALS")
```
ajvb, line 161: I believe this should be `GOOGLE_APPLICATION_CREDENTIALS`?

ajvb (changes requested):

gcpkms/keysource.go
multani: One of the goal of this pull request is to be able to read credentials directly out of an environment variable; `GOOGLE_APPLICATION_CREDENTIALS` contains the _path_ to a file containing the credentials, not the credentials itself.

This is not supported natively by the Google SDK (I can find the related issues in their repository, if needed), but other consumers may provide this, for example the [Terraform Google provider](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#using-terraform-cloud). Being able to read the credentials directly out of an environment variable would open up the possibility to use sops in context where it's not possible to read the credentials otherwise.

Hence, the deliberate use of this `GOOGLE_CREDENTIALS` environment variable, like the Google provider for Terraform does.

gcpkms/keysource.go
multani: It would be worth:

* Checking if `GOOGLE_CREDENTIALS` is set, in which case [`CredentialsFromJSON`](https://pkg.go.dev/golang.org/x/oauth2/google#CredentialsFromJSON) can be called with its content
* Otherwise, call [`FindDefaultCredentials`](https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials) and directly return its result

So that we have the special use-case when `GOOGLE_CREDENTIALS` is set, otherwise sops fallback to the default behavior of the SDK, what do you think?

ajvb: I agree, though I'd suggest that we reverse the order. It's good to stay as close to the common cloud provider SDK's as possible and this would ensure we don't break any backwards compatibility.

hiddeco: I think @multani's suggestion already is backwards compatible. Given it is not a standard environment variable at the moment, there is some intent for it to be used.

A regular user would not have this set, while by changing it around, it would not be taken into account on GCP instances where the right environment default values are automatically inject at times. Which makes `FindDefaultCredentials` always work, and `GOOGLE_CREDENTIALS` unreachable.

joshkaplinsky: I agree with @multani's thoughts here. I'll get the the call to `FindDefaultCredentials` integrated. Seems like it would be better to support both cases `GOOGLE_CREDENTIALS` and `GOOGLE_APPLICATION_CREDENTIALS` to integrate with a wider set of implementations.

gcpkms/keysource.go
```diff
@@ -131,12 +131,15 @@ func (key MasterKey) createCloudKMSService() (*cloudkms.Service, error) {
...
 
 	ctx := context.Background()
-	client, err := google.DefaultClient(ctx, cloudkms.CloudPlatformScope)
-	if err != nil {
+	var options []option.ClientOption
+
+	if credentials, err := getGoogleCredentials(); err != nil {
 		return nil, err
+	} else if len(credentials) > 0 {
+		options = append(options, option.WithCredentialsJSON(credentials))
```
joshkaplinsky, line 139: Setting client option only if a credential is returned from `GOOGLE_CREDENTIALS`. Otherwise, default to using default application credentials as defined in the [`cloudkms.NewService()`](https://github.com/googleapis/google-api-go-client/blob/5ba22f18bcf26dbbf4cd1432504269141274170f/cloudkms/v1/cloudkms-gen.go#L17-L22)

gcpkms/keysource.go
joshkaplinsky: Yes agreed, thank you @multani. 

Apologies, I misread your original [comment](https://github.com/mozilla/sops/pull/953#issuecomment-996788870). I've updated these changes to support using either `GOOGLE_CREDENTIALS` or `GOOGLE_APPLICATION_CREDENTIALS` such that if a key or path isn't provided in `GOOGLE_CREDENTIALS` it will fallback to the expected behavior from `cloudkms.NewService()` -- which uses the default application credentials automatically.

gcpkms/keysource.go
ajvb: Ok sounds good.

ajvb (approved):

PR #966, opened 2021-12-12
[Fix] sops multi recipient for age encryption

cedi:
This fixes a bug with age encryption when specifying multiple age recipients

I encountered an issue when I tried so specify multiple age recipients
in the .sops.yaml config file of my repository.

I tried running `sops --age 'agePubKey1,agePubKey2' -e -i values.secret.yaml`
which produced an appropriate file with two entries in the `/sops/age/-`
part of the encrypted yaml file.

However, I then continued to set multiple recipients in my .sops.yaml
file to simplify handling:

```yaml
creation_rules:
  - encrypted_regex: '^(data|stringData|spec)$'
    age: 'agePubKey1,agePubKey2'
```

However, this resulted in encryption only being done for the first
specified agePubKey, not the second or third one.

After digging a bit trough the code, I think this should fix it.

I verified the fix locally on my machine and got it working.

Signed-off-by: Cedric Kienzler <github@cedric-kienzler.de>


cedi:
I would love to get some traction on this by the new maintainer @rforsythe

README.rst
```diff
@@ -199,6 +215,19 @@ identity will be tried in sequence until one is able to decrypt the data.
...
+Test with the dev AGE key
+**********************
+
+If you want to test **sops** without having to do a bunch of setup, you can use
+the example files and AGE key provided with the repository::
+
+	$ git clone https://github.com/mozilla/sops.git
+	$ cd sops
+	$ export SOPS_AGE_KEY_FILE="$(pwd)/age/keys.txt"
+	$ sops example.yaml
```
ajvb, line 227: I'm a fan of moving the example files to using `age`, but I'd like to move that to separate PR or add a test in CI for this. It feels more likely that we could back out the change to the example files without a test for them, and I don't want to back out your fix for the multiple recipients.

age/keysource_test.go
```diff
@@ -31,26 +31,43 @@ func TestMasterKeyFromRecipientWithLeadingAndTrailingSpaces(t *testing.T) {
...
+	assert.Equal(len(keys), 2)
+	assert.Equal(keys[0].Recipient, "age1yt3tfqlfrwdwx0z0ynwplcr6qxcxfaqycuprpmy89nr83ltx74tqdpszlw")
+	assert.Equal(keys[1].Recipient, "age1tmaae3ld5vpevmsh5yacsauzx8jetg300mpvc4ugp5zr5l6ssq9sla97ep")
+}
+
 func TestAge(t *testing.T) {
 	assert := assert.New(t)
 
-	key, err := MasterKeyFromRecipient("age1yt3tfqlfrwdwx0z0ynwplcr6qxcxfaqycuprpmy89nr83ltx74tqdpszlw")
+	keys, err := MasterKeysFromRecipients("age1yt3tfqlfrwdwx0z0ynwplcr6qxcxfaqycuprpmy89nr83ltx74tqdpszlw,age1tmaae3ld5vpevmsh5yacsauzx8jetg300mpvc4ugp5zr5l6ssq9sla97ep")
```
ajvb, line 49: I'm wondering if as part of this change we should make `MasterKeyFromRecipient` a private func within the `age/` lib? Could help make sure this doesn't get re-introduced.

cedi: I would love that! Sounds like an absolute sane default to me!

README.rst
cedi: I will separate this.
Thanks for your feedback!

age/keysource_test.go
```diff
@@ -24,73 +24,104 @@ func TestMasterKeysFromRecipientsEmpty(t *testing.T) {
 func TestMasterKeyFromRecipientWithLeadingAndTrailingSpaces(t *testing.T) {
 	assert := assert.New(t)
 
-	key, err := MasterKeyFromRecipient("  age1yt3tfqlfrwdwx0z0ynwplcr6qxcxfaqycuprpmy89nr83ltx74tqdpszlw  ")
+	commaSeparatedRecipients := "  age1yt3tfqlfrwdwx0z0ynwplcr6qxcxfaqycuprpmy89nr83ltx74tqdpszlw  ,  age1tmaae3ld5vpevmsh5yacsauzx8jetg300mpvc4ugp5zr5l6ssq9sla97ep  "
```
ajvb, line 27: @cedi Let's keep (or add) one test that verifies that `MasterKeyFromRecipient()` works with just one key, as a similar safe guard for compatibility. After that is added I'd say this is ready to be merged!

cedi: good idea. I'm on it

cedi: I just realised you actually made the mistake you warned about [here](https://github.com/mozilla/sops/pull/966#discussion_r830294838) and why you suggested making `MasterKeyFromRecipient()` private. It's not `MasterKeyFromRecipient()` but `MasterKeysFromRecipients()` 😄 

The naming on those functions is really unfortunate. I think making the "singular" version private was the right call!

ajvb (approved):

PR #968, opened 2021-12-14
README: clarify default location of the age keys

rombert:
According to https://github.com/mozilla/sops/blob/adfe49c1ead94808a2f06b6cf65758434238e00f/age/keysource.go#L108,
the directory structure is userConfigDir/sops/age .


ajvb:
This is fixed in `develop`. Thank you tho!

Issue #543, opened 2019-10-02
Add ability to strip encrypted or unencrypted suffix on decryption


kristaxox:
This would allow the use of sops without needing to update key names in the areas that they are used to have `_encrypted` or `_unencrypted`.

Encrypted example:
```
db:
    password_encrypted: ENC[...]
```

Upon decryption with a "strip suffix option":
```
db:
    password: pw123
```

endorama:
Following https://github.com/mozilla/sops/issues/539 I'm going to work on this!

autrilla:
I'm with @jvehent here https://github.com/mozilla/sops/issues/539#issuecomment-533555867. I'm not sure this is SOPS's job.

It's not that I don't think your use case is valid. As maintainers, I think we have to be careful what features we allow into our code base, since once they're in there, we basically have to support them forever.

endorama:
I do understand that, but requiring everyone using that feature to have a wrapper script to remove those suffix if very painful for the users. As `sops` is already performing such a removal for key metadata, I would expect it to remove also the suffix (I would agree with @davidovich https://github.com/mozilla/sops/issues/539#issuecomment-533559825 in considering that a sops level metadata). 

> As maintainers, I think we have to be careful what features we allow into our code base,

As I'm not a maintainer, is not in my interest to just push some code without approval, but if your fear the maintenance costs may be worth discussing this looking at a possible implementation? 🙂

autrilla:
I will gladly look at it, but I can make no guarantee that it will be merged.

jvehent:
I understand the temptation to bake every use case into a single tool, but I really think you should follow the unix approach here, and make a simple separate tool that edits your yaml files the way you want.

Sops is a security tool, and every non-security use-case we support into it makes it larger and more difficult to audit.

endorama:
> Sops is a security tool, and every non-security use-case we support into it makes it larger and more difficult to audit.

I would agree if this feature was not already present, in this case I believe is "partially" present: this missing piece makes using this feature a burden in different use cases.  
I do agree that a reduced surface is good for security but IMO security tools should also be easy, intuitive and functional to use.

I'd like to highlight a couple of use cases that are currently only achievable with `encrypted-regex` or cannot be achieved at all.  
(I would prefer *not* to require my team to use regex, they are brittle and fragile)

[...]

PR #974, opened 2021-12-19
Support using comments to select parts to encrypt

mitar:
This PR adds support to annotate comments with a string (e.g., `sops:enc`) which can then be matched with a regex. If it matches, the corresponding value (the one which follows the comment) is encrypted while other values are not. (There is also the opposite regex available, to select those values which should not be encrypted.)

This enables the YAML file to have the same structure encrypted and decrypted, without having to add suffixes or manage complex regexes to match keys. See #543 for more discussion.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1438,9 +1438,18 @@ that match the supplied regular expression. For example, this command:
 will not encrypt the values under the ``description`` and ``metadata`` keys in a YAML file
 containing kubernetes secrets, while encrypting everything else.
 
+For YAML files, another method is to use ``--encrypted-comment-regex`` which will
+only encrypt comments and values which have a preceding comment matching the supplied
+regular expression.
+
+Conversely, you can opt in to only left certain keys without encrypting by using the
+``--unencrypted-comment-regex`` option, which will leave the values and comments
+unencrypted when they have a preeceding comment that matches the supplied regular expression.
+
 You can also specify these options in the ``.sops.yaml`` config file.
 
-Note: these four options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--encrypted-regex`` and ``--unencrypted-regex`` are
+Note: these six options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--encrypted-regex``,
+``--unencrypted-regex``, ``--encrypted-comment-regex``, and ``--unencrypted-comment-regex`` are
 mutually exclusive and cannot all be used in the same file.
 
 Encryption Protocol
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -685,6 +685,14 @@ func main() {
 			Usage: "set the encrypted key regex. When specified, only keys matching the regex will be encrypted.",
 		},
 		cli.StringFlag{
+			Name:  "unencrypted-comment-regex",
+			Usage: "set the unencrypted comment suffix. When specified, only keys that have comment matching the regex will be left unencrypted.",
+		},
+		cli.StringFlag{
+			Name:  "encrypted-comment-regex",
+			Usage: "set the encrypted comment suffix. When specified, only keys that have comment matching the regex will be encrypted.",
+		},
+		cli.StringFlag{
 			Name:  "config",
 			Usage: "path to sops' config file. If set, sops will not search for the config file recursively.",
 		},
@@ -738,6 +746,8 @@ func main() {
 		encryptedSuffix := c.String("encrypted-suffix")
 		encryptedRegex := c.String("encrypted-regex")
 		unencryptedRegex := c.String("unencrypted-regex")
+		encryptedCommentRegex := c.String("encrypted-comment-regex")
+		unencryptedCommentRegex := c.String("unencrypted-comment-regex")
 		conf, err := loadConfig(c, fileName, nil)
 		if err != nil {
 			return toExitError(err)
@@ -756,6 +766,12 @@ func main() {
 			if unencryptedRegex == "" {
 				unencryptedRegex = conf.UnencryptedRegex
 			}
+			if encryptedCommentRegex == "" {
+				encryptedCommentRegex = conf.EncryptedCommentRegex
+			}
+			if unencryptedCommentRegex == "" {
+				unencryptedCommentRegex = conf.UnencryptedCommentRegex
+			}
 		}
 
 		cryptRuleCount := 0
@@ -771,12 +787,18 @@ func main() {
 		if unencryptedRegex != "" {
 			cryptRuleCount++
 		}
-
-		if cryptRuleCount > 1 {
-			return common.NewExitError("Error: cannot use more than one of encrypted_suffix, unencrypted_suffix, encrypted_regex or unencrypted_regex in the same file", codes.ErrorConflictingParameters)
+		if encryptedCommentRegex != "" {
+			cryptRuleCount++
+		}
+		if unencryptedCommentRegex != "" {
+			cryptRuleCount++
 		}
 
-		// only supply the default UnencryptedSuffix when EncryptedSuffix and EncryptedRegex are not provided
+		if cryptRuleCount > 1 {
+			return common.NewExitError("Error: cannot use more than one of encrypted_suffix, unencrypted_suffix, encrypted_regex, unencrypted_regex, encrypted_comment_regex, or unencrypted_comment_regex in the same file", codes.ErrorConflictingParameters)
+		}
+
+		// only supply the default UnencryptedSuffix when EncryptedSuffix, EncryptedRegex, and others are not provided
 		if cryptRuleCount == 0 {
 			unencryptedSuffix = sops.DefaultUnencryptedSuffix
 		}
@@ -798,17 +820,19 @@ func main() {
 				return toExitError(err)
 			}
 			output, err = encrypt(encryptOpts{
-				OutputStore:       outputStore,
-				InputStore:        inputStore,
-				InputPath:         fileName,
-				Cipher:            aes.NewCipher(),
-				UnencryptedSuffix: unencryptedSuffix,
-				EncryptedSuffix:   encryptedSuffix,
-				UnencryptedRegex:  unencryptedRegex,
-				EncryptedRegex:    encryptedRegex,
-				KeyServices:       svcs,
-				KeyGroups:         groups,
-				GroupThreshold:    threshold,
+				OutputStore:             outputStore,
+				InputStore:              inputStore,
+				InputPath:               fileName,
+				Cipher:                  aes.NewCipher(),
+				UnencryptedSuffix:       unencryptedSuffix,
+				EncryptedSuffix:         encryptedSuffix,
+				UnencryptedRegex:        unencryptedRegex,
+				EncryptedRegex:          encryptedRegex,
+				UnencryptedCommentRegex: unencryptedCommentRegex,
+				EncryptedCommentRegex:   encryptedCommentRegex,
+				KeyServices:             svcs,
+				KeyGroups:               groups,
+				GroupThreshold:          threshold,
 			})
 		}
 
@@ -953,13 +977,15 @@ func main() {
 					return toExitError(err)
 				}
 				output, err = editExample(editExampleOpts{
-					editOpts:          opts,
-					UnencryptedSuffix: unencryptedSuffix,
-					EncryptedSuffix:   encryptedSuffix,
-					UnencryptedRegex:  unencryptedRegex,
-					EncryptedRegex:    encryptedRegex,
-					KeyGroups:         groups,
-					GroupThreshold:    threshold,
+					editOpts:                opts,
+					UnencryptedSuffix:       unencryptedSuffix,
+					EncryptedSuffix:         encryptedSuffix,
+					UnencryptedRegex:        unencryptedRegex,
+					EncryptedRegex:          encryptedRegex,
+					UnencryptedCommentRegex: unencryptedCommentRegex,
+					EncryptedCommentRegex:   encryptedCommentRegex,
+					KeyGroups:               groups,
+					GroupThreshold:          threshold,
 				})
 			}
 		}
--- a/config/config.go
+++ b/config/config.go
@@ -109,20 +109,22 @@ type destinationRule struct {
 }
 
 type creationRule struct {
-	PathRegex         string `yaml:"path_regex"`
-	KMS               string
-	AwsProfile        string `yaml:"aws_profile"`
-	Age               string `yaml:"age"`
-	PGP               string
-	GCPKMS            string     `yaml:"gcp_kms"`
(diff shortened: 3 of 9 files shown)


cmd/sops/main.go
```diff
@@ -666,11 +666,19 @@ func main() {
 		},
 		cli.StringFlag{
 			Name:  "unencrypted-regex",
-			Usage: "set the unencrypted key suffix. When specified, only keys matching the regex will be left unencrypted.",
+			Usage: "set the unencrypted key regex. When specified, only keys matching the regex will be left unencrypted.",
```
mitar, line 669: This is unrelated, but I noticed it and fixed.

config/config.go
```diff
@@ -241,12 +245,21 @@ func configFromRule(rule *creationRule, kmsEncryptionContext map[string]*string)
 	if rule.EncryptedSuffix != "" {
 		cryptRuleCount++
 	}
+	if rule.UnencryptedRegex != "" {
```
mitar, line 248: This is unrelated, but I noticed it and fixed.

config/config_test.go
```diff
@@ -396,17 +419,29 @@ func TestLoadConfigFileWithEncryptedSuffix(t *testing.T) {
...
 func TestLoadConfigFileWithUnencryptedRegex(t *testing.T) {
-	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithRegexParameters, t), "/conf/path", "barbar", nil)
+	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithUnencryptedRegexParameters, t), "/conf/path", "barbar", nil)
 	assert.Equal(t, nil, err)
 	assert.Equal(t, "^dec:", conf.UnencryptedRegex)
 }
 
 func TestLoadConfigFileWithEncryptedRegex(t *testing.T) {
-	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithRegexParameters, t), "/conf/path", "barbar", nil)
+	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithEncryptedRegexParameters, t), "/conf/path", "barbar", nil)
```
mitar, line 428: This is unrelated, but I noticed it and fixed.

mitar:
Rebased this as well.

mitar:
Rebased to latest develop.

mitar:
I rebased to latest main.

Gui13:
Hey guys, we are interested in the merging of this PR, it solves a long-standing issue of the `--encrypted-suffix` which requires us to perform `sed -i 's/encrypted_suffix//g'` on all our files after sops decryption.

The linked issue, #543, is 4 years old, and this change elegantly solves the issue.

Is there something I can do to help on this?

hiddeco:
This did not make the cut for `v3.8.0`, as this version will contain many changes already (especially in the area of rewriting all key source implementations to their latest API clients). Making it risky to add much more on top of it. It is however scheduled to be looked at for `v3.9.0`.

cmd/sops/main.go
felixfontein: Could you please move all unrelated fixes to another PR?

mitar: Done: https://github.com/getsops/sops/pull/1300

config/config.go
mitar: Moved to https://github.com/getsops/sops/pull/1300.

config/config_test.go
mitar: Moved to https://github.com/getsops/sops/pull/1300.

mitar:
This PR now depends on https://github.com/getsops/sops/pull/1300.

sops.go
```diff
@@ -242,21 +242,30 @@ func (branch TreeBranch) walkValue(in interface{}, path []string, onLeaves func(
...
-		newV, err := branch.walkValue(v, path, onLeaves)
+		if c, ok := v.(Comment); ok {
+			commentsStack[len(commentsStack)-1] = append(commentsStack[len(commentsStack)-1], c.Value)
+		}
+		newV, err := branch.walkValue(v, path, commentsStack, onLeaves)
 		if err != nil {
 			return nil, err
 		}
 		in[i] = newV
+		if _, ok := v.(Comment); !ok {
```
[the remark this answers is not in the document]
mitar: OK.

sops.go
```diff
@@ -270,79 +279,127 @@ func (branch TreeBranch) walkBranch(in TreeBranch, path []string, onLeaves func(
...
 			return nil, fmt.Errorf("Tree contains a non-string key (type %T): %s. Only string keys are"+
 				"supported", item.Key, item.Key)
 		}
-		newV, err := branch.walkValue(item.Value, append(path, key), onLeaves)
+		newV, err := branch.walkValue(item.Value, append(path, key), commentsStack, onLeaves)
 		if err != nil {
 			return nil, err
 		}
 		in[i].Value = newV
+		if _, ok := item.Value.(Comment); !ok {
```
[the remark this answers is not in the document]
mitar: OK.

sops.go
```diff
@@ -242,21 +242,30 @@ func (branch TreeBranch) walkValue(in interface{}, path []string, onLeaves func(
...
 }
 
-func (branch TreeBranch) walkBranch(in TreeBranch, path []string, onLeaves func(in interface{}, path []string) (interface{}, error)) (TreeBranch, error) {
+func (branch TreeBranch) walkBranch(in TreeBranch, path []string, commentsStack [][]string, onLeaves func(in interface{}, path []string, commentsStack [][]string) (interface{}, error)) (TreeBranch, error) {
+	commentsStack = append(commentsStack, []string{})
 	for i, item := range in {
-		if _, ok := item.Key.(Comment); ok {
-			enc, err := branch.walkValue(item.Key, path, onLeaves)
+		if c, ok := item.Key.(Comment); ok {
+			commentsStack[len(commentsStack)-1] = append(commentsStack[len(commentsStack)-1], c.Value)
```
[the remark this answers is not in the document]
mitar: This does not have to be undone. Stack is not modified because append returns a slice copy.

What we are doing here is collecting which comments are active.

I will add comments to explain code a bit, it seems it could be useful.

mitar:
@felixfontein: I addressed all review comments. I moved unrelated changes to a separate commit which is also part of https://github.com/getsops/sops/pull/1300, so once that is merged it will not be in this PR anymore.

I also rebased to the current main branch.

mitar:
I rebased after #1300 was merged.

README.rst
```diff
@@ -1438,9 +1438,18 @@ that match the supplied regular expression. For example, this command:
 will not encrypt the values under the ``description`` and ``metadata`` keys in a YAML file
 containing kubernetes secrets, while encrypting everything else.
 
+For YAML files, another method is to use ``--encrypted-comment-regex`` which will
+only encrypt comments and values which have a preceding comment matching the supplied
+regular expression.
+
+Conversely, you can opt in to only left certain keys without encrypting by using the
+``--unencrypted-comment-regex`` option, which will leave the values and comments
+unencrypted when they have a preeceding comment that matches the supplied regular expression.
```
felixfontein, line 1447: The comment can also be on the same line, like `foo: bar # ENC`. (Which is the case because that comment is moved before `x: y`.)

felixfontein (commented):
I noticed a problem: when I store this as `x.yml`
```
# x
x: x
```
and run
```
sops --encrypt --unencrypted-comment-regex ENC x.yml | sops --decrypt --input-type yaml --output-type yaml /dev/stdin
```
I get
```
MAC mismatch. File has A4ABD4448C49562D828115D13A1FCCEA927F52B4D5459297F8B43E42DA89238BC13626E43DCB38DDB082488927EC904FB42057443983E88585179D50551AFE62, computed 9D5B4CB3A99652A79C6149B4C1B9F17CE15E5BBB5E13D2054610BDE0C7BE57B58CAEE9B503F432AF0F27F857D26D549494D2D86CF3E68F3DA78554089C86CD8F
```
This is caused because `ENC` matches the encrypted comment.

I'm not really sure what to do here. Potentially `shouldBeEncrypted` should know whether it is decrypting, and if it is, and the regex for the comment matches, it should first check whether the value looks like an encrypted string (https://github.com/getsops/sops/blob/main/aes/cipher.go#L186) before accepting the regex match. And at the same time, when it encounters such comments (that match the regex) during encryption, it should reject the file. Does that sound reasonable? Do you have another idea?

mitar:
Thank you for the review.

> This is caused because ENC matches the encrypted comment.

Nice catch. I didn't really expect for people to use such simple regexes. Personally I use `sops:enc` which does not seem to have the issue you are describing (it cannot appear in encoded string).

Maybe we should just document this issue and suggest a regex (e.g., `sops:enc`)? So documentation could be something like: "do not pick a regex which can match encrypted values" (with link to the format of encrypted values)?

But it is not really nice for the security tool like this one to be able to be misconfigured.

> Potentially shouldBeEncrypted should know whether it is decrypting, and if it is, and the regex for the comment matches, it should first check whether the value looks like an encrypted string.

I think we should do a check like:
* Try to match the comment with encryptedValueRegexp (matching the structure of encrypted values) and with given regexp.
* If only one of them match, good, we know what to do.
* If both match, then we transform the comment by removing content matched by the encryptedValueRegexp and try to match given regexp again.
* If it still matches, good, we know what to do.
* If it does not match, we abort and complain that the regexp is too broad.

I think this is slightly better in sense that it informs the user about misconfiguration and that they should fix the regexp.

> And at the same time, when it encounters such comments (that match the regex) during encryption, it should reject the file.

So during encryption, every time we encrypt any comment, we try to match the result with the given regexp, and if it does match the regexp, we abort and complain that the regexp is too broad. If I understand it correctly, I think this is fine.

So we should just make sure to help the user pick a good regexp by detecting too broad (or should we way "too simple"?) regexp and complain in such case.

mitar:
@felixfontein Do you agree with my proposal above?

felixfontein:
Hmm, reading all this again after some time mainly shows me that this is something we have to be very, very careful with.

One other solution that came to my mind when looking at this again: during encryption, once a comment is encrypted, check whether the encrypted comment matches `UnencryptedCommentRegex` or `EncryptedCommentRegex` (when provided). If any of them matches, reject the whole file.

mitar:
> One other solution that came to my mind when looking at this again: during encryption, once a comment is encrypted, check whether the encrypted comment matches UnencryptedCommentRegex or EncryptedCommentRegex (when provided). If any of them matches, reject the whole file.

Hm, isn't this the same as we have been discussing already? I wrote:

> So during encryption, every time we encrypt any comment, we try to match the result with the given regexp, and if it does match the regexp, we abort and complain that the regexp is too broad. If I understand it correctly, I think this is fine.

So I think this is great. We seems to be in agreement what to do when **encrypting**. Just check if the result matches the regexp and abort if it does.

But I think the question is what to do when **decrypting**. So I can imagine a scenario where you encrypted only few lines with a comment like `sops:enc`. Great. But then in the encrypted file you decide to change `sops:enc` to `enc` comment. And then you try to decrypt that with an updated CLI argument. What should happen? This is why I proposed:

> I think we should do a check like:
>
> * Try to match the comment with encryptedValueRegexp (matching the structure of encrypted values) and with given regexp.
> * If only one of them match, good, we know what to do.
> * If both match, then we transform the comment by removing content matched by the encryptedValueRegexp and try to match given regexp again.
 > * If it still matches, good, we know what to do.
 > * If it does not match, we abort and complain that the regexp is too broad.
>
> I think this is slightly better in sense that it informs the user about misconfiguration and that they should fix the regexp.

Maybe a simpler way would be to store (and MAC) used regexp to encrypt into config section of the file. And when decrypting only use the regexp from the config section. So if you want to change the regexp/comment, you have to decrypt and re-encrypt?

felixfontein:
> > One other solution that came to my mind when looking at this again: during encryption, once a comment is encrypted, check whether the encrypted comment matches UnencryptedCommentRegex or EncryptedCommentRegex (when provided). If any of them matches, reject the whole file.
> 
> Hm, isn't this the same as we have been discussing already? I wrote:
> 
> > So during encryption, every time we encrypt any comment, we try to match the result with the given regexp, and if it does match the regexp, we abort and complain that the regexp is too broad. If I understand it correctly, I think this is fine.
>
> So I think this is great. We seems to be in agreement what to do when **encrypting**. Just check if the result matches the regexp and abort if it does.

Yes, this part we agree upon.

> But I think the question is what to do when **decrypting**. So I can imagine a scenario where you encrypted only few lines with a comment like `sops:enc`. Great. But then in the encrypted file you decide to change `sops:enc` to `enc` comment. And then you try to decrypt that with an updated CLI argument. What should happen?

IMO: if you modify the encrypted file so it won't decrypt anymore, it's your own fault if it fails.

> This is why I proposed:
> 
> > I think we should do a check like:
> > 
> > * Try to match the comment with encryptedValueRegexp (matching the structure of encrypted values) and with given regexp.
> > * If only one of them match, good, we know what to do.
> > * If both match,

In case both match, I would simply error out. The user did something wrong (assuming we don't have a bug) and they need to fix it manually. TBH I would simply check the encryptedValueRegexp regular expression, and if it matches, assume that the comment is encrypted (and not even check the other regexp).

> >  then we transform the comment by removing content matched by the encryptedValueRegexp and try to match given regexp again.
> > * If it still matches, good, we know what to do.
> > * If it does not match, we abort and complain that the regexp is too broad.
> >
> > I think this is slightly better in sense that it informs the user about misconfiguration and that they should fix the regexp.

I would avoid that. It makes decryption less efficient (two regular expressions to test instead of potentially only one per comment), and increases complexity for a situation that should not arise if users do not modify sops encrypted files by hand.

> Maybe a simpler way would be to store (and MAC) used regexp to encrypt into config section of the file.

Right now the MAC only checks the message, and not the metadata. I would avoid extending it. We could add another MAC which only covers metadata, but also that is something that should not happen in this PR.

> And when decrypting only use the regexp from the config section.

That's what should happen anyway.

> So if you want to change the regexp/comment, you have to decrypt and re-encrypt? In general it might be nice user experience to store all those CLI config switches into config section so that you do not have to specify them again when decrypting (and match them exactly, and know what was used when encrypting).

If you have to specify any switch again when decrypting, it is a bug that needs to be fixed ASAP. The only options that should affect decryption are general things like `--output`, `--output-type`, `--ignore-mac`.

felixfontein:
I rebased this PR to do some (unrelated) work on top of it; the resulting commit is contained in #1387.

mitar:
Thanks. I will look into this again after holidays.

mitar:
BTW, if you are currently working on this and you are mentally in this logic and willing, feel free to wrap up with this PR yourself. If it is easier for you.

felixfontein:
Not right now, but I might try to invest more time in this during the next days.

felixfontein:
I'll work on it in #1392.

mitar:
This is continued in #1392.

PR #975, opened 2021-12-20
perform de/encryptWithGPGBinary before de/encryptWithCryptoOpenPGP

karlschriek:
First decrypt/encrypt from local GPG binary before attempting to contact the (deprecated) OpenPGP server

Addresses mozilla#749 and mozilla#902.


colemickens:
Thanks for this; a real shame it can't be merged. I couldn't figure out why re-encrypting my secrets was taking forever until I did it one day without an Internet connection. Hard to even understand why this patch was even necessary instead of being the original implementation behavior.

PR #976, opened 2021-12-21
Enable usage of vault's TokenHelpers

pruiz:
Added support for vault's token helpers (https://www.vaultproject.io/docs/commands/token-helper), as exposed by vault api. This allows working against multiple vault instances, obtaining a valid token depending on the address of the vault client being requested.

### DIFF
--- a/hcvault/keysource.go
+++ b/hcvault/keysource.go
@@ -128,25 +128,29 @@ func vaultClient(address string) (*api.Client, error) {
 	if cli.Token() != "" {
 		return cli, nil
 	}
-	homePath, err := homedir.Dir()
+	// TokenHelper requires VAULT_ADDR available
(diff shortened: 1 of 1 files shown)


rkusler-intermedia:
This would be a useful feature

rosskusler:
@pruiz I'm using a custom build of sops with your patch applied too and would definitely be interested in this being added upstream :)

pruiz:
Hi, I was wondering if updating this PR would make sense for SOPS mantainers to be included in a future release?

felixfontein:
> Hi, I was wondering if updating this PR would make sense for SOPS mantainers to be included in a future release?

@getsops/maintainers ^ any ideas? I'm not familiar with HashiVault / OpenBao.

PR #992, opened 2022-01-30
missing `-e` on line 192 when introducing `--age`

calebTree:
I added `--encrypt` so that the --age introduction command is correct


calebTree:
this is wrong
what it really needs is to remove the redirect at the end
either way hopefully this line can be corrected

ajvb:
Apologies, this is fixed in the `develop` branch, but has yet to be merged into `master`.

PR #999, opened 2022-02-08
upgrades deps

dani-santos-code:
upgrades depenedencies


dani-santos-code:
this was a mistake. sorry

PR #1008, opened 2022-02-15
Fix flakey test run on OSX

acastle:
Docker Desktop for OSX has special handling for `localhost` for forwarding traffic to the internal docker VM (specifically for localhost and not `127.0.0.1`). Using `resource.GetHostPort` will return an OS appropriate host and eliminate some the flaky test behavior when running this on other operating systems.


ajvb (approved):

ajvb:
Thank you @acastle!

Issue #870, opened 2021-05-14
Docker image no longer be bumped by CI


anthr76:
Hello today I went to pull the latest version of SOPS though I noticed that the last time a image was pushed was 8 months ago. Assuming there's a problem with the CI? Is this intentional?

ikedam:
Available
* v3.6.1 (15 Sep 2020)

Not available
* v3.7.0 (25 Mar 2021)
* v3.7.1 (9 Apr 2021)

ikedam:
The alpine version image looks unable to be built:
```
$ docker build -f Dockerfile.alpine -t mozilla/sops:alpine .
...
#12 67.90 # filippo.io/age/internal/bech32
#12 67.90 /go/pkg/mod/filippo.io/age@v1.0.0-beta7/internal/bech32/bech32.go:41:15: invalid operation: top >> i (shift count type int, must be unsigned integer)
#12 67.90 /go/pkg/mod/filippo.io/age@v1.0.0-beta7/internal/bech32/bech32.go:74:20: invalid operation: mod >> shift (shift count type int, must be unsigned integer)
#12 67.90 note: module requires Go 1.13
#12 113.4 make: *** [Makefile:13: install] Error 2
------
executor failed running [/bin/sh -c CGO_ENABLED=1 make install]: exit code: 2
```

ikedam:
Created: #905

jvosantos:
There's a pull request pending since august, any updates?

ikedam:
This is fixed with #1012.

ikedam:
(But not released yet)

PR #1018, opened 2022-03-05
Build alpine container in CI (#870)

ikedam:
Docker containers are no longer available in dockerhub since v3.7.0.
It's caused for the alpine version container uses go-1.12 and fails to be built.
It's fixed in #1012.

I want to let CI test building alpine containers so that the problem won't be reproduced.


ajvb (approved):

ikedam:
The CI failure will be fixed in #1032.

Issue #1022, opened 2022-03-09
Problem with building the alpine container in circle


ajvb:
https://app.circleci.com/pipelines/github/mozilla/sops/128/workflows/64607008-01d6-4551-9bd0-b1181fdb2f40/jobs/431

FnTm:
@ajvb looks like an addition instead of a change snuck into the Alpine Dockerfile update recently: https://github.com/mozilla/sops/commit/dc2267e3721a1037895050f4fd02974a31692ded#diff-f865864730d7062c04cc13b1b85ba03ed5dbef3fa02fcb50b87d9117d0af1f4eR2 . My guess is that this is what is breaking the build.

ajvb:
Fixed by #1025

PR #1025, opened 2022-03-09
Remove duplicated stage from Dockerfile.alpine

shouichi:
Building image fails with the following error.

```
$ docker build -t sops -f Dockerfile.alpine .
Sending build context to Docker daemon  884.2kB
Step 1/11 : FROM golang:1.12-alpine3.10 AS builder
 ---> 6d6865cf688e
builder stage name already used
```


FnTm:
This should be a fix for #1022 .

ajvb (approved):

PR #1026, opened 2022-03-10
Explicitly build linux amd64 binary

### LINKED ISSUE #1023: Latest (3.7.2) linux amd64 binary not working
```./sops-v3.7.2.linux.amd64
-bash: ./sops-v3.7.2.linux.amd64: cannot execute binary file: Exec format error
```

Did something go wrong in the CI process? macOS binary is completely fine.

FnTm:
Previous setup relied implicitly of the correct file to be there. Introduction of arm64 builds broke that implicit assumption.

Fixes #1023


FnTm:
@ajvb could I get a review please? This should fix #1023 where there was an implicit dependency in the release file which I didn't know about.

ajvb (approved):

Issue #969, opened 2021-12-14
Cannot create GCP KMS when using Google Workload Identity Federation


kraxx:
I'm authenticating and authorizing a GitHub Action workflow to impersonate a GCP service account using [Workload Identity Federation](https://cloud.google.com/iam/docs/workload-identity-federation), which has permissions to decrypt a SOPS-encrypted file. I'm using https://github.com/google-github-actions/auth to do so, which exports the `GOOGLE_APPLICATION_CREDENTIALS` environment variable and sets it to a generated credentials file. However, when it comes to using `sops` to decrypt the file, I encounter this error:
```
Failed to get the data key required to decrypt the SOPS file.

Group 0: FAILED
  <*** path to GCP keyring ***>: FAILED
    - | Cannot create GCP KMS service: google: error getting
      | credentials using GOOGLE_APPLICATION_CREDENTIALS environment
      | variable: unknown credential type: "external_account"
```

Indeed, this credential file is of type `external_account`, not `service_account` which would normally work.

The error is generated [here](https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/gcpkms/keysource.go#L139). Looking further into this, I see [the function is deprecated](https://pkg.go.dev/google.golang.org/api/cloudkms/v1#New), and on top of that, the version of the [pkg.go.dev/google.golang.org/api](https://pkg.go.dev/google.golang.org/api@v0.63.0) package used in `sops` is severely outdated.

My suspicion lays in the outdated google api package which may not support workload identity federation, but of course there may be other reason(s) for why it's not supported.

enchorb:
Yup, getting this same issue - seems if the sops team upgrades to the latest oath2 library this can be resolved? 

ie:  https://github.com/googleapis/repo-automation-bots/issues/2752
ie:  https://github.com/GoogleCloudPlatform/docker-credential-gcr/issues/95

soggycactus:
I have this issue as well - can we get an update about how to resolve this issue? I'd really rather not generate a JSON key as it's less secure than using Workload Identity

zmitry:
any updates on this?

sean-morris:
Same issue here as well - thanks for the help

j-bro:
@kraxx The dependencies seem to have been updated in #1024, worth another look to see if this fixes the issue with Workload Identity Federation.

zmitry:
on 3.7.2 still have this bug
[...]

PR #1030, opened 2022-03-18
Support Workload Identity Federation

duythinht:
A lot of people trying to auth with google kms to get decryption key by Workload Identity Federation, but old oauth2 lib from 2019 does not support this.
Ref: https://github.com/mozilla/sops/issues/969
This PR just simple, upgrade the lib to current oauth2 version, and make it work with WIF.


ajvb:
Thank you for the PR @duythinht - but this was already completed in #1024

PR #1039, opened 2022-04-11
Fix security vulernabilities by bumping dependency versions

andrei-radulescu:
* CVE-2020-28852
* CVE-2022-27191
* CVE-2020-28852
* CVE-2022-27191


ajvb:
Thank you for the PR but these are already resolved in `develop`.

PR #1064, opened 2022-05-27
age: improve identity loading, add tests, tidy

hiddeco:
This adds improvements to identity loading, extensive test coverage
and a general tidying of bits of code. The improvements are based on a
fork of the age key source in the [Flux project's kustomize-controller](https://github.com/fluxcd/kustomize-controller/tree/327a3560b35c5994647938958aa5523f319094b8/internal/sops/age),
which was built due to SOPS' limitations around identity management
without relying on runtime environment variables.

- It introduces a `ParsedIdentity` type which contains a slice of age
  identities, and can be applied to the `MasterKey`. When applied,
  further loading of identities from the runtime environment is skipped
  for `Decrypt` operations. This is most useful when working with SOPS
  as an SDK, in combination with e.g. a local key service server
  implementation.
- The `Identity` field has been deprecated in the `MasterKey` struct.
  Presence of the field was misleading, as it is not actually used.
- Any detected identity reference is now loaded, instead of it assuming
  a priority order. This makes more sense, as age is able to work with
  a set of loaded identities. If no environment variables are defined,
  the existence of the keys.txt in the user's config directory is
  required.
- Decrypt logs have been added to match other key sources.
- Extensive test coverage.


age/keysource.go
```diff
@@ -169,55 +210,84 @@ func (key *MasterKey) ToString() string {
...
-		key, err := masterKeyFromRecipient(recipient)
+	if ageKey, ok := os.LookupEnv(SopsAgeKeyEnv); ok {
+		readers[SopsAgeKeyEnv] = strings.NewReader(ageKey)
+	}
 
+	if ageKeyFile, ok := os.LookupEnv(SopsAgeKeyFileEnv); ok {
+		f, err := os.Open(ageKeyFile)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to open %s file: %w", SopsAgeKeyFileEnv, err)
```
hiddeco, line 233: The other option would be to log any loading failure, and only error if `identities == 0`.

ajvb: I think this way is preferable. A bit more defense around user mistakes, especially in automation using sops.

ajvb (approved):

ajvb:
This is awesome, thank you @hiddeco!

PR #1072, opened 2022-06-07
gcpkms: update SDK to latest, add tests, tidy

hiddeco:
This PR updates the GCP KMS client to latest, adds more extensive test
coverage, and general tidying of bits of code.

The improvements are based on a fork of the key source in the Flux
project's kustomize-controller, built due to SOPS' limitation around
credential management without relying on runtime environment variables.

- Updates the deprecated `google.golang.org/api/cloudkms/v1` to
  `cloud.google.com/go/kms/apiv1`.
- It introduces a `CredentialJSON` type which holds a Service Account
  credential file, and can be applied to the `MasterKey`.
  When applied, the provided credentials are used in the GCP KMS
  service client configuration, instead of relying on
  `GOOGLE_CREDENTIALS`, or the default client environment variables.
  This is most useful when working with SOPS as an SDK, in combination
  with e.g. a local key service server implementation.
- Test coverage.

The forked version of this has compatability (and integration) tests to
ensure it works with current SOPS:

- https://github.com/fluxcd/kustomize-controller/blob/cbb0fc9df5da12a54cf998392dbd80d28522def5/internal/sops/gcpkms/keysource_integration_test.go#L39
- https://github.com/fluxcd/kustomize-controller/blob/cbb0fc9df5da12a54cf998392dbd80d28522def5/internal/sops/gcpkms/keysource_integration_test.go#L59


gcpkms/mock_kms_server_test.go
```diff
@@ -0,0 +1,327 @@
+// Copyright 2019 Google LLC
```
hiddeco, line 1: @ajvb copyright around this can not be avoided, it is copied from the tests of the SDK (with some changes to drop deprecated dependencies).

gcpkms/keysource.go
```diff
@@ -1,167 +1,230 @@
...
+	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
+	"google.golang.org/grpc"
+)
 
-	"github.com/sirupsen/logrus"
-	"golang.org/x/net/context"
-	cloudkms "google.golang.org/api/cloudkms/v1"
+const (
+	// SopsGoogleCredentialsEnv can be set as an environment variable to
+	// SopsGpgExecEnv can be set as an environment variable as either a path to
```
ajvb, line 22: The verbiage in this comment block seems off.

ajvb (approved):

PR #1080, opened 2022-06-22
Simplify AWS KMS Key ARN regexp

mhmxs:
The numeric range should replace to any digit.


ajvb:
@mhmxs Thank you for the PR but this implementation is changing.

PR #1085, opened 2022-07-05
hcvault: update API, add tests, tidy

hiddeco:
This PR updates the Vault API and client to latest, adds more extensive
test coverage, and general tidying of bits of code.
The improvements are based on a fork of the key source in the Flux
project's kustomize-controller, built due to SOPS' limitation around
credential management without relying on runtime environment variables.

- Vault API and client have been updated to latest.
- It introduces a `Token` type which holds a Vault token, and can be
  applied to the `MasterKey`. When applied, the token is used in the
  Vault client configuration, instead of relying on the `VAULT_TOKEN`
  environment variables, or the `.vault-token` file in the user's home
  directory. This is most useful when working with SOPS as an SDK, in
  combination with e.g. a local key service server implementation.
- Extensive test coverage.

The forked version of this has compatability tests to ensure it works
with current SOPS:

- https://github.com/fluxcd/kustomize-controller/blob/62fb2d96a297c7e344050f46ee54074ef66dd438/internal/sops/hcvault/keysource_test.go#L130
- https://github.com/fluxcd/kustomize-controller/blob/62fb2d96a297c7e344050f46ee54074ef66dd438/internal/sops/hcvault/keysource_test.go#L202


ajvb:
🎉

PR #1104, opened 2022-08-08
Allow read data from stdin on Windows

### LINKED ISSUE #739: Piping using `/dev/stdin` doesn't work on Windows
The "workaround" to get piping to work is not cross-platform.

- https://stackoverflow.com/questions/7395157/windows-equivalent-to-dev-stdin

I highly recommend handling it like `kubectl` and other tools where you can pass in `-` as file name which triggers reading from `stdin`.

So something like `echo "asdf" | sops -` or `echo "asdf" | sops -f -`

Is this something you would accept a PR for?

jkroepke:
Fixes #739

Allow `-` as cross platform alternative to `/dev/stdin` for reading data from STDIN.

```
% cat secrets.yaml | ~/go/src/go.mozilla.org/sops/cmd/sops/sops --input-type yaml --output-type yaml -d -
podAnnotations:
    secret: value
% ~/go/src/go.mozilla.org/sops/cmd/sops/sops -d secrets.yaml
podAnnotations:
    secret: value
```

### DIFF
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -129,7 +129,7 @@ func EncryptTree(opts EncryptTreeOpts) error {
 func LoadEncryptedFile(loader sops.EncryptedFileLoader, inputPath string) (*sops.Tree, error) {
 	fileBytes, err := ReadFile(inputPath)
 	if err != nil {
-		return nil, NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
+		return nil, err
 	}
 	path, err := filepath.Abs(inputPath)
 	if err != nil {
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -127,7 +127,7 @@ func EncryptTree(opts EncryptTreeOpts) error {
 
 // LoadEncryptedFile loads an encrypted SOPS file, returning a SOPS tree
 func LoadEncryptedFile(loader sops.EncryptedFileLoader, inputPath string) (*sops.Tree, error) {
-	fileBytes, err := ioutil.ReadFile(inputPath)
+	fileBytes, err := ReadFile(inputPath)
 	if err != nil {
 		return nil, NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
 	}
@@ -440,4 +440,20 @@ func PrettyPrintDiffs(diffs []Diff) {
 			color.New(color.FgRed).Printf("--- %s\n", c.ToString())
 		}
 	}
+}
+
+func ReadFile(path string) (content []byte, err error) {
+	if path == "-" {
+		content, err = ioutil.ReadAll(os.Stdin)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to from stdin: %w", err)
+		}
+	} else {
+		content, err = ioutil.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to read %q: %w", path, err)
+		}
+	}
+
+	return content, nil
 }
--- a/cmd/sops/encrypt.go
+++ b/cmd/sops/encrypt.go
@@ -58,7 +58,7 @@ func encrypt(opts encryptOpts) (encryptedFile []byte, err error) {
 	// Load the file
 	fileBytes, err := common.ReadFile(opts.InputPath)
 	if err != nil {
-		return nil, common.NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
+		return nil, err
 	}
 	branches, err := opts.InputStore.LoadPlainFile(fileBytes)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


jkroepke:
In case tests are missing, please give me a hint where should I add the tests, thanks.

cmd/sops/encrypt.go
```diff
@@ -57,7 +56,7 @@ func ensureNoMetadata(opts encryptOpts, branch sops.TreeBranch) error {
 
 func encrypt(opts encryptOpts) (encryptedFile []byte, err error) {
 	// Load the file
-	fileBytes, err := ioutil.ReadFile(opts.InputPath)
+	fileBytes, err := common.ReadFile(opts.InputPath)
 	if err != nil {
 		return nil, common.NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
```
mulder999, line 61: Use error from common.ReadFile function

cmd/sops/main.go
```diff
@@ -710,17 +710,21 @@ func main() {
...
-			return toExitError(err)
-		}
-		if _, err := os.Stat(fileName); os.IsNotExist(err) {
-			if c.String("add-kms") != "" || c.String("add-pgp") != "" || c.String("add-gcp-kms") != "" || c.String("add-hc-vault-transit") != "" || c.String("add-azure-kv") != "" || c.String("add-age") != "" ||
-				c.String("rm-kms") != "" || c.String("rm-pgp") != "" || c.String("rm-gcp-kms") != "" || c.String("rm-hc-vault-transit") != "" || c.String("rm-azure-kv") != "" || c.String("rm-age") != "" {
-				return common.NewExitError("Error: cannot add or remove keys on non-existent files, use `--kms` and `--pgp` instead.", codes.CannotChangeKeysFromNonExistentFile)
+
+		fileName := c.Args()[0]
+		if fileName != "-" {
+			fileName, err := filepath.Abs(fileName)
```
mulder999, line 716: This shall be moved in common.ReadFile

cmd/sops/common/common.go
```diff
@@ -127,7 +127,7 @@ func EncryptTree(opts EncryptTreeOpts) error {
 
 // LoadEncryptedFile loads an encrypted SOPS file, returning a SOPS tree
 func LoadEncryptedFile(loader sops.EncryptedFileLoader, inputPath string) (*sops.Tree, error) {
-	fileBytes, err := ioutil.ReadFile(inputPath)
+	fileBytes, err := ReadFile(inputPath)
 	if err != nil {
 		return nil, NewExitError(fmt.Sprintf("Error reading file: %s", err), codes.CouldNotReadInputFile)
```
mulder999, line 132: Use error from common.ReadFile

mulder999 (commented):
Seems mostly OK for me. A few improvements needed now that error message is managed centrally.

cmd/sops/main.go
jkroepke: I do not get this. What should be moved exactly?

mulder999: My bad sorry.

mulder999 (approved):

jkroepke:
How to move forward here?

jkroepke:
@ajvb (chosen as an active maintainer) could you take a look here?

jkroepke:
Changing merge target to develop

jkroepke:
Reading https://github.com/mozilla/sops/discussions/927, its like it will take some time.

cmd/sops/common/common.go
```diff
@@ -441,3 +441,19 @@ func PrettyPrintDiffs(diffs []Diff) {
 		}
 	}
 }
+
+func ReadFile(path string) (content []byte, err error) {
```
hiddeco, line 445: I think this tries to do two things which really should be separate things.

In my opinion, handling of files should really be _just_ handling of files, with `os.Stdin` not classifying as such. Part because from the perspective of SOPS as an SDK, `decrypt.Decrypt` magically reading from `os.Stdin` would be kind of surprising.

Having said this, I would deal with `-` as a special path case within commands, and e.g. create another wrapper around `DataWithFormat` for an `io.Reader`.

jkroepke: > files should really be just handling of files

I don't know. On Unix, stdin is just an other file descriptor and could be handle as file, too. (like `/dev/stdin`)

Living the unix philosophy, there should not be a difference between `/mnt/file` and `/proc/self/fd/1`.

From technical point of view, it may should fine. 

Creating an other wrapper, it possible. But the other wrapper would be the exact same copy/paste code.

hiddeco: How is this the same copy/paste code?

```go
func Reader(r io.Reader, format string) (cleartext []byte, err error) {
        encryptedData, err := io.ReadAll(r)
	if err != nil {
		return nil, err
	}
	return DataWithFormat(encryptedData, FormatFromString(format))
}

hiddeco: NB: I am fine with a function called `ReadFileOrStdin` (or something something) existing (and utilized) within `cmd/sops`, but I wouldn't want this to be the default for `decrypt`.

PR #1109, opened 2022-08-15
Add manage command

dooferlad:
Prior to this change, if we wanted to add or remove a master key, we had to rotate the data key. While there is nothing wrong with rotating the data key, having to do so at the same time as adding or removing a master key makes git history noisy.

This change adds a manage command, which allows separate master key management to data key rotation.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -552,6 +552,10 @@ func main() {
 			Name:  "rotate, r",
 			Usage: "generate a new data encryption key and reencrypt all values with the new key",
 		},
+		cli.BoolFlag{
+			Name:  "manage, m",
+			Usage: "manage master keys without reencrypting all values with a new data key",
+		},
 		cli.StringFlag{
 			Name:   "kms, k",
 			Usage:  "comma separated list of KMS ARNs",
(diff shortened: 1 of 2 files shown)


cmd/sops/main.go
```diff
@@ -552,6 +552,10 @@ func main() {
 			Name:  "rotate, r",
 			Usage: "generate a new data encryption key and reencrypt all values with the new key",
 		},
+		cli.BoolFlag{
+			Name:  "manage, m",
+			Usage: "manage master keys without reencrypting all values with a new data key",
```
felixfontein, line 557: I think this requires a BIG FAT SECURITY WARNING for the case that you actually want to remove master keys. Or maybe the code should simply fail if a master key is removed and this flag is set.

dooferlad: Yes, good point - removing a master key one is probably happening because the key has been compromised or is no longer in use, in which case it shouldn't be trusted anyway. I will update the PR.

felixfontein, line 557: Sorry that I didn't think about this earlier, but adding new master keys without cycling the data password also has potential security implications that can be dangerous. Think of a devops git repo where a file is readable by users A and B and contains some high-value secrets. At some point it is decided that parts of that file also need to be readable by C, so the things that should not be read by C are moved to another file, and then this command is used to give C access to this file.

Since the data key has not been cycled, C now has access to the data key **and** to the git history, allowing C to also read the parts that C was not supposed to be able to read.

This is probably why so far there's only `rotate`, which always rotates the data key.

dooferlad: That is reasonable - do you think this could land with a big fat warning? In my case we wanted to add a second key so we had a backup key and there isn't a history problem. I really don't mind dropping this since I have already used it locally.

felixfontein: I personally think it is ok with a big fat warning, since there are cases (like the one you are describing) where this is really useful. But I'm just a random person making comments here, my opinion doesn't count that much ;) @ajvb what do you think?

PR #1115, opened 2022-08-27
--version without network request

### LINKED ISSUE #1100: sops --version (prevent phone home)
I'm running sops in a air-gapped area without internet access, running command sops --version / -v is making it wait for ages as it tries to go external...

[warning] failed to retrieve latest version from upstream: Get "https://raw.githubusercontent.com/mozilla/sops/master/version/version.go": dial tcp IP : i/o timeout

Is there a way to skip/prevent this?

felixfontein:
Adds a new flag `--disable-version-check` which disables the version check (phone home) on `--version`.

I would argue that this behavior should be the default, but for backwards compatibility the other way around is also fine for me. As long as it's possible to turn it off.

Fixes #1100

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -552,6 +552,10 @@ func main() {
 			Name:  "rotate, r",
 			Usage: "generate a new data encryption key and reencrypt all values with the new key",
 		},
+		cli.BoolFlag{
+			Name:  "disable-version-check",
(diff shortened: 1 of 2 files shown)


ajvb (approved):

felixfontein:
@ajvb thanks for reviewing and merging!

Issue #305, opened 2018-03-07
Ignore KMS when decrypting


jashandeep-sohi:
Is there someway to prevent sops from attempting to use KMS and prefer PGP instead when decrypting a file?

autrilla:
Not currently, no. I suppose you could reorder the keys in the encrypted YAML, as SOPS goes through them in order.

jvehent:
Out of curiosity: why do you think having the ability to ignore KMS is a feature we need?

mccutchen:
At `$JOB` we're wrapping sops in some higher-level tooling, which takes care of assuming the right role for a user working with secrets and populating the environment with appropriate AWS credentials before invoking sops.

We also provide a `--recover-secrets` flag for this tooling, which basically has the effect you're looking for here @jashandeep-sohi.  It does the simplest thing we could think of:  When `--recover-secrets` is given, we do not populate the environment with AWS credentials, which means sops's attempts to use KMS will fail and it will automatically fall back to PGP.

So even though this functionality is not built into sops, you can at least simulate it by manipulation of the environment in which sops is running.  Hope this helps!

jvehent:
would a `--use-key pgp[i]` option, where `i` is the position of the key, address your need?

jashandeep-sohi:
@jvehent I work offline/near-offline a lot and sometimes the latency of trying to reach the KMS is too much. I do like the idea of a generic solution like `--use-key (pgp|kms)[i]`.

jpadilla:
I'm actually running into this too. We're using sops in development with PGP and KMS for production. KMS will always be tried first and fail while in development. A `--use-key` flag would be pretty useful, I guess you could optionally specify the index of the key to use, but would expect `--use-key pgp` to just try to decrypt with all PGP keys.

brad-jones:
> Not currently, no. I suppose you could reorder the keys in the encrypted YAML, as SOPS goes through them in order.

@autrilla I did attempt this and it appeared to make no difference what so ever, still attempts to call out to KMS.

You were referring to simply moving `sops.pgp` to appear above `sops.kms`?

Whats more the second you edit the file `sops` then reorders it such that the `pgp` "YAML" key is last again.

Obviously each and every time a developer decrypts a secret locally thats a KMS call that costs money. If they happened to have a valid PGP key it would make sense to try that first. So the operation is cheaper and faster.

[...]

PR #1121, opened 2022-09-03
Sort masterkeys to try offline decrypt methods first

bkreitch:
This is to address #305. Sort master keys so PGP/age keys appear before the others and being used first.

### DIFF
--- a/sops.go
+++ b/sops.go
@@ -42,6 +42,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -665,7 +666,11 @@ func (m Metadata) GetDataKeyWithKeyServices(svcs []keyservice.KeyServiceClient) 
 // returning as soon as one key service succeeds.
 func decryptKeyGroup(group KeyGroup, svcs []keyservice.KeyServiceClient) ([]byte, error) {
 	var keyErrs []error
-	for _, key := range group {
+	// Sort MasterKeys in the group so we try offline ones first
+	// Use sorted indices to avoid group slice modification
+	indices := sortKeyGroupIndices(group)
+	for _, indexVal := range indices {
+		key := group[indexVal]
 		part, err := decryptKey(key, svcs)
 		if err != nil {
 			keyErrs = append(keyErrs, err)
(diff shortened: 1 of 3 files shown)


bkreitch:
@ajvb Hi, sorry to bother you, but could you maybe take a look please? Hopefully it's a quick win :)

hiddeco:
This is a sensible change, I think. It would be great if you could rebase your work and sign-off your commit. Thanks! :bow:

bkreitch:
@hiddeco Ok, rebased and signed-off. Great to see that the project is back to life! 🎉

hiddeco:
I have to check this against the history from the issue, which is why I moved it one milestone further.

But do know it’s on the (now active) backlog. 🍒

felixfontein (approved):
Looks good to me.

sops.go
```diff
@@ -664,6 +665,12 @@ func (m Metadata) GetDataKeyWithKeyServices(svcs []keyservice.KeyServiceClient)
 // returning as soon as one key service succeeds.
 func decryptKeyGroup(group KeyGroup, svcs []keyservice.KeyServiceClient) ([]byte, error) {
 	var keyErrs []error
+
+	// Sort MasterKeys in the group so we try offline ones first
+	sort.SliceStable(group, func(i, j int) bool {
```
felixfontein, line 670: I'm a bit worried that this modifies the passed-in data. How much work would it be to sort a list of indices instead?

felixfontein:
https://github.com/getsops/sops/issues/305#issuecomment-1793766331 might be preferable to this solution, since it is more flexible.

sops.go
bkreitch: @felixfontein you mean to sort a list of indices and then use this list to iterate over group in for loop below? Should be a small change.

felixfontein: Yes, that's what I mean.

bkreitch: @felixfontein Ok, I've changed it to use indices. Please let know if you have suggestion for better variable names.

bkreitch:
> [#305 (comment)](https://github.com/getsops/sops/issues/305#issuecomment-1793766331) might be preferable to this solution, since it is more flexible.

@felixfontein Hi, I've opened #1345 which supports comma-separated `--decryption-order` to replace this PR.
(or I can push the new code into this PR if it's more convenient)

Issue #878, opened 2021-05-31
Allow to set shamir_threshold to 1 or disable it.


schrej:
We manage a lot of secrets in a single repository, which need to be accessible by different groups of people. In order to avoid having to set all keys on each `creation_rule` we'd like to use YAML anchors and references in order to manage groups of keys. Unfortunately the YAML spec has no way of merging anchors contents into a single list. It's only possible to have objects/maps referenced by an anchor. By using key groups it's possible to just create one entry for each group. See below for an example.

Unfortunately the automatically enabled shamir key exchange when using multiple key groups blocks us from using this. It's also not possible to set the shamir_threshold to 1, as that obviously makes shamir useless.

Being able to disable shamir, e.g. by setting the threshold to 1, would enable an easy way of managing groups of keys.

If there is any other way of achieving this within sops.yaml that I haven't found yet, please let me know.

```yaml
groups:
  groupA: &groupA
    age:
     - agekey1
     - agekey2
  groupB: &groupB
    age:
     - agekey3
     - agekey4
creation_rules:
  - path_regex: /example/
    shamir_threshold: 1
    key_groups:
    - *groupA
    - *groupB
    - age:
      - agekey5
```
this gets expanded to
```yaml
creation_rules:
  - path_regex: /example/
    shamir_threshold: 1
    key_groups:
    - age:
      - agekey1
      - agekey2
   - age:
      - agekey3
      - agekey4
    - age:
      - agekey5
```
CC @MaxRink

felixfontein:
The 3.9.0 release will allow to solve this with merge groups (#1493). Setting shamir_threshold to 1 is no longer needed for this.

PR #1123, opened 2022-09-08
Allow shamir_threshold: 1

puiterwijk:
This allows the case where multiple keygroups are used, but any one of the groups needs to be able to decrypt the secret. This is for example useful if some people use age keys and some use pgp, but each of them should be able to decrypt the secret by themselves.

Fixes: #878

### DIFF
--- a/shamir/shamir.go
+++ b/shamir/shamir.go
@@ -175,8 +175,8 @@ func Split(secret []byte, parts, threshold int) ([][]byte, error) {
 	if parts > 255 {
 		return nil, fmt.Errorf("parts cannot exceed 255")
 	}
-	if threshold < 2 {
-		return nil, fmt.Errorf("threshold must be at least 2")
+	if threshold < 1 {
+		return nil, fmt.Errorf("threshold must be at least 1")
 	}
 	if threshold > 255 {
 		return nil, fmt.Errorf("threshold cannot exceed 255")
@@ -238,9 +238,9 @@ func Split(secret []byte, parts, threshold int) ([][]byte, error) {
 // Combine is used to reverse a Split and reconstruct a secret
 // once a `threshold` number of parts are available.
 func Combine(parts [][]byte) ([]byte, error) {
-	// Verify enough parts provided
-	if len(parts) < 2 {
-		return nil, fmt.Errorf("less than two parts cannot be used to reconstruct the secret")
+	// Verify some parts provided
+	if len(parts) < 1 {
+		return nil, fmt.Errorf("no parts cannot be used to reconstruct the secret")
 	}
 
 	// Verify the parts are all the same length
--- a/shamir/shamir_test.go
+++ b/shamir/shamir_test.go
@@ -20,12 +20,27 @@ func TestSplit_invalid(t *testing.T) {
 		t.Fatalf("expect error")
 	}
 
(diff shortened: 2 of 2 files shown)


dali99:
>if some people use age keys and some use pgp, but each of them should be able to decrypt the secret by themselves.

You can actually mix PGP and age keys together in a single group.

```
key_groups:
- age:                                                                                                                                                                                
  - *user1                                                                                                                                                                       
  - *host2                                                                                                                                                                       
  pgp:                                                                                                                                                                                
  - *user2
```

Note the lack of a `-` on the pgp line to put them in the same group

tonobo:
Well, this allows grouping multiple keys together, which is currently not possible:

```yaml
keys:
  dept_a: &dept_a
    age:
      - &user_a age...
      - &user_b age
  dept_b: &dept_b
    age:
      - &user_c age...
      - &user_d age


creation_rules:
  - path_regex: \.*flux/.*
    encrypted_regex: '^(stringData|data)$'
    shamir_threshold: 1
    key_groups:
      - *dept_a
      - *dept_b
```

felixfontein (commented):
The change looks correct.

I'm still not convinced that this is a really good idea, since there is no good reason to allow the threshold to be 1, except as a hack. It might be better to extend the config file format instead to make joining lists for the same key type using YAML anchors easier.

On the other hand, allowing a threshold of 1 doesn't really hurt. The worst case I can think of is that some user notices that a threshold of 1 works and thinks that it is the smallest value that forces that keys from more than one group. (I do hope nobody tries that, but on the other hand some folks also decided to use an exponent of 1 for RSA, so... :) )

pbsds:
Using a threshold=1 as a means to _merge_ key groups is a hack, which also removes the ability to use shamir thresholding should it be needed. I'm rather in favor of either flattening lists, or introducing a `merge` type:

Flattening only works within each type of key:


```yaml
keys:
  dept_a: &dept_a
    - &user_a age...
    - &user_b age...
  dept_b: &dept_b
    - &user_c age...
    - &user_d age...

creation_rules:
  - path_regex: \.*flux/.*
    encrypted_regex: '^(stringData|data)$'
    shamir_threshold: 1
    key_groups:
    - age:
      - *dept_a
      - *dept_b
```

While a `merge` target can deal with multiple kinds of keys, and be used agnostically:

```yaml
keys:
  dept_a: &dept_a
    age:
      - &user_a age...
      - &user_b age...
  dept_b: &dept_b
    age:
      - &user_c age...
      - &user_d age...

creation_rules:
  - path_regex: \.*flux/.*
    encrypted_regex: '^(stringData|data)$'
    shamir_threshold: 1
    key_groups:
    - merge:
      - *dept_a
      - *dept_b
```

jonasbadstuebner:
I would opt for the `merge` target. Are you open for a new PR in this regard @felixfontein ?

felixfontein:
`merge` sounds good to me!

felixfontein:
While the problem sketched in #878 can now be solved with merge groups, I'm not sure about what was mentioned in this PR:

> This allows the case where multiple keygroups are used, but any one of the groups needs to be able to decrypt the secret. This is for example useful if some people use age keys and some use pgp, but each of them should be able to decrypt the secret by themselves.

@puiterwijk I think this can also all be covered with merge groups, but maybe I'm missing something?

PR #1124, opened 2022-09-11
Do not report version when upstream check fails

stoned:
... otherwise you could get something like the following (here sops is run without network connection available):
```
$ ./sops --version
sops 3.7.3
[warning] failed to retrieve latest version from upstream: Get "https://raw.githubusercontent.com/mozilla/sops/master/version/version.go": dial tcp: lookup raw.githubusercontent.com: Temporary failure in name resolution

[warning] failed to compare current version with latest: Version string empty
 (latest)
$
```

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -26,11 +26,12 @@ func PrintVersion(c *cli.Context) {
 		outdated, err := AIsNewerThanB(upstreamVersion, Version)
 		if err != nil {
 			out += fmt.Sprintf("\n[warning] failed to compare current version with latest: %v\n", err)
-		}
-		if outdated {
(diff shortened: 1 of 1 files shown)


hiddeco:
Already fixed via #1115

felixfontein:
@hiddeco this hasn't been fixed. What this PR is fixing is that in case of a comparison error (when the new version is fetched), that both the error and `(latest)` (which is wrong) are printed.

(Probably the whole comparison logic should be skipped when a network error happens, though, instead of a network error triggering the "Version string empty" error.)

hiddeco:
@felixfontein you are absolutely right, thanks.

@stoned any chance you can rebase and sign-off your commit?

hiddeco (approved):
Thank you @stoned :bow: :mango:

PR #1092, opened 2022-07-13
azkv: update Azure SDK to v0.10.0

hiddeco:
This depends on #1091 being merged into `develop` first, as I took the `go.mod` file state from there as a base.

Other than that, this PR does nothing more than dealing with the method (signature) and client construction changes from the `v0.6.0` release of the `azkeys` module. This should be the last major change in their SDK.

I manually ran the integration test suite which passes without any issues:

```console
$ go test -json -tags integration ./...
=== RUN   TestMasterKey_Encrypt
--- PASS: TestMasterKey_Encrypt (1.66s)
=== RUN   TestMasterKey_Decrypt
--- PASS: TestMasterKey_Decrypt (1.00s)
=== RUN   TestMasterKey_EncryptDecrypt_RoundTrip
--- PASS: TestMasterKey_EncryptDecrypt_RoundTrip (0.48s)
...
PASS
ok  	go.mozilla.org/sops/v3/azkv	5.069s

Process finished with the exit code 0

PR #1147, opened 2022-11-21
chore: update Go dependencies

hiddeco:
This excludes an update of the Azure SDKs, which has been dealt with separately in #1092.


hiddeco:
Merging this would address #1136.

hiddeco:
Merging this would now also address #1163, #1169, #1176

hiddeco:
Merging this would also address #1189, #1181, #1208 (and probably more...)

onedr0p (approved):

PR #1148, opened 2022-11-21
build: update Go to 1.20

hiddeco:
This updates Go used for builds to 1.20, and updates a forgotten Alpine 3.15 version in a `Dockerfile` to 3.18.


devstein (approved):

hiddeco (commented):
It may actually be wise to update the version in `go.mod` to `1.19`, as this is the lowest supported stable version. Will add a commit in a bit.

Issue #634, opened 2020-02-20
Deprecate storing aws profile within the encrypted file / with the key


ajvb:
There seems to have been a number of issues surrounding our use of AWS profiles:

- https://github.com/mozilla/sops/issues/471
- https://github.com/mozilla/sops/issues/614
- https://github.com/mozilla/sops/issues/619
- https://github.com/mozilla/sops/issues/474
- https://github.com/mozilla/sops/issues/439

Storing the AWS Profile in the encrypted file breaks expectations because it cannot be decrypted on systems where that profile doesn't exist. This is very different from how folks are used to interacting with AWS.

I'd propose that we deprecate storing the aws profile within the key and instead require users to pass `--aws-profile` every time they want to use an AWS profile.

ajvb:
@autrilla What do you think?

autrilla:
Given all the confusion around it I think we should probably deprecate it, and also deprecate `--aws-profile`. The AWS lets you use environment variables to control what profile is used, users should just use that.

pnguyen-okta:
It's really useful to pass the profile using `--aws-profile`. I can understand reducing confusion but is it possible to just deprecate storing the aws profile in the encrypted file and keep `--aws-profile`?

thanks

ajvb:
I really like how we implemented this in `frost`: https://github.com/mozilla/frost/blob/master/aws/client.py#L17-L32

You can pass in a profile using `--aws-profiles` (though for `sops` it would be `--aws-profile`) and it gets passed to the session creation, otherwise we just do the default AWS auth flow.

I'd also consider expanding this to removing support for setting AWS roles in the config. Most tools that interact with AWS via the command line require you to authenticate to AWS outside of the context of the tool and it makes sense to me that sops functions the same way. It seems like most of the issues are around people being confused it doesn't work that way (i.e. https://github.com/mozilla/sops/issues/555).

autrilla:
Why do you prefer using a flag vs an environment variable? It’s kind of the
same thing with different syntax, no?

Using a flag would let us simplify the code a bit...

jvehent:
> I'd also consider expanding this to removing support for setting AWS roles in the config.

Original support for that came from a requirement to invoke KMS from various accounts transparently, by having Sops assume roles in those accounts. I believe this feature is still in use in our own teams.

dsalaza4:
[...]

PR #1150, opened 2022-12-04
draft: feat(gcp-kms): impersonate service account

eraac:
Implementation of the possibility of impersonate service account to call the GCP KMS api. 

I'm not 100% familiar with the project yet, I'm not sure I've done the work correctly, feel free to point my errors if any

For the unit tests, I guess I can probably add test for the sops metadata to ensure the `gcp_impersonate_service_account` is correctly set, but I need to be guide a bit here. Otherwise concerning the feature itself, I've no idea how to test it (except manually)

I still have to update the README, I'll do this tomorrow

Thanks for the review ❤️


eraac:
I got the same kind of issue from aws-profile https://github.com/mozilla/sops/issues/634

I want to use default application credential for encrypt (developpers credentials), and another one via impersonate for decrypt (CI). The current workaround is to edit encrypted file metadata, but isn't very conveniant ... 

But unfortunately, this change seems to be more complicated and I guess isn't related to this PR

PR #1155, opened 2022-12-08
Google App Default Credentials Unusable

### LINKED ISSUE #1151: Implicit and explicit Google Application Default Credentials cannot be used
Google API clients have the ability to use credentials from a well-known path.  This path is where the `gcloud` SDK writes them when you execute `gcloud auth login --update-adc` or `gcloud auth application-default login`.  There is no need to set the `GOOGLE_APPLICATION_CREDENTIALS` envvar in this case (though you can to override the path).  And from the code side, there is no client config needed for it to automatically use that credential.  It Just Works(R).

PR #953 broke app default credentials while making the GCP KMS integration more CI-friendly.  If one did NOT set `GOOGLE_CREDENTIALS`, sops would try to create the KMS client with a zero-byte JSON credentials slice, which caused a JSON parse error.

Only a small change is necessary to let #953 coexist with `GOOGLE_APPLICATION_DEFAULT` credentials.  PR incoming.

ryanhos:
Closes #1151, which broke using app default credentials (and relying on workload identity on GCP hosts), by assuming that GOOGLE_CREDENTIALS would always contain a valid value.

### DIFF
--- a/gcpkms/keysource.go
+++ b/gcpkms/keysource.go
@@ -217,8 +217,8 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
 			return nil, err
 		}
 		if credentials != nil {
-			opts = append(opts, option.WithCredentialsJSON(key.credentialJSON))
-		}
+			opts = append(opts, option.WithCredentialsJSON(credentials))
+		} // ELSE: KMS library will use Google Application Default Credentials.
 	}
 	if key.grpcConn != nil {
 		opts = append(opts, option.WithGRPCConn(key.grpcConn))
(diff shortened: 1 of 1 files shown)


enchorb:
@ajvb Can you merge this

gcpkms/keysource.go
```diff
@@ -238,8 +238,10 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 func getGoogleCredentials() ([]byte, error) {
-	defaultCredentials := os.Getenv(SopsGoogleCredentialsEnv)
-	if _, err := os.Stat(defaultCredentials); err == nil {
+	defaultCredentials, wasSet := os.LookupEnv(SopsGoogleCredentialsEnv)
+	if !wasSet {
+		return nil, nil
+	} else if _, err := os.Stat(defaultCredentials); err == nil {
 		return os.ReadFile(defaultCredentials)
 	}
 	return []byte(defaultCredentials), nil
```
hiddeco, line 247:
```suggestion
	if defaultCredentials, ok := os.LookupEnv(SopsGoogleCredentialsEnv); ok {
		if _, err := os.Stat(defaultCredentials); err == nil {
			return os.ReadFile(defaultCredentials)
		}
		return []byte(defaultCredentials), nil
	}
	return nil, nil
```

Seems more Go idiomatic.

hiddeco:
Thank you for your contribution! :bow: 

In addition to my review comment, it would be great if you could rebase this and sign-off your commit.

PR #1174, opened 2023-02-04
Encrypt in a package and export method encrypt to be used in import "go.mozilla.org/sops/v3/encrypt"

skrese:
In Go projects where you need encrypting with sops it'd be nice to have a direct access to the method `func encrypt(opts encryptOpts) (encryptedFile []byte, err error)` in **cmd/sops/encrypt.go**. In that manner you could do something like:
```
import (
...
    "go.mozilla.org/sops/v3/encrypt"
...
)
...
encryptedFile, _ := encrypt.Encrypt(encrypt.EncryptOpts{
                        InputStore:        &yaml.Store{},
                        OutputStore:       &yaml.Store{},
                        InputBytes:        buffer,
                        Cipher:            aes.NewCipher(),
                        KeyServices:       svcs,
                        UnencryptedSuffix: "_unencrypted",
                        EncryptedSuffix:   "",
                        UnencryptedRegex:  "",
                        EncryptedRegex:    "",
                        KeyGroups:         groups,
                        GroupThreshold:    0,
                    })
```

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -28,6 +28,7 @@ import (
 	publishcmd "go.mozilla.org/sops/v3/cmd/sops/subcommand/publish"
 	"go.mozilla.org/sops/v3/cmd/sops/subcommand/updatekeys"
 	"go.mozilla.org/sops/v3/config"
+	"go.mozilla.org/sops/v3/encrypt"
 	"go.mozilla.org/sops/v3/gcpkms"
 	"go.mozilla.org/sops/v3/hcvault"
 	"go.mozilla.org/sops/v3/keys"
@@ -796,7 +797,7 @@ func main() {
 			if err != nil {
 				return toExitError(err)
 			}
-			output, err = encrypt(encryptOpts{
+			output, err = encrypt.Encrypt(encrypt.EncryptOpts{
 				OutputStore:       outputStore,
 				InputStore:        inputStore,
 				InputPath:         fileName,
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -1,4 +1,4 @@
-package main
+package encrypt
 
 import (
 	"io/ioutil"
@@ -14,7 +14,7 @@ import (
 	"go.mozilla.org/sops/v3/version"
 )
 
-type encryptOpts struct {
+type EncryptOpts struct {
 	Cipher            sops.Cipher
 	InputStore        sops.Store
 	OutputStore       sops.Store
@@ -46,7 +46,7 @@ func (err *fileAlreadyEncryptedError) UserError() string {
 	return wordwrap.WrapString(message, 75)
 }
 
-func ensureNoMetadata(opts encryptOpts, branch sops.TreeBranch) error {
+func ensureNoMetadata(opts EncryptOpts, branch sops.TreeBranch) error {
 	for _, b := range branch {
 		if b.Key == "sops" {
 			return &fileAlreadyEncryptedError{}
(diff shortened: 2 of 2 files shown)


Gitlit1010 (approved):

hiddeco (commented):
What justifies exposing this while the underlying API mechanics are publicly available and allow you to reconstruct the behavior with precision?

skrese:
@hiddeco appreciate for reviewing the code. What I wanted is to include sops as a library in a Go project and then use it for encryption (also decryption) but couldn't achieve that with public API. Hence, exported encrypt.Encrypt and encrypt.EncryptOpts. Have I missed something and I could do it with API? My PR allows me to do:
```
encryptedFile, _ := encrypt.Encrypt(encrypt.EncryptOpts{
                        InputStore:        &yaml.Store{},
                        OutputStore:       &yaml.Store{},
                        InputBytes:        buffer,
                        Cipher:            aes.NewCipher(),
                        KeyServices:       svcs,
                        UnencryptedSuffix: "_unencrypted",
                        EncryptedSuffix:   "",
                        UnencryptedRegex:  "",
                        EncryptedRegex:    "",
                        KeyGroups:         groups,
                        GroupThreshold:    0,
                    })
```

hiddeco:
You can absolutely achieve this using the public API by using the lower-level APIs as they are used within the (now private) encrypt function, the key service client you need to replicate SOPS is available in: https://github.com/getsops/sops/blob/main/keyservice/client.go#L14-L17.

With this, it should be possible using something like:

```go
	// One of https://github.com/getsops/sops/tree/main/stores,
	// for the file type you are working with. There are some utilities to figure
	// out what store to used based on a path, see e.g.
	// common.DefaultStoreForPathOrFormat()
	inputStore := &yaml.Store{}

	inputB := []byte(`<your bytes to encrypt>`)
	branches, err := inputStore.LoadPlainFile(inputB)
	if err != nil {
		return nil, fmt.Errorf("could not load plain file: %w", err)
	}

	tree := sops.Tree{
		Branches: branches,
		Metadata: sops.Metadata{
			// Your options from your own comment.
		},
		// In most scenarios, FilePath is not strictly required if you solely want to encrypt
		// bytes. Internally, they are used for auditing purposes which is not really actively
		// being utilized by 99% of the users.
		// FilePath: "",
	}

	// Local client which is basically how `sops` behaves. If you want to make custom
	// configurations for client secrets, etc. There is a `NewCustomLocalClient()`.
	ks := keyservice.NewLocalClient()
	dataKey, errs := tree.GenerateDataKeyWithKeyServices(ks)
	if len(errs) > 0 {
		return nil, fmt.Errorf("could not generate data key: %s", errs)
	}

	// Perform the encrypt operation.
	if err = common.EncryptTree(common.EncryptTreeOpts{
		DataKey: dataKey,
		Tree:    &tree,
		Cipher:  opts.Cipher,
	}); err != nil {
		return nil, err
	}

	// Output the encrypted bytes.
	encryptedB, err := inputStore.EmitEncryptedFile(tree)
	if err != nil {
		return nil, fmt.Errorf("could not marshal tree: %w", err)
	}
```

Which will break you free from all sorts of assumptions around returned error codes, etc. specifically designed for the CLI, which do currently exist in the `decrypt` package. While also providing fine grain control over any possible options you may want to configure now or later, and the encryption process as a whole.

skrese:
@hiddeco I think you're pointing me to the right direction. Although still not sure how to set up groups (as I did in my original solution):
`KeyGroups:         groups,`
The problem I had, I couldn't set up groups for my purposes, where I needed KMS key and multiple age keys. If I understand correctly you're suggesting I should use `NewCustomLocalClient()`, right?

hiddeco:
Key groups are for encryption instructions (and the loading looks pretty much like https://github.com/getsops/sops/blob/7b3a032adee818183189c89fcbbae212fbd50318/cmd/sops/main.go#L1088, which should be OK).

However, the further private parts are by default resolved from the runtime environment (via `NewLocalClient()`). If your goal is to have those work in an isolated way, this is now possible in the `v3.8.0-rc.1` release using `NewCustomLocalClient()`, and I posted the ingredients in a Twitter thread (but they should eventually be documented in this repository): https://twitter.com/hiddeco/status/1695065074769997862

PR #1183, opened 2023-03-07
Fix #1044 - Immediately close the temporary file once created

### LINKED ISSUE #1044: sops edits a readonly temp file on Windows
I have a sops encrypted file `foo.enc.json` and when I run `sops foo.enc.json`, vim tells me:

```
"~\AppData\Local\Temp\1012848330\foo.enc.json" [readonly][Incomplete last line][unix format] 3 lines, 13 characters
```

This happens both on Windows and Cygwin, so I guess it's an issue with the Windows binary. This issue doesn't appear on the Linux binary. Tested on Sops v3.7.2.

I think the function in `editTree` in `edit.go` is where it's at, but my Go's real rusty. Please tell me if I can assist in some way.

https://github.com/mozilla/sops/blob/v3.7.2/cmd/sops/edit.go#L110

Varriount:
Immediately close the temporary file created by `sops` before invoking the user's editor.

On Windows, the Go standard library opens file handles with only shared read and write access (thus excluding deletion access). This can prevent other processes (like editors) that require opening handles with delete access from writing to those same files.

### DIFF
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -44,7 +44,7 @@ type editExampleOpts struct {
 }
 
 type runEditorUntilOkOpts struct {
-	TmpFile        *os.File
+	TmpFileName    string
 	OriginalHash   []byte
 	InputStore     sops.Store
 	ShowMasterKeys bool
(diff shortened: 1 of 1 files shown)


mrcarter7:
Is there an alternative solution here? This was closed back in March but still having the same issue. Thanks!

felixfontein:
@mrcarter7 does this issue still happen with 3.8.0-rc.1? (See https://github.com/getsops/sops/issues/1044#issuecomment-1720116418)

mrcarter7:
I saw the comment you linked late yesterday I'll try 3.8.0-rc.1 later today. Thanks!

mrcarter7:
Okay I do get the temp file in my editor but it is deleted from the file system and not in my current working directory but I think that's intended. I'll keep working with it. Thanks!

PR #1184, opened 2023-03-07
Close the temporary file before invoking the user's editor 

Varriount:
On Windows, the Go standard library opens file handles with only shared read and write access (thus excluding deletion access). This can prevent other processes (like editors) that require opening handles with delete access from writing to those same files.

### DIFF
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -44,7 +44,7 @@ type editExampleOpts struct {
 }
 
 type runEditorUntilOkOpts struct {
-	TmpFile        *os.File
+	TmpFileName    string
 	OriginalHash   []byte
 	InputStore     sops.Store
 	ShowMasterKeys bool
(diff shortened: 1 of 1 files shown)


cmd/sops/edit.go
```diff
@@ -135,18 +135,20 @@ func editTree(opts editOpts, tree *sops.Tree, dataKey []byte) ([]byte, error) {
 		return nil, common.NewExitError(fmt.Sprintf("Could not write output file: %s", err), codes.CouldNotWriteOutputFile)
 	}
 
-	// Close temporary file, since Windows won't delete the file unless it's closed beforehand
-	defer tmpfile.Close()
+	// Close the temporary file immediately
+	// On Windows, the standard library opens files in a way that prevents certain editors from writing to them.
+	tmpfile.Close()
+	var tmpfileName = tmpfile.Name()
```
Varriount, line 141: I can change everything back to using the existing `tmpfile` variable if required. I extracted the name because I felt it was better to stop using a closed file entirely (to prevent possible future confusion about the state of the file).

holiman (approved):
Clearly an improvement IMO

hiddeco (approved):
This looks indeed like a better approach given the platform constraints.

If you can rebase the PR and sign-off the commit, while maybe also tidying the commit to adhere to the 50/72 rule. I would be really happy to have this included in next release :bow: :pineapple:

PR #1188, opened 2023-03-23
gcpkms: Add support for oauth tokens using env var

warwick-mitchell1:
Add support for CLOUDSDK_AUTH_ACCESS_TOKEN environment to pass though an oauth access token directly or via a file.

### DIFF
--- a/gcpkms/keysource.go
+++ b/gcpkms/keysource.go
@@ -16,13 +16,15 @@ import (
 	"google.golang.org/grpc"
 
 	"go.mozilla.org/sops/v3/logging"
+	"golang.org/x/oauth2"
 )
 
+// SopsGoogleCredentialsEnv or SopsGoogleCredentialsOAuthEnv can be set as an
+// environment variable as either a path to a credentials file, or directly
+// as the variable's value in JSON format.
 const (
-	// SopsGoogleCredentialsEnv can be set as an environment variable as either
-	// a path to a credentials file, or directly as the variable's value in JSON
-	// format.
-	SopsGoogleCredentialsEnv = "GOOGLE_CREDENTIALS"
+	SopsGoogleCredentialsEnv      = "GOOGLE_CREDENTIALS"
+	SopsGoogleCredentialsOAuthEnv = "CLOUDSDK_AUTH_ACCESS_TOKEN"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


gcpkms/keysource.go
```diff
@@ -233,14 +234,32 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
-	}
-	return []byte(defaultCredentials), nil
+// JSON format. It returns an error if the file cannot be read, and will return
+// a nil value if neither environment variables are set.
+func getGoogleCredentials() (option.ClientOption, error) {
+	if envCredentials, isSet := os.LookupEnv(SopsGoogleCredentialsEnv); isSet {
+		credentials := []byte(envCredentials)
+		if _, err := os.Stat(envCredentials); err == nil {
+			if credentials, err = os.ReadFile(envCredentials); err != nil {
+				return nil, err
```
moritzschmitz-oviva, line 247: Perhaps don't exit here, but instead check `SopsGoogleCredentialsOAuthEnv` as well?

moritzschmitz-oviva: @warwick-mitchell1 I believe this is also done by the application default credentials: https://cloud.google.com/docs/authentication/application-default-credentials#search_order

warwick-mitchell1: You maybe right, this isn't the part I'm really adding with this PR though, this was already like this. I'm adding the Oauth token below which isn't covered by application-defaultcreds

moritzschmitz-oviva: I understand this. However, with your added functionality I believe it makes sense to also adjust the old behavior. If any fail condition happens with the credentials by file lookup, continue to check whether credentials by token area available before failing completely.

moritzschmitz-oviva (approved):

moritzschmitz-oviva:
Couldn't find anything in the Google docs, but the Terraform provider documentation provides a few more (and different) environment variables: https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#authentication-configuration

moritzschmitz-oviva:
Let's do this!

gcpkms/keysource.go
```diff
@@ -233,14 +234,32 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
+		if _, err := os.Stat(envToken); err == nil {
+			if token, err = os.ReadFile(envToken); err != nil {
+				return nil, err
+			}
+		}
+		tokenSource := oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: string(token)},
+		)
+		return option.WithTokenSource(tokenSource), nil
+	}
```
hiddeco, line 263: I think we should not check for a file here, but rather assume what's documented in https://cloud.google.com/sdk/docs/authorizing. Which only mentions:

> Set the `CLOUDSDK_AUTH_ACCESS_TOKEN` environment variable to the **access token value**.

dirsigler:
are there any blockers for this?

I currently also facing issues with SOPS using Googles access token.

mincua:
bumping this up, what's blocking it? need support for ODIC token auth

felixfontein:
It looks like #1578 is adding the same functionality.

PR #1214, opened 2023-05-22
gcpkms: fix WithCredentialsJSON to credentials

zchee:
gcpkms: fix `WithCredentialsJSON` to credentials.

The `key.credentialJSO` is already parsed, and this line condition is `if credentials != nil {`, so use the `credentials` variable.


zchee:
cc: @ajvb @joshkaplinsky

hiddeco:
Closing this in favor of #1155. Thank you :bow:

PR #1226, opened 2023-06-23
ocikms: Oracle Cloud KMS support

### LINKED ISSUE #981: Support for Oracle Cloud Vault / KMS
What can I say, their free tier is rather generous...

It would be great if there were support for Oracle Cloud KMS:
https://www.oracle.com/security/cloud-security/key-management/
https://www.oracle.com/security/cloud-security/key-management/faq

They uh, seem to call it "Cloud Infrastructure Vault", "Key Management Service", etc, somewhat interchangeably. In the Portal it's relegated to the far corner of the Security tab with just "Vault".

deblasis:
Hi there, I found myself needing this so I thought it was a good opportunity to contribute a little.

Happy to make the necessary changes, I tried to comply with style as much as possible. 

Fixes #981 

cc: @colemickens, @davidpinhas, @b-dean

## Basic Demo

https://github.com/mozilla/sops/assets/29378614/f68a7a34-95a0-4a1c-846c-c61f1a091d5a


Makefile
```diff
@@ -36,7 +36,7 @@ generate: keyservice/keyservice.pb.go
 	$(GO) generate
 
 %.pb.go: %.proto
-	protoc --go_out=plugins=grpc:. $<
+	protoc --go_out=. --go-grpc_out=require_unimplemented_servers=false:.  $<
```
deblasis, line 39: This should be the current way of generating files with protoc since 1.20 I believe.

The generated files are also split now as you can see below.

The flag `require_unimplemented_servers=false` is necessary to preserve the previous functionality.

Please note that I am using: 
image

I have also tried 
```
protoc-gen-go v1.23.0
protoc        v3.13.0
```
but I am still getting errors with the plugin settings.

I don't think this is an issue but just flagging for the reviewer

keyservice/keyservice.proto
```diff
@@ -1,5 +1,7 @@
 syntax = "proto3";
 
+option go_package = "/keyservice";
```
deblasis, line 3: Added this ad best practice and because I was getting errors because of this as well when doing `make generate`

jpark0910:
Is there anything else blocking this PR from being merged main for the 3.10.0 milestone release? is there an ETA on 3.10.0?

felixfontein:
I personally won't have time to review this anytime soon (see https://github.com/getsops/sops/pull/1097#issuecomment-2607904325). I'm not sure if anyone else from the maintainers has time either, but it might well be that this won't make it into 3.10 and will get moved to the 3.11 milestone (once that exists).

jacobhuemmer:
Any update?

ocikms/keysource.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func extractRefs(key *MasterKey) (string, string, error) {
+	parts := strings.Split(key.Ocid, ".")
+	if len(parts) != ocidParts {
+		return "", "", fmt.Errorf("OCID length is %s, expected %d", key.Ocid, ocidParts)
+	}
+	region := parts[3]
+	vault_ref := parts[4]
+	return region, vault_ref, nil
```
b-dean, line 85: @deblasis I'm not really sure what to do with this, because I assume the ocid in your demo video is real, but mine ends up with a different ocid and a different endpoint

My ocid is of this form:

```
ocid1.vault.oc1.iad.example.aaaaaREDACTED
```

which then my endpoint ends up being:
```
https://example-crypto.kms.us-ashburn-1.oraclecloud.com
```

with `example` there being what you call `vault_ref` here. 

So I'm not sure if we can extract the parts reliably from the key OCID. Having worked w/ OCI for a while this doesn't surprise me. They probably have several different iterations of vaults and the endpoints aren't consistent.

The best I can think of is either:
- also require the vault ocid and call [GetVault](https://docs.oracle.com/en-us/iaas/tools/go-sdk-examples/65.101.0/keymanagement/GetVault.go.html) to get the crypto endpoint
- use [ListVaults](https://docs.oracle.com/en-us/iaas/api/#/en/key/release/VaultSummary/ListVaults) to list all the vaults in the compartment, and then use the crypto management endpoint on each to call [ListKeys](https://pkg.go.dev/github.com/oracle/oci-go-sdk/v65@v65.101.0/keymanagement#KmsManagementClient.ListKeys) until you find the key that was specified by the sops args and then return the crypto endpoint url of the vault that has that key 🤮 

it'd be really nice to only have to say the key ocid, but that second option of loop through all of them seems horrible. any other ideas?

I suppose you could also have:
- try both versions of the crypto endpoint url (such as both `iad` and `us-ashburn-1`) with some sort of short timeout to figure out which is the correct endpoint.

b-dean (changes requested):
going to need some way to provide the correct crypto endpoint (or find it)

ocikms/keysource.go
b-dean: @deblasis, good news! there's an easy way to deal with this. behold:

```go
package main

import (
        "fmt"
        "strings"

        "github.com/oracle/oci-go-sdk/v65/common"
)

func main() {
        keys := []string{
                "ocid1.key.oc1.uk-london-1.asdadsasdagz5aacmg.abwgiljtjasdasdasdagugpfe7wrtngukihgkybqxcoozz7sbh6lq",
                "ocid1.key.oc1.iad.asdadsasdagz5aacmg.abwgiljtjasdasdasdagugpfe7wrtngukihgkybqxcoozz7sbh6lq",
        }

        for i, ocid := range keys {
                endpoint, err := endpoint(ocid)
                if err != nil {
                        fmt.Println(err.Error())
                } else {
                        fmt.Printf("key %d endpoint: %s\n", i, endpoint)
                }
        }
}

func endpoint(ocid string) (string, error) {
        region, vaultExt, err := extractRefs(ocid)
        if err != nil {
                return "", err
        }

        cryptoEndpointTemplate := common.StringToRegion(region).EndpointForTemplate("kms", "https://{vaultExt}-crypto.kms.{region}.{secondLevelDomain}")
        cryptoEndpoint := strings.Replace(cryptoEndpointTemplate, "{vaultExt}", vaultExt, 1)
        return cryptoEndpoint, nil
}

const ocidParts = 6

func extractRefs(ocid string) (string, string, error) {
        parts := strings.Split(ocid, ".")
        if len(parts) != ocidParts {
                return "", "", fmt.Errorf("OCID length is %s, expected %d", ocid, ocidParts)
        }
        region := parts[3]
        vaultExt := parts[4]
        return region, vaultExt, nil
}
```

which outputs the following:
```
$ go run main.go
key 0 endpoint: https://asdadsasdagz5aacmg-crypto.kms.uk-london-1.oraclecloud.com
key 1 endpoint: https://asdadsasdagz5aacmg-crypto.kms.us-ashburn-1.oraclecloud.com
```

based on the your `keysource.go` and this [example I found in the oci-go-sdk](https://github.com/oracle/oci-go-sdk/blob/2d0ed3015dc7a5fa5d954d0d51c0fa95dc04f42a/example/container_image_signing/utils.go#L323-L332)

b-dean:
so it does work now, but it's painfully slow. 

it took 361s to encrypt or decrypt `{"foo": "bar"}`. when I did the same thing with `oci kms crypto encrypt` cli, it was under a second.

deblasis:
> so it does work now, but it's painfully slow.
> 
> it took 361s to encrypt or decrypt `{"foo": "bar"}`. when I did the same thing with `oci kms crypto encrypt` cli, it was under a second.

Hi @b-dean

My fault.

I believe that OCI KMS tried Instance Principal auth even when you had valid credentials in env vars, causing the timeout hell you experienced.

Now it checks if env vars/config work first, skipping Instance Principal entirely. Same pattern Azure and GCP use. 

Your operations should go from 361s to microseconds. 🤞 

Since I was there, I stole other patterns from the other providers and improved testability.

b-dean:
@deblasis, your latest version works much better. 

I was thinking about the change you made in 4e025d8, and the composing provider is supposed to take a slice of things and then only call the method (like `p.TenancyOCID()`) if none of the earlier providers returned one. It'd be nice if the instance principal one was not conditional, and the early exist wasn't there either.

So .. I made this in `github.com/ontariosystems/oci-cli-env-provider@v0.2.0`

```go
p := ocep.LazyConfigProvider(auth.InstancePrincipalConfigurationProvider)
```

so it won't call the function to create the thing unless no other provider returns a valid `TenancyOCID()` (or whatever method is being called).

I also added a better composing provider that actually loops through everything for `AuthType()`, instead of just the first thing.

all together you could have:
```
// newIPProvider is a variable to allow tests to stub the Instance Principal provider factory
var newIPProvider = auth.InstancePrincipalConfigurationProvider

func configurationProvider() (common.ConfigurationProvider, error) {
	var providers []common.ConfigurationProvider

	// 1) Prefer the CLI-compatible envs used widely in CI/containers (envs only; no implicit fallbacks)
	providers = append(providers, ocep.OciCliEnvironmentConfigurationProvider())

	// 2) Native SDK envs (OCI_tenancy_ocid, OCI_user_ocid, OCI_fingerprint, OCI_private_key_path, OCI_region)
	providers = append(providers, common.ConfigurationProviderEnvironmentVariables("OCI", ""))

	// 3) File-based fallbacks
	if cfg := os.Getenv(OCICLIConfigFile); cfg != "" {
		if prof := os.Getenv(OCICLIProfile); prof != "" {
			if p, err := common.ConfigurationProviderFromFileWithProfile(cfg, prof, ""); err == nil {
				providers = append(providers, p)
			}
		} else {
			if p, err := common.ConfigurationProviderFromFile(cfg, ""); err == nil {
				providers = append(providers, p)
			}
		}
	}

	// 4) Instance principals for compute instances
	providers = append(providers, ocep.LazyConfigProvider(newIPProvider))	

	// 5) Always keep a last-resort default config provider at the end
	providers = append(providers, common.DefaultConfigProvider())

	return ocep.ComposingConfigProvider(providers), nil
```

I do wonder if the last to should be switched. the instance principal one *only* works on instances, where as the default works so long as you have things correct in `~/.oci/config` or whatever other default locations.

So that'd be:
1. `OCI_CLI` env
2. `OCI` env
3. config files from `OCI_CLI_CONFIG_FILE` and `OCI_CLI_PROFILE`
4. default config from sdk
5. instance profile config

thoughts?

deblasis:
Hey @b-dean! Thanks for the detailed suggestions and for adding those improvements to [github.com/ontariosystems/oci-cli-env-provider](https://github.com/ontariosystems/oci-cli-env-provider) v0.2.0 

About this, my personal preference would be not to have the dependency at all to be honest. I added it initially but I am not a fan of it. Nothing against your library (it's solid!). Just a matter of codebase hygiene if that makes sense.

I'd like a pointer from the maintainers on this matter as they might find that a totally self contained implementation is preferable over abstractions coming from external packages whenever possible. While in doubt currently we have a hybrid approach where for now we just use `ocep.OciCliEnvironmentConfigurationProvider`. 

https://github.com/getsops/sops/pull/1226/files#diff-a085f93310390109561334c42f9f26e612c342ded5b8318a80c71c8357f34728R28

Let's see what they say. I'd be happy to:

1. re-implement that here and remove the dependency entirely
or
2. to offload the configuration logic completely to the external package
or 
3. leave as is

My preference is on 1.

That being said, I went ahead and implemented the lazy evaluation pattern locally in a very similar fashion to your approach. I realized that the early exit was already handled by `common.ComposingConfigurationProvider` internally.

I did take your advice on the provider ordering as well. Runtime magic now is last resort, allowing for config overrides.

b-dean:
@deblasis , that makes sense. It'd be nice if Oracle would've just put some of this in their sdk to begin with. I'm fine if you want to copy parts of the code I wrote directly in here, if the sops maintainers would rather not add more libraries.

I mainly published that because I wanted to use the `OCI_CLI_*` env and not have to make `~/.oci/config` everywhere I'd like to use sops.

b-dean (approved):

deblasis:
> @deblasis , that makes sense. It'd be nice if Oracle would've just put some of this in their sdk to begin with. 

> I mainly published that because I wanted to use the `OCI_CLI_*` env and not have to make `~/.oci/config` everywhere I'd like to use sops.

I think the best thing you can do is to file a PR [upstream](https://github.com/oracle/oci-go-sdk/) to add native support for OCI_CLI_* env straight into the SDK! :)

deblasis:
Test failed due to a race condition originating from the SDK code itself.

I filed a PR over there with repro steps and a fix 

https://github.com/oracle/oci-go-sdk/pull/603

ianmuge:
Is this still being considered?

harrisonquick:
Hi all,

Also curious if this is still being considered. We have a production use case for OCI KMS support; @deblasis I see you're still engaged - would help with rebasing / conflicts / testing / other updated be useful? Any chance a maintainer could confirm whether this PR is still a candidate for merging?

deblasis:
Hi @harrisonquick, I think the PR is blocked on https://github.com/getsops/sops/issues/2016 

As far as I understand it, and my memory might be a bit clouded (it has been a while), the maintainers want to move towards a pluggable system where KMSs can be added without making changes to SOPS itself, understandable.

Not sure what it is blocking that effort currently.

If you don't need to be bleeding edge, just use my fork and tweak it as you with until the pluggable system is ready.
You can always pull from upstream any change into it and use it. That's the beauty of OSS.

I had a production use case too and it worked very well.

harrisonquick:
@deblasis Thank you for your update! I will try to dig in further on the pluggable architecture status / project direction, and will happily use your fork until that becomes available.

deblasis:
> @deblasis Thank you for your update! I will try to dig in further on the pluggable architecture status / project direction, and will happily use your fork until that becomes available.

You are welcome @harrisonquick, you actually inspired me into taking action so thank you! 
I went on and built something that might give shape to the final solution or become it. Everyone wins either way. 

All the details are here: https://github.com/getsops/sops/issues/2016#issuecomment-5456092313

PR #1245, opened 2023-07-07
add specific status code for sops errors

kaedwen:
I would like to introduce explicit handling of sops errors to find out what did go wrong. This is a simple change without  and we can know what went wrong during the run.


sops.go
```diff
@@ -58,17 +58,24 @@ import (
...
 
-func (e sopsError) Error() string {
-	return string(e)
+func (e SopsError) Error() string {
+	return e.message
 }
 
 // MacMismatch occurs when the computed MAC does not match the expected ones
-const MacMismatch = sopsError("MAC mismatch")
+var MacMismatch = &SopsError{10, "MAC mismatch"}
```
felixfontein, line 75: Please note that there is already another exit code defined in cmd/sops/codes/codes.go. Please use the constants from there.

sops.go
```diff
@@ -58,17 +58,24 @@ import (
...
 
 // MetadataNotFound occurs when the input file is malformed and doesn't have sops metadata in it
-const MetadataNotFound = sopsError("sops metadata not found")
+var MetadataNotFound = &SopsError{11, "sops metadata not found"}
```
felixfontein, line 78: There is no code yet for that, a new constant should be added to cmd/sops/codes/codes.go and used here.

felixfontein:
The commit history looks pretty bad, can you please try to rebase with the current `main` branch to get rid of all these extra commits? Thanks.

felixfontein:
@kaedwen ping

PR #1250, opened 2023-07-11
Revamp release automation

hiddeco:
This pull request changes the release automation so that it uses [GoReleaser](https://goreleaser.com).

The configuration for GoReleaser is in the file `.goreleaser.yaml`. The configuration for the GitHub Actions workflow is in the file `release.yml`.

This configuration is quite sophisticated, and ensures at least the following:

- The release is built for multiple platforms and architectures, including Linux, macOS, and Windows, and for both AMD64 and ARM64.
- The release includes multiple packages in Debian and RPM formats.
- For every binary, a corresponding SBOM is generated and published.
- For all binaries, a checksums file is generated and signed using [Cosign](https://docs.sigstore.dev/cosign/overview/) with GitHub OIDC.
- Both Debian and Alpine Docker multi-arch images are built and pushed to GitHub Container Registry and Quay.io.
- The container images are signed using [Cosign](https://docs.sigstore.dev/cosign/overview/) with GitHub OIDC.
- [SLSA provenance](https://slsa.dev/provenance/v0.2) metadata is generated for release artifacts and container images.

#### :warning: Notable changes

- The previous `Dockerfile` and `Dockerfile.alpine` have been removed in favor of Dockerfiles specifically crafted for GoReleaser in `.release/`. These Dockerfiles need a specially crafted build context (with pretty much just the binary in it), and do not work using a normal `docker build ...`.
- While the release artifacts have the intention to be as backwards compatible as possible in terms of file names, etc. Your mileage may vary here, more specifically the `.rpm` and `.deb` artifacts.
- The `.darwin` artifact is now a ["fat binary"](https://en.wikipedia.org/wiki/Fat_binary#Apple's_Universal_binary).

#### :eyes: Preview (on fork)

- https://github.com/hiddeco/sops/releases/tag/v3.8.0-rc.38

#### :factory_worker: Preview workflow (on fork)

- https://github.com/hiddeco/sops/actions/runs/5815114942
- https://github.com/hiddeco/sops/actions/runs/5826897259

#### :desktop_computer: Testing locally

Run `make release-snapshot` and view `dist/` contents.


hiddeco:
@getsops/maintainers this is ready for an early review. I need to update the PR message to include all details, but also really need to catch some sleep at the moment.

The thing to focus on in the preview is the file formats of the artifacts, in comparison to https://github.com/getsops/sops/releases/tag/v3.7.3. _Most_ of them should be fully backwards compatible (the binaries), but for the package formats I have reached the absolute limits of what I deemed possible.

In addition, it includes all the niceness of a checksum file signed with Cosign, SBOMs for the binaries (but not containers, which include a 1:1 copy of the binary &mdash; we _can_ add this later), and SLSA provenance for all artifacts uploaded to GitHub and the container images themselves.

hiddeco:
If desirable, I can tidy my commits per touched file (or something) before merge.

.goreleaser.yaml
```diff
@@ -0,0 +1,367 @@
...
+    goos: linux
+    goarch: arm64
+    ids:
+      - binary-linux
+    dockerfile: .release/Dockerfile
+    build_flag_templates:
+      - "--pull"
+      - "--platform=linux/arm64"
+      - "--label=org.opencontainers.image.created={{ .Date }}"
+      - "--label=org.opencontainers.image.name={{ .ProjectName }}"
```
devstein, line 196: I _think_ this should be 
```suggestion
      - "--label=org.opencontainers.image.title={{ .ProjectName }}"
```

if were following the  [OCI pre-defined keys](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

.goreleaser.yaml
```diff
@@ -0,0 +1,367 @@
...
+    goos: linux
+    goarch: amd64
+    ids:
+      - binary-linux
+    dockerfile: .release/Dockerfile
+    build_flag_templates:
+      - "--pull"
+      - "--platform=linux/amd64"
+      - "--label=org.opencontainers.image.created={{ .Date }}"
+      - "--label=org.opencontainers.image.name={{ .ProjectName }}"
```
devstein, line 178: Same `.title` comment

devstein (approved):
This is amazing @hiddeco  The test release worked perfectly for me 🥇 

One non-blocking comment/question

.goreleaser.yaml
hiddeco: You are absolutely right, great catch :bow:

onedr0p (approved):

sabre1041 (approved):
Really impressive @hiddeco!

LGTM

PR #1247, opened 2023-07-11
Rename Go module to `github.com/getsops/sops/v3`

hiddeco:
This PR renames the Go module from `go.mozilla.org/sops/v3` to
`github.com/getsops/sops/v3` without a major version bump, to align
with new stewardship.

For more information around this change, refer to
https://github.com/getsops/sops/issues/1246.

For a one-liner to change the `go.mod` and any import paths in your
Go project making use of this module, run:

```shell
find /path/to/repo -type f \( -name "*.go" -o -name "go.mod" \) -exec sed -i 's|go.mozilla.org/sops/v3|github.com/getsops/sops/v3|g' {} \;
# Or, on macOS:
find /path/to/repo -type f \( -name "*.go" -o -name "go.mod" \) -exec sed -i '' 's|go.mozilla.org/sops/v3|github.com/getsops/sops/v3|g' {} \;
```

Fixes #1246

PR #1244, opened 2023-07-06
Update various bits of documentation

hiddeco:
Bare minimum to incorporate:

- Organization name change
- CNCF entry (including Sandbox policies)
- Note around project's relationship to Mozilla
- Up-to-date(ish) information to be able to do a release

Essentially, what we really need is https://github.com/getsops/community/issues/9, and no reStructuredText.

_(Here's a gem of wisdom for your mental treasure chest: In the fascinating realm of RST, the ancient sages have decreed that the mystical art of combining bold text with a link is a feat as elusive as finding a unicorn eating ice cream.)_

PR #1254, opened 2023-08-02
Update repository URL from mozilla/sops to getsops/sops

tuxiqae:


devstein (commented):
Hi @tuxiqae thanks for making a PR to fix this! There is already an open PR for this renaming migration, https://github.com/getsops/sops/pull/1247 , so I'm going to close this.

tuxiqae:
The PR in question does not take care of the same issue.
PR #1247 renames the module name, PR #1254 (this one) renames the name of the repo itself.
It might be smarter to add my commit to PR #1247, but that's up to you, I do not think that those two things should depend on one another.

devstein:
@tuxiqae Thank you for clarifying 😅

felixfontein:
Isn't this already covered by #1244?

PR #1258, opened 2023-08-11
Improve Make targets and address various issues

hiddeco:
This improves the `Makefile` a bit more, while still not being picture perfect. Things that could still use some love:

- Testing keys should really not pollute the user their GnuPG home directory, but rather be e.g. imported to some project specific key ring.
- Binary for functional tests should not be stored in `./functional-tests/sops`. But instead put in `./bin/sops`, from where it would be used instead (while building using GoReleaser to match release build?)


devstein (approved):
LGTM

PR #1261, opened 2023-08-16
version: rewrite command to use GitHub endpoints

hiddeco:
Based on difficulties observed in https://github.com/getsops/sops/issues/1238#issuecomment-1621315939

This changes the logic of parsing the `version.go` file from a certain
branch to instead make use of the GitHub latest release redirect or
API[1] endpoints for checking if `sops` is on the latest version.

Detaching any future release of SOPS from specific file structures
and/or branches, and (theoretically) freeing it from the requirement of
having to bump the version in-code during release (as this is also done
using `-ldflags` during build). Were it not for the fact that we have
to maintain it for backwards compatibility.

[1]: https://docs.github.com/en/free-pro-team@latest/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,44 +1,57 @@
 package version
 
 import (
-	"bufio"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/blang/semver"
+	"github.com/hashicorp/go-cleanhttp"
 	"github.com/urfave/cli"
 )
 
-// Version represents the value of the current semantic version
+// Version represents the value of the current semantic version.
 var Version = "3.7.3"
 
-// PrintVersion handles the version command for sops
+// PrintVersion prints the current version of sops. If the flag
+// `--disable-version-check` is set, the function will not attempt
+// to retrieve the latest version from the GitHub API.
+//
+// If the flag is not set, the function will attempt to retrieve
+// the latest version from the GitHub API and compare it to the
+// current version. If the latest version is newer, the function
+// will print a message to stdout.
 func PrintVersion(c *cli.Context) {
-	out := fmt.Sprintf("%s %s", c.App.Name, c.App.Version)
+	out := strings.Builder{}
+
+	out.WriteString(fmt.Sprintf("%s %s", c.App.Name, c.App.Version))
+
 	if c.Bool("disable-version-check") {
-		out += "\n"
+		out.WriteString("\n")
 	} else {
-		upstreamVersion, err := RetrieveLatestVersionFromUpstream()
+		upstreamVersion, upstreamURL, err := RetrieveLatestReleaseVersion()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


version/version.go
```diff
@@ -61,31 +75,36 @@ func AIsNewerThanB(A, B string) (bool, error) {
 	return false, nil
 }
 
-// RetrieveLatestVersionFromUpstream gets the latest version from the source code at Github
+// RetrieveLatestVersionFromUpstream fetches the latest release version
+// from the GitHub API. The function returns the latest version as a string,
+// or an error if the request failed or the response could not be parsed.
 func RetrieveLatestVersionFromUpstream() (string, error) {
-	resp, err := http.Get("https://raw.githubusercontent.com/getsops/sops/master/version/version.go")
+	req, err := http.NewRequest("GET", "https://api.github.com/repos/getsops/sops/releases/latest", nil)
```
felixfontein, line 82: The main disadvantage of using this URL is that this can get you quickly rate-limited when calling this often enough from the same IP address.

Alternatively it is possible to `GET https://github.com/getsops/sops/releases/latest/` and extract the version from the `Location` header, **but** that is relying on GitHub not changing the behavior (https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases doesn't document it is a redirect for example, only that you can use the URL to get the latest release), and it did causes problems when the repository is renamed/moved (since it returns a 301 redirect to the other repo's latest URL then). On the other hand, it does not have a rate limit (or at least none I ever reached), similar to https://raw.githubusercontent.com/getsops/sops/...

Or in other words, no matter which method you use, it has drawbacks :)

hiddeco: In an initial version I looked at `GITHUB_TOKEN` and `GH_TOKEN` environment variables to try to prevent the rate limiting from happening.

I am aware of the other work around, but worried it may eventually break. Whereas this is a real API, and versioned.

felixfontein: Looks good to me! With the fallback it should be pretty robust, assuming that GitHub doesn't do some really crazy things :-)

version/version.go
```diff
@@ -1,51 +1,68 @@
...
+				if outdated {
+					out.WriteString(fmt.Sprintf("\n[info] a new version of sops (%s) is available, you can update by visiting: %s\n", upstreamVersion, upstreamURL))
+				} else {
+					out.WriteString(" (latest)\n")
+				}
 			}
 		}
 	}
-	fmt.Fprintf(c.App.Writer, "%s", out)
+	fmt.Fprintf(c.App.Writer, out.String())
```
felixfontein, line 49: I think
```suggestion
	fmt.Fprintf(c.App.Writer, "%s", out)
```
is better since it won't process `%` in `out.String()`, if these appear for some reason. These could have been part of `err` for example.

felixfontein (commented):
Besides that, LGTM!

PR #1262, opened 2023-08-16
Tidy deprecated things in code

hiddeco:


aes/cipher.go
```diff
@@ -172,7 +171,7 @@ func (c Cipher) Encrypt(plaintext interface{}, key []byte, additionalData string
 	case bool:
 		encryptedType = "bool"
 		// The Python version encodes booleans with Titlecase
-		plainBytes = []byte(strings.Title(strconv.FormatBool(value)))
+		plainBytes = []byte(fmt.Sprintf("%t", value))
```
felixfontein, line 174: Isn't this a behavior change? Before it was resulting in `True` and `False`, now it results in `true` and `false`.

Since [strings.Title](https://pkg.go.dev/strings#Title) is deprecated, how about writing this out?
```suggestion
		if (value) {
			plainBytes = []byte("True")
		} else {
			plainBytes = []byte("False")
		}
```
?

sops.go
```diff
@@ -727,7 +727,7 @@ func ToBytes(in interface{}) ([]byte, error) {
 	case float64:
 		return []byte(strconv.FormatFloat(in, 'f', -1, 64)), nil
 	case bool:
-		return []byte(strings.Title(strconv.FormatBool(in))), nil
+		return []byte(fmt.Sprintf("%t", in)), nil
```
felixfontein, line 730: Similarly here.

felixfontein (commented):
I don't really know the grpc/gcpkms/keyservice/kms code, so I mainly looked at the changes that don't require knowledge in that area.

aes/cipher.go
hiddeco: You are absolutely right.

hiddeco:
@felixfontein the remainders are all covered by tests, and closely following the deprecation notices yielded by the packages themselves.

felixfontein (approved):

PR #930, opened 2021-09-02
add configuration for stores

jamesgoodhouse:
this will allow for setting of parameters specific to each store, such as indentation level for YAML (https://github.com/mozilla/sops/issues/900).

PR #1273, opened 2023-09-08
Allow configuration of indentation for YAML and JSON stores

### LINKED ISSUE #900: Configurable indentation for yaml output in .sops.yaml
It would be great if we could configure our indentation for yaml in the .sops.yaml configuration file. Right now it's just hardcoded to set the indent to 4 all the time.

https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/stores/yaml/store.go#L322

Yaml spec does not dictate 4 spaces.
https://yaml.org/spec/1.2/spec.html#id2777534
```
In YAML block styles, structure is determined by indentation. In general, indentation is defined as a zero or more space characters at the start of a line.
```

### LINKED ISSUE #1028: To let `sops` emit YAML parser compatible .json files?
When `sops` is used to encrypt a .yaml file, the output is a .yaml file. When `sops` is used to encrypt any other file, the output is a `.json`file.

My colleagues were writing a check in Python to validate that all files that by a naming convention were supposed to be encrypted, was actually encrypted properly. To do that, we looked to parse them and look for a `sops` key as a basic check. Since all `sops` encrypted files are YAML or JSON, we figured we could use a YAML parser and it would work. But, it didn't. We had to use a YAML parser for the .yaml files and a JSON parser for the .json files.

It turns out that while almost any JSON is valid YAML, [JSON with tabs isn't valid YAML](https://yaml.org/faq.html). The `sops` generated .json files included tabs instead of spaces!

## Questions

__The main question__

Should `sops` keep emitting tabs in its emitted `.json` files, or could it switch to emitting spaces, and by doing so emit .json files that are both JSON and YAML parser compatible?

## Related code

From code inspection, this seemed relevant to understand how files are written.

https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/stores/json/store.go#L281-L300

## Related issues

- https://github.com/mozilla/sops/issues/434

### LINKED ISSUE #1048: Allow for specifying the indent character to be used and how many (tabs OR spaces!)
Hello,

We are using sops as part of a pipeline that originally generates json pretty-formatted files that are space-indented. Unfortunately when we then use sops, [it decides](https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/stores/json/store.go#L220) to indent using tabs, which can cause our diffs to be large and noisy.

We'd like sops to allow defining an indent character of choice and how many of those characters, similar to many other frameworks that work with JSON.

Ph0tonic:
This PR aims to finish the work started in #930, thanks to @jamesgoodhouse.

I added the use of the Indent param in the store as well as a new `--indent` option in the cli. 
Let me know what you think about it.

Fixes #900, fixes #1048 and fixes #1028

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1130,15 +1130,15 @@ Below is an example of publishing to Vault (using token auth with a local dev in
 Important information on types
 ------------------------------
 
-YAML and JSON type extensions
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+YAML, JSON, ENV and INI type extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 SOPS uses the file extension to decide which encryption method to use on the file
 content. ``YAML``, ``JSON``, ``ENV``, and ``INI`` files are treated as trees of data, and key/values are
 extracted from the files to only encrypt the leaf values. The tree structure is also
 used to check the integrity of the file.
 
-Therefore, if a file is encrypted using a specific format, it need to be decrypted
+Therefore, if a file is encrypted using a specific format, it needs to be decrypted
 in the same format. The easiest way to achieve this is to conserve the original file
 extension after encrypting a file. For example:
 
@@ -1165,11 +1165,14 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
 JSON indentation
 ~~~~~~~~~~~~~~~~
 
-``sops`` indent ``SOPS`` files by default using one ``tab``. However, you can change
-this default behaviour to use spaces by either using the additional ``--indent=2`` cli option or
-by configuring ``.sops.yaml`` with the code below. (value ``0`` is no indentation)
+SOPS indents ``JSON`` files by default using one ``tab``. However, you can change
+this default behaviour to use ``spaces`` by either using the additional ``--indent=2`` CLI option or
+by configuring ``.sops.yaml`` with the code below.
+
+The special value ``0`` disables indentation, and ``-1`` uses a single tab.
 
 .. code:: yaml
+
   stores:
       json:
           indent: 2
@@ -1177,11 +1180,12 @@ by configuring ``.sops.yaml`` with the code below. (value ``0`` is no indentatio
 YAML indentation
 ~~~~~~~~~~~~~~~~
 
-``sops`` indent ``YAML`` files by default using 4 spaces. However, you can change
-this default behaviour by either using the additional ``--indent=2`` cli option or
-by configuring ``.sops.yaml`` with :
+SOPS indents ``YAML`` files by default using 4 spaces. However, you can change
+this default behaviour by either using the additional ``--indent=2`` CLI option or
+by configuring ``.sops.yaml`` with:
 
 .. code:: yaml
+
   stores:
       yaml:
           indent: 2
--- a/README.rst
+++ b/README.rst
@@ -1162,8 +1162,8 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
 
     $ cat myfile.json | sops --input-type json --output-type json -d /dev/stdin
 
-JSON indentation
-~~~~~~~~~~~~~~~~
+JSON and JSON_binary indentation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 SOPS indents ``JSON`` files by default using one ``tab``. However, you can change
 this default behaviour to use ``spaces`` by either using the additional ``--indent=2`` CLI option or
@@ -1175,6 +1175,8 @@ The special value ``0`` disables indentation, and ``-1`` uses a single tab.
 
   stores:
       json:
+          indent: 2
+      json_binary:
           indent: 2
 
 YAML indentation
--- a/README.rst
+++ b/README.rst
@@ -1162,8 +1162,22 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
 
     $ cat myfile.json | sops --input-type json --output-type json -d /dev/stdin
 
+YAML indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``YAML`` files by default using 4 spaces. However, you can change
+this default behaviour be either using the additional ``--indent=2`` cli option or
+by configuring ``.sops.yaml`` with :
+
+.. code:: yaml
+  stores:
+      yaml:
+          indent: 2
+
+
 YAML anchors
 ~~~~~~~~~~~~
+
 SOPS only supports a subset of ``YAML``'s many types. Encrypting YAML files that
 contain strings, numbers and booleans will work fine, but files that contain anchors
 will not work, because the anchors redefine the structure of the file at load time.
--- a/README.rst
+++ b/README.rst
@@ -1166,8 +1166,8 @@ JSON indentation
 ~~~~~~~~~~~~~~~~
 
 ``sops`` indent ``SOPS`` files by default using one ``tab``. However, you can change
-this default behaviour to use spaces be either using the additional ``--indent=2`` cli option or
-by configuring ``.sops.yaml`` with :
+this default behaviour to use spaces by either using the additional ``--indent=2`` cli option or
+by configuring ``.sops.yaml`` with the code below. (value ``0`` is no indentation)
 
 .. code:: yaml
   stores:
@@ -1178,7 +1178,7 @@ YAML indentation
 ~~~~~~~~~~~~~~~~
 
 ``sops`` indent ``YAML`` files by default using 4 spaces. However, you can change
-this default behaviour be either using the additional ``--indent=2`` cli option or
+this default behaviour by either using the additional ``--indent=2`` cli option or
 by configuring ``.sops.yaml`` with :
 
 .. code:: yaml
--- a/README.rst
+++ b/README.rst
@@ -1162,6 +1162,18 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
 
     $ cat myfile.json | sops --input-type json --output-type json -d /dev/stdin
 
+JSON indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``SOPS`` files by default using one ``tab``. However, you can change
+this default behaviour to use spaces be either using the additional ``--indent=2`` cli option or
+by configuring ``.sops.yaml`` with :
+
+.. code:: yaml
+  stores:
+      json:
+          indent: 2
+
 YAML indentation
 ~~~~~~~~~~~~~~~~
 
@@ -1173,7 +1185,6 @@ by configuring ``.sops.yaml`` with :
   stores:
       yaml:
           indent: 2
-
 
 YAML anchors
 ~~~~~~~~~~~~
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -30,15 +30,10 @@ type ExampleFileEmitter interface {
 	EmitExample() []byte
 }
 
-type Configurable interface {
-	Configure(*config.Config)
-}
-
 // Store handles marshaling and unmarshaling from SOPS files
 type Store interface {
 	sops.Store
 	ExampleFileEmitter
-	Configurable
 }
 
 type storeConstructor = func(*config.StoresConfig) Store
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -10,6 +10,7 @@ import (
 	"github.com/getsops/sops/v3"
 	"github.com/getsops/sops/v3/cmd/sops/codes"
 	. "github.com/getsops/sops/v3/cmd/sops/formats"
+	"github.com/getsops/sops/v3/config"
 	"github.com/getsops/sops/v3/keys"
 	"github.com/getsops/sops/v3/keyservice"
 	"github.com/getsops/sops/v3/kms"
@@ -29,32 +30,37 @@ type ExampleFileEmitter interface {
 	EmitExample() []byte
 }
 
+type Configurable interface {
+	Configure(*config.Config)
+}
+
 // Store handles marshaling and unmarshaling from SOPS files
 type Store interface {
 	sops.Store
 	ExampleFileEmitter
+	Configurable
 }
 
-type storeConstructor = func() Store
+type storeConstructor = func(*config.StoresConfig) Store
 
-func newBinaryStore() Store {
-	return &json.BinaryStore{}
+func newBinaryStore(c *config.StoresConfig) Store {
+	return json.NewBinaryStore(&c.JSONBinary)
 }
 
-func newDotenvStore() Store {
-	return &dotenv.Store{}
+func newDotenvStore(c *config.StoresConfig) Store {
+	return dotenv.NewStore(&c.Dotenv)
 }
 
-func newIniStore() Store {
-	return &ini.Store{}
+func newIniStore(c *config.StoresConfig) Store {
+	return ini.NewStore(&c.INI)
 }
 
-func newJsonStore() Store {
-	return &json.Store{}
+func newJsonStore(c *config.StoresConfig) Store {
+	return json.NewStore(&c.JSON)
 }
 
-func newYamlStore() Store {
-	return &yaml.Store{}
+func newYamlStore(c *config.StoresConfig) Store {
+	return yaml.NewStore(&c.YAML)
 }
 
 var storeConstructors = map[Format]storeConstructor{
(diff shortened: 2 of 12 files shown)


cmd/sops/main.go
```diff
@@ -1051,11 +1055,35 @@ func keyservices(c *cli.Context) (svcs []keyservice.KeyServiceClient) {
...
+		// Ignore config not found errors returned from FindConfigFile since the config file is not mandatory
+		configPath, _ = config.FindConfigFile(".")
+	}
+	storesConf, _ := config.LoadStoresConfig(configPath)
+<<<<<<< Updated upstream
+=======
+	if context.Int("indent") != 0 {
+		storesConf.YAML.Indent = context.Int("indent")
+	}
+>>>>>>> Stashed changes
```
devstein, line 1084: Merge issue

Ph0tonic: Ho thanks, I forgot to save my file before committing. Fixed !

hiddeco:
In favor of adding support for this but not before 3.9.0.

cmd/sops/main.go
```diff
@@ -700,6 +700,10 @@ func main() {
 			Name:  "shamir-secret-sharing-threshold",
 			Usage: "the number of master keys required to retrieve the data key with shamir",
 		},
+		cli.IntFlag{
+			Name:  "indent",
+			Usage: "the number of space to indent yaml encoded file for encryption",
```
felixfontein, line 705:
```suggestion
			Usage: "the number of space to indent YAML encoded file for encryption",
```

cmd/sops/main.go
```diff
@@ -1051,11 +1055,32 @@ func keyservices(c *cli.Context) (svcs []keyservice.KeyServiceClient) {
...
 func inputStore(context *cli.Context, path string) common.Store {
-	return common.DefaultStoreForPathOrFormat(path, context.String("input-type"))
+	var configPath string
+	if context.String("config") != "" {
+		configPath = context.String("config")
+	} else {
+		// Ignore config not found errors returned from FindConfigFile since the config file is not mandatory
+		configPath, _ = config.FindConfigFile(".")
+	}
+	storesConf, _ := config.LoadStoresConfig(configPath)
```
felixfontein, line 1065: It looks to me that lines 1058 to 1065 could be moved to a function, since they appear twice in this file (here and in `outputStore()`).

cmd/sops/common/common.go
```diff
@@ -29,32 +30,37 @@ type ExampleFileEmitter interface {
 	EmitExample() []byte
 }
 
+type Configurable interface {
+	Configure(*config.Config)
+}
```
felixfontein, line 35: What is this used for? Maybe I missed something, but this never seems to be called, and all implementations don't do anything, not even the one for the YAML store.

felixfontein (commented):
Thanks for your contribution! I've some first comments after glancing over the implementation.

cmd/sops/main.go
Ph0tonic: Changed

cmd/sops/common/common.go
Ph0tonic: You are right, it's useless, good catch. It's a remainder of PR #930 from which I started. Maybe he had another idea in mind.

felixfontein:
For some reason this doesn't compile in CI anymore. The only error printed is `Error: `, which is not exactly helpful. I'm not sure whether this short error output is caused by GHA or by Go itself. In another PR, #1274, the compile steps worked, but the functional tests fail, also with `Error: `, so this *might* be related to GHA. Did you try to compile this locally?

felixfontein:
Meh, actually the problem was on my side - apparently GHA changed some code and the error reports are now loaded from another domain name, which I was blocking by default.

Ph0tonic:
> Meh, actually the problem was on my side - apparently GHA changed some code and the error reports are now loaded from another domain name, which I was blocking by default.

I found an issue in `main.go` which I fixed. 👍

felixfontein:
How about using tabs if no indent is specified, and the amount of spaces if `--indent` is specified (or the JSON indent value is configured in the config)?

It might also make sense to allow `0` as a valid value for JSON. (IIRC parts of the code assume that not specified == 0.)

Ph0tonic:
> How about using tabs if no indent is specified, and the amount of spaces if `--indent` is specified (or the JSON indent value is configured in the config)?
> 
> It might also make sense to allow `0` as a valid value for JSON. (IIRC parts of the code assume that not specified == 0.)

Yes, we could do that, I think it would make sense.
I just see one technical problem if we want to allow 0, I do not think it's possible to differentiate whether zero was provided during the unmarshaling of `.sops.yaml` or if it was empty and got the default 0 value.

Ph0tonic:
I applied what we discussed. However, while looking at other issues, #1028 seems to propose having default json indentation with spaces instead of tab to be compatible with yaml parser. I have no idea whether it is a good idea to change the default indentation of json to 4 spaces instead of one tab.
What do you think ?

felixfontein:
> However, while looking at other issues, #1028 seems to propose having default json indentation with spaces instead of tab to be compatible with yaml parser. I have no idea whether it is a good idea to change the default indentation of json to 4 spaces instead of one tab.

I tend a bit to keep the tabs. No idea what the other maintainers think though :)

In any case, as long as it is possible to explicitly configure indentation by tabs (for example by setting indentation to a negative number) I'm ok with the default changing, assuming there is a majority for changing it.

> I just see one technical problem if we want to allow 0, I do not think it's possible to differentiate whether zero was provided during the unmarshaling of `.sops.yaml` or if it was empty and got the default 0 value.

I would expect it is possible to provide a default value. I'm not that familiar with Go's serialization/deserialization mechanism though...

felixfontein:
I think you can simply assign default values to `conf` in https://github.com/getsops/sops/blob/d46b70272adf866fd25ff3f4356aa6c692acfaf6/config/config.go#L252-L253 before loading, like
```
	conf := &configFile{}
        conf.Stores.JSON.Indent = -1
	err = conf.load(confBytes)
```

Ph0tonic:
Great thank you for your feedback, I appreciate it 👍. I implemented a default value of -1 which allows to support no indentation at all if the user provides `0`. I think we can keep the tab as the default and then users can configure whatever they prefer. Let's see what other maintainers think of it.

cmd/sops/main.go
```diff
@@ -1050,12 +1054,31 @@ func keyservices(c *cli.Context) (svcs []keyservice.KeyServiceClient) {
...
 func inputStore(context *cli.Context, path string) common.Store {
-	return common.DefaultStoreForPathOrFormat(path, context.String("input-type"))
+	storesConf, _ := loadStoresConfig(context, path)
+	return common.DefaultStoreForPathOrFormat(storesConf, path, context.String("input-type"))
 }
 
 func outputStore(context *cli.Context, path string) common.Store {
-	return common.DefaultStoreForPathOrFormat(path, context.String("output-type"))
+	storesConf, _ := loadStoresConfig(context, path)
+	if context.Int("indent") != 0 {
```
felixfontein, line 1075: I think this should also work:
```suggestion
	if context.IsSet("indent") {
```
(see https://github.com/urfave/cli/blob/v1-maint/context.go#L55)

The main advantage is that this allows to pass `--indent 0` on CLI and make that override a non-zero indentation for JSON. (For YAML, it will obviously not work.)

README.rst
```diff
@@ -1144,8 +1144,35 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
 
     $ cat myfile.json | sops --input-type json --output-type json -d /dev/stdin
 
+JSON indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``SOPS`` files by default using one ``tab``. However, you can change
+this default behaviour to use ``spaces`` by either using the additional ``--indent=2`` cli option or
```
felixfontein, line 1151:
```suggestion
this default behaviour to use ``spaces`` by either using the additional ``--indent=2`` CLI option or
```

README.rst
```diff
@@ -1144,8 +1144,35 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
...
+.. code:: yaml
+  stores:
+      json:
+          indent: 2
+
+YAML indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``YAML`` files by default using 4 spaces. However, you can change
+this default behaviour by either using the additional ``--indent=2`` cli option or
```
felixfontein, line 1165:
```suggestion
this default behaviour by either using the additional ``--indent=2`` CLI option or
```

README.rst
```diff
@@ -1144,8 +1144,35 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
...
+by configuring ``.sops.yaml`` with :
```
felixfontein, line 1166:
```suggestion
by configuring ``.sops.yaml`` with:
```

felixfontein (commented):
I only noticed some small typographic details in the README, besides that it looks good to me!

README.rst
```diff
@@ -1144,8 +1144,35 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
...
+JSON indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``SOPS`` files by default using one ``tab``. However, you can change
+this default behaviour to use ``spaces`` by either using the additional ``--indent=2`` CLI option or
+by configuring ``.sops.yaml`` with the code below.
+
+The special value ``0`` disables indentation, and ``-1`` uses a single tab.
+
+.. code:: yaml
```
felixfontein, line 1156: Without that newline, all the content are parameters and the directive has no content, thus showing an empty code listing.
```suggestion
.. code:: yaml

```

README.rst
```diff
@@ -1144,8 +1144,35 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
...
+          indent: 2
+
+YAML indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``YAML`` files by default using 4 spaces. However, you can change
+this default behaviour by either using the additional ``--indent=2`` CLI option or
+by configuring ``.sops.yaml`` with:
+
+.. code:: yaml
```
felixfontein, line 1168:
```suggestion
.. code:: yaml

```

felixfontein (commented):
Sorry, noticed some more docs things ;-)

oaxiento:
It'd be great to see this get merged soon. We really need that feature. Thanks @Ph0tonic @felixfontein @devstein for your effort!

hiddeco:
The plan is to get a patch release out first, before this will be available in the `v3.9.0` release. Thanks for your patience :bow:

README.rst
```diff
@@ -1144,8 +1144,37 @@ When operating on stdin, use the ``--input-type`` and ``--output-type`` flags as
 
     $ cat myfile.json | sops --input-type json --output-type json -d /dev/stdin
 
+JSON indentation
+~~~~~~~~~~~~~~~~
+
+``sops`` indent ``SOPS`` files by default using one ``tab``. However, you can change
```
nogweii, line 1150: Uh, I believe this is supposed to be "sops indents *JSON* files", no?

Ph0tonic: Nice catch, thank you !

README.rst
```diff
@@ -203,7 +203,7 @@ configuration directory. On Linux, this would be ``$XDG_CONFIG_HOME/sops/age/key
 On macOS, this would be ``$HOME/Library/Application Support/sops/age/keys.txt``. On
 Windows, this would be ``%AppData%\sops\age\keys.txt``. You can specify the location
 of this file manually by setting the environment variable **SOPS_AGE_KEY_FILE**.
-Alternatively you can provide the the key(s) directly by setting the **SOPS_AGE_KEY**
+Alternatively, you can provide the key(s) directly by setting the **SOPS_AGE_KEY**
```
felixfontein, line 206: Can you move unrelated (to the functionality) fixes to another PR (like htis one), and squash some of the commits? Thanks!

felixfontein:
> The plan is to get a patch release out first, before this will be available in the `v3.9.0` release. Thanks for your patience 🙇

@hiddeco do you already know whether you want to do another patch release before starting to merge new features for 3.9.0?

README.rst
Ph0tonic: Done in #1339

hiddeco:
I think it should be fine to start merging new features, as I was planning on doing `v3.9.0` next.

hiddeco (commented):
It would be great if this could be rebased (to get rid of any merge commits), while also smashing some of the e.g. `Update README.rst` commits, etc. together.

For the rebase:

```sh
git remote add upstream https://github.com/getsops/sops
git fetch upstream && git rebase upstream/main
```

For tidying the commits (https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History):

```
git rebase -i upstream/main
```

Ph0tonic:
> It would be great if this could be rebased (to get rid of any merge commits), while also smashing some of the e.g. `Update README.rst` commits, etc. together.
> 
> For the rebase:
> 
> ```shell
> git remote add upstream https://github.com/getsops/sops
> git fetch upstream && git rebase upstream/main
> ```
> 
> For tidying the commits (https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History):
> 
> ```
> git rebase -i upstream/main
> ```

Thanks for helping, here it is :+1:

felixfontein:
Note that commit 07c06c7ac47f68b33b9ef9cbd7f66b6d29d1e953 contains some merge artefacts that are removed again in commit c7b04e1b156699d29393f5043cda1a732fdd7bbe. Would you mind rebasing again to clean up both commits so that the merge artefacts are not part of the commits added to the `main` branch?

Besides that, I think this PR is ready for merging. @hiddeco @devstein do you have further comments?

Ph0tonic:
> Note that commit [07c06c7](https://github.com/getsops/sops/commit/07c06c7ac47f68b33b9ef9cbd7f66b6d29d1e953) contains some merge artefacts that are removed again in commit [c7b04e1](https://github.com/getsops/sops/commit/c7b04e1b156699d29393f5043cda1a732fdd7bbe). Would you mind rebasing again to clean up both commits so that the merge artefacts are not part of the commits added to the `main` branch?
> 
> Besides that, I think this PR is ready for merging. @hiddeco @devstein do you have further comments?

Fixed ! 👍

stores/json/store.go
```diff
@@ -237,7 +249,11 @@ func (store Store) treeBranchFromJSON(in []byte) (sops.TreeBranch, error) {
 
 func (store Store) reindentJSON(in []byte) ([]byte, error) {
 	var out bytes.Buffer
-	err := json.Indent(&out, in, "", "\t")
+	indent := "\t"
+	if store.config.Indent != -1 {
+		indent = strings.Repeat(" ", store.config.Indent)
+	}
```
felixfontein, line 255:
```suggestion
	} else if store.config.Indent < -1 {
	    ... tell user that negative values != -1 are not OK ...
	}
```

config/config.go
```diff
@@ -63,9 +63,32 @@ func FindConfigFile(start string) (string, error) {
...
 
+type DotenvStoreConfig struct{}
+
+type INIStoreConfig struct{}
+
+type JSONStoreConfig struct{
+	Indent int `yaml:"indent"`
+}
+
+type JSONBinaryStoreConfig struct{}
```
felixfontein, line 74: Shouldn't this also have an `indent` setting?

felixfontein (commented):
While looking at this again I have some more comments (or maybe I forgot some context from before).

config/config.go
Ph0tonic: Yes, you are right, I will add this parameter

Ph0tonic: Here it is, I implemented it this way so we can have a different configuration for `json` and `json_binary`. Let me know what you think.

cmd/sops/main.go
```diff
@@ -1065,12 +1069,31 @@ func keyservices(c *cli.Context) (svcs []keyservice.KeyServiceClient) {
...
+	return common.DefaultStoreForPathOrFormat(storesConf, path, context.String("input-type"))
 }
 
 func outputStore(context *cli.Context, path string) common.Store {
-	return common.DefaultStoreForPathOrFormat(path, context.String("output-type"))
+	storesConf, _ := loadStoresConfig(context, path)
+	if context.IsSet("indent") {
+		indent := context.Int("indent")
+		storesConf.YAML.Indent = indent
+		storesConf.JSON.Indent = indent
```
felixfontein, line 1093:
```suggestion
		storesConf.JSON.Indent = indent
		storesConf.JSONBinary.Indent = indent
```

felixfontein (commented):
Besides that, looks good to me!

cmd/sops/main.go
Ph0tonic: Nice catch, thanks.

felixfontein (approved):

felixfontein:
If nobody objects, I'll merge this during the next few days.

hiddeco (approved):

felixfontein:
@jamesgoodhouse @Ph0tonic thanks a lot for working on this and getting it over the finish line! :)

consideRatio:
Wieee happy to learn that the json files are now being valid YAML files by no longer including tabs! Nice work on this @Ph0tonic @jamesgoodhouse and reviewers! :tada:

PR #1274, opened 2023-09-10
Allow to pass multiple paths to 'updatekeys'

sj14:
Allow to pass multiple paths to the `updatekeys` command, e.g. `sops updatekeys file1 file2 file3`.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -515,7 +515,7 @@ func main() {
 		},
 		{
 			Name:      "updatekeys",
-			Usage:     "update the keys of a SOPS file using the config file",
+			Usage:     "update the keys of SOPS files using the config file",
 			ArgsUsage: `file`,
 			Flags: append([]cli.Flag{
 				cli.BoolFlag{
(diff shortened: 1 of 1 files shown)


cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -23,30 +22,34 @@ type Opts struct {
...
-		return err
-	}
-	info, err := os.Stat(path)
-	if err != nil {
-		return err
-	}
-	if info.IsDir() {
-		return fmt.Errorf("can't operate on a directory")
+func UpdateKeys(inputPaths []string, opts Opts) error {
+	for _, inputPath := range inputPaths {
```
hiddeco, line 26: This changes a publicly exposed function breaking backwards compatibility. To avoid this, the loop should for example be dealt with in `main.go`, or a new function should be introduced.

sj14: Adjusted.

felixfontein (approved):
Looks good to me.

cmd/sops/main.go
```diff
@@ -528,18 +528,20 @@ func main() {
...
+						KeyServices: keyservices(c),
+						Interactive: !c.Bool("yes"),
+						ConfigPath:  configPath,
+						InputType:   c.String("input-type"),
+					})
+					if cliErr, ok := err.(*cli.ExitError); ok && cliErr != nil {
+						return cliErr
+					} else if err != nil {
+						return common.NewExitError(err, codes.ErrorGeneric)
+					}
```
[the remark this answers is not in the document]
sj14: Adjusted.

felixfontein (commented):
The current version makes the return value less usable. I'm not sure whether anyone relied on the return values of the `updatekeys` command, but in case any did, they are now broken.

How about slightly adjusting the behavior so that in case there is one argument, the old code is used, and if there are at least two arguments, the new code is used?

sj14:
> How about slightly adjusting the behavior so that in case there is one argument, the old code is used, and if there are at least two arguments, the new code is used?

Adjusted. I'm also now printing the amount of failed key updates when multiple arguments are used. I hope that's fine.

felixfontein (approved):

hiddeco (approved):

felixfontein:
@sj14 thanks for your contribution!

PR #1277, opened 2023-09-11
Enable Dependabot for Docker, GitHub Actions and Go Mod

hiddeco:
This enables Dependabot using two groups, one for GitHub Actions and one for Go Modules.

In the future, we may want to split the Go Modules into multiple groups. For example, one for each key source with a misc catch-all group for any other dependency.

xref: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-version-updates-into-one-pull-request

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,41 @@
+version: 2
+
+updates:
+  - package-ecosystem: "docker"
+    directory: "/"
+    labels: ["dependencies"]
+    schedule:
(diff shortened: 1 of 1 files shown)


devstein (approved):
Thanks for adding 🙌

felixfontein (approved):
Should probably be merged after #1276 so that it doesn't create a PR for GHA first :)

PR #1281, opened 2023-09-12
keyservices: address logging regression

hiddeco:
Replace the logging of failed encryption and decryption attempts from
error to info level.

This to address a regression in which an encryption or decryption
attempt with a series of keys would result in a list of failed attempts
logged to stderr even when the operation itself eventually succeeded.

xref: https://github.com/getsops/sops/discussions/1270#discussioncomment-6883254

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -11,8 +11,8 @@ import (
 
 	"filippo.io/age"
 	"filippo.io/age/armor"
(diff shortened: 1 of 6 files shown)


devstein (approved):
LGTM

felixfontein (approved):

PR #1286, opened 2023-09-15
Better error reporting for missing gpg binary

makkes:
The error returned by `gpgExec` has just been swallowed. Now it is stringified and returned together with any output to stderr.

### DIFF
--- a/pgp/keysource.go
+++ b/pgp/keysource.go
@@ -132,17 +132,17 @@ func (d GnuPGHome) Import(armoredKey []byte) error {
 	}
 
 	args := []string{"--batch", "--import"}
-	_, stderrBuf, err := gpgExec(d.String(), args, bytes.NewReader(armoredKey))
+	_, stderr, err := gpgExec(d.String(), args, bytes.NewReader(armoredKey))
 	if err != nil {
-		stderr := stderrBuf.String()
+		stderrStr := strings.TrimSpace(stderr.String())
 		errStr := err.Error()
 		var sb strings.Builder
 		sb.WriteString("failed to import armored key data into GnuPG keyring")
-		if len(stderr) > 0 {
-			fmt.Fprintf(&sb, ": %s", stderr)
+		if len(stderrStr) > 0 {
 			if len(errStr) > 0 {
-				fmt.Fprintf(&sb, ": %s", errStr)
+				fmt.Fprintf(&sb, " (%s)", errStr)
 			}
+			fmt.Fprintf(&sb, ": %s", stderrStr)
 		} else if len(errStr) > 0 {
 			fmt.Fprintf(&sb, ": %s", errStr)
 		}
--- a/pgp/keysource.go
+++ b/pgp/keysource.go
@@ -9,6 +9,7 @@ package pgp //import "github.com/getsops/sops/v3/pgp"
 import (
 	"bytes"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


makkes:
@hiddeco I can't directly add you as a review but would you mind reviewing this so we can close out https://github.com/fluxcd/kustomize-controller/issues/966, pls?

pgp/keysource_test.go
```diff
@@ -24,6 +24,21 @@ var (
 	mockFingerprint = "B59DAF469E8C948138901A649732075EA221A7EA"
 )
 
+func TestGnuPGHome_Import_With_Missing_Binary(t *testing.T) {
```
hiddeco, line 27: Should be placed below `TestGnuPGHome_Import`.

hiddeco: Also wondering what a failed import looks like now with a working GnuPG binary path, as I am not sure what gets logged to `stdout`.

makkes: Good point. I added an assertion to the existing case `TestGnuPGHome_Import`.

pgp/keysource.go
```diff
@@ -133,7 +133,8 @@ func (d GnuPGHome) Import(armoredKey []byte) error {
 	args := []string{"--batch", "--import"}
 	err, _, stderr := gpgExec(d.String(), args, bytes.NewReader(armoredKey))
 	if err != nil {
-		return fmt.Errorf("failed to import armored key data into GnuPG keyring: %s", strings.TrimSpace(stderr.String()))
+		return fmt.Errorf("failed to import armored key data into GnuPG keyring: %s (stderr: %s)",
+			err, strings.TrimSpace(stderr.String()))
```
hiddeco, line 137:
```suggestion
		return fmt.Errorf("failed to import armored key data into GnuPG keyring: %s (%s)",
			strings.TrimSpace(stderr.String()), err)
```

would be better in this cause I think.

hiddeco: Other option would be to do something like `<stderr>: <err>`.

makkes: I made it one of:

- `failed to import armored key data into GnuPG keyring`
- `failed to import armored key data into GnuPG keyring: <stderr>`
- `failed to import armored key data into GnuPG keyring: <err>`
- `failed to import armored key data into GnuPG keyring: <stderr>: <err>`

In reality, when stderr is non-empty then err will likely also not be empty but now we're safe in the case where it is.

hiddeco (dismissed):
Thank you @makkes :1st_place_medal:

pgp/keysource_test.go
```diff
@@ -66,10 +66,27 @@ func TestGnuPGHome_Import(t *testing.T) {
 	err, _, stderr = gpgExec(gnuPGHome.String(), []string{"--list-secret-keys", mockFingerprint}, nil)
 	assert.NoErrorf(t, err, stderr.String())
 
-	assert.Error(t, gnuPGHome.Import([]byte("invalid armored data")))
+	err = gnuPGHome.Import([]byte("invalid armored data"))
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "gpg: no valid OpenPGP data found.\ngpg: Total number processed: 0\n: exit status 2")
```
hiddeco, line 71:
```
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
: exit status 2
```

Does still look far from optimal :grimacing:

makkes: Yeah. I mean, we can just strip all newlines but I didn't want to change too much.

hiddeco (approved):
Slightly tuned and the message formatting a bit, and think this is the least unpleasant way to deal with this for now.

Thanks a lot for taking care of this @makkes :bow:

PR #1287, opened 2023-09-16
Add linting for RST and MD files

felixfontein:
Uses [rstcheck](pypi.org/project/rstcheck/ ) and [markdownlint](https://github.com/markdownlint/markdownlint).

I don't have much experience with markdownlint, but it seems to work well and seems to be useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ SYFT_VERSION        ?= v0.87.0
 GORELEASER          := $(BIN_DIR)/goreleaser
 GORELEASER_VERSION  ?= v1.20.0
 
+RSTCHECK            := $(shell command -v rstcheck)
+MARKDOWNLINT        := $(shell command -v mdl)
+
 export PATH := $(BIN_DIR):$(PATH)
 
 .PHONY: all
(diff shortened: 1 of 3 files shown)


hiddeco:
Think it would be good if these tools were available in the `Makefile` as well, to allow people to run them before CI catches things.

onedr0p:
I am not sure what the github action is doing, should it open a PR with the changes? or when a pull request is opened or updated run these tools and commit back to the PR branch?

felixfontein:
@onedr0p right now the GHA does 1.; I'm not aware of any way to do 2. or 3.

> I tend to go with option 1 but it leaves the burden up to the PR author to fix the issues manually.

I think that's acceptable, as long as we help them if they are not familiar with the doc formats (RST is especially error-prone, which IMO makes it even more important to add some validation).

hiddeco:
Think 1 (or 4) is also the most desirable because linters can be wrong at times, as their observation lacks human reasoning which in terms of formatting may be different (for example, because of readability in edge cases). Otherwise, you may end up fighting a machine.

felixfontein:
Should I move out the docs changes into another PR? Might make reviewing / discussing easier.

Makefile
```diff
@@ -45,6 +45,20 @@ vendor:
...
 	$(GO) vet ./...
 
+
+.PHONY: checkdocs
+checkdocs: checkrst checkmd
+
+.PHONY: checkrst
+RST_FILES=$(shell find . -name '*.rst' | grep -v /vendor/ | sort)
+checkrst: $(RST_FILES)
+	rstcheck $^
```
hiddeco: Would probably be good to have some type of `command -v` check here, with instructions on how to install these things (or targets to install them into `BIN_DIR` in case of absence).

felixfontein: I've added a check in 8e1da49381e49d7104020da967cc1ab988cbeb2b.

hiddeco:
@felixfontein this requires a rebase now that #1320 is merged. :sparkles:

hiddeco (approved):

hiddeco (approved):
Thank you @felixfontein 🙇

felixfontein:
@hiddeco @onedr0p thanks for reviewing and merging!

PR #1288, opened 2023-09-16
Delete sops encrypted file we don't have keys for

### LINKED ISSUE #977: Missed key for some encrypted files
The issue with files (as example) like this:

https://github.com/mozilla/sops/blob/develop/stores/json/test_resources/example.json

File encrypted with GPG, where the FP: `E5297818703249D0C60E19E6824612478D1A4CCD`, which doesn't exist anywhere.

It is not possible to work on unit tests (for https://github.com/mozilla/sops/pull/887) without having that file decrypted.

Discussion: https://github.com/mozilla/sops/discussions/970

felixfontein:
While looking at #977 I tried to figure out which test actually uses that file. I wasn't able to find one, so I decided to just delete it and see what CI says. If we're lucky that solves the problem. If not, at least I get a better idea which test needs it.

Fixes #977.


hiddeco (approved):
We can always restore it from the Git history if we ever get second thoughts

felixfontein:

PR #1289, opened 2023-09-16
Improve handling of errors when binary store handles bad data

### LINKED ISSUE #596: SOPS panics when input type is not binary but output type is and input contains a 'data' mapping key.
Sops (latest version) panics, when decrypting yaml file with explicit type specification. Same file with .yaml extension is decrypted successfully.

```
[anton@vm ]$ diff auth.yaml.enc2 auth.enc.yaml

[anton@vm ]$ sops --version
sops 3.5.0 (latest)

[anton@vm ]$ sops -d auth.enc.yaml >/dev/null ; echo $?
0

[anton@vm ]$ sops --input-type yaml auth.yaml.enc2
panic: interface conversion: interface {} is sops.TreeBranch, not string

goroutine 1 [running]:
go.mozilla.org/sops/v3/stores/json.BinaryStore.EmitPlainFile(0xc00036b100, 0x1, 0x1, 0x242, 0x1b6, 0xc000010030, 0x0, 0x0)
	/home/ajvb/gocode/src/go.mozilla.org/sops/stores/json/store.go:50 +0x156
main.editTree(0x1211f80, 0xc000370ba0, 0x1226bc0, 0x190eba8, 0x1226b00, 0x190eba8, 0xc00010e640, 0x41, 0x0, 0xc00036c120, ...)
	/home/ajvb/gocode/src/go.mozilla.org/sops/cmd/sops/edit.go:126 +0xbbe
main.edit(0x1211f80, 0xc000370ba0, 0x1226bc0, 0x190eba8, 0x1226b00, 0x190eba8, 0xc00010e640, 0x41, 0x0, 0xc00036c120, ...)
	/home/ajvb/gocode/src/go.mozilla.org/sops/cmd/sops/edit.go:105 +0x2c5
main.main.func8(0xc000128dc0, 0x0, 0x0)
	/home/ajvb/gocode/src/go.mozilla.org/sops/cmd/sops/main.go:767 +0x23c7
gopkg.in/urfave/cli%2ev1.HandleAction(0xe62ca0, 0x10603e0, 0xc000128dc0, 0x0, 0x0)
	/home/ajvb/gocode/pkg/mod/gopkg.in/urfave/cli.v1@v1.20.0/app.go:490 +0xc8
gopkg.in/urfave/cli%2ev1.(*App).Run(0xc0002db860, 0xc0000d4000, 0x4, 0x4, 0x0, 0x0)
	/home/ajvb/gocode/pkg/mod/gopkg.in/urfave/cli.v1@v1.20.0/app.go:264 +0x58c
main.main()
	/home/ajvb/gocode/src/go.mozilla.org/sops/cmd/sops/main.go:823 +0x2de4

```

```
cat auth.enc.yaml 
apiVersion: ENC[AES256_GCM,data:igo=,iv:2TZajuaEplwhxOLoc2wo/T9CbPip1cVh1J/uw5mcUR8=,tag:6Np390u1MMouj4n3TyhCww==,type:str]
data:
    auth: ENC[AES256_GCM,data:Ghp/DXmLL98ujqDJ1rJMyOlVdGL9KtUPigpfOzzpRBxBiF0lGv3aCZQOyLAfZGdi9UwZAUJomo4=,iv:OdglsUCjRmW9nZ7BK+DIj0tgipD+UwFLzeeUo6TtWqg=,tag:jpdAL3JCTPSMpZCtg8/Qww==,type:str]
kind: ENC[AES256_GCM,data:H16uOwEp,iv:WYnXYG0/AheXt7mWB33idNl1VAy7VF3pmpRPaiZd634=,tag:ThEFBFmTYZylv27QokOhKw==,type:str]
metadata:
    creationTimestamp: null
    name: ENC[AES256_GCM,data:uitF6Yx0lSs=,iv:4wZfhMtOdvET4q8hcWXpVfil2piv6UX4kK2OOlj/ekY=,tag:QHWcgibKq2sgMjPXFqV/RA==,type:str]
sops:
    kms: []
    gcp_kms:
    -   resource_id: xxx
        created_at: '2019-12-18T21:17:49Z'
        enc: CiQA8Xd3G6oq6KSE0uLyRP2trk24LeMP7GhcINOf3oYcffYGggUSTBJKCgwV1Dc3s3RQrBnikj0SKD35HZm0UxNj/ubXrA7HipNIabReUTRHFm19ghVNuEhEJzwCkVWxvSkaEP1kDwpX7TBPoejP4VW2Eso=
    azure_kv: []
    lastmodified: '2019-12-18T21:17:49Z'
    mac: ENC[AES256_GCM,data:w1BLQbnPT0tHDElqsrCMC5NICG/rbkxVGxZgDFmjgFbFkWFOg6A3UoqhXMHTiiAmyBEDXLHk4pPtzRX7O/67+GRR3DwaM5Lt8AaYB0AkT7O60TCWx23IzCu/jdnrj9K7rCrqNh+Fbdmihv6g2U6wjG0GuTx2dsDTRdxPB15juvE=,iv:ushcpUzLozcklWBQYPNtHq1ZltPzr994nch5yQ0FmMY=,tag:nF85o6scy4krdQiNIjzmGQ==,type:str]
    pgp: []
    unencrypted_suffix: _unencrypted
    version: 3.4.0
```

felixfontein:
When the binary store tries to write data and `data` is not there, or `data` is not a string, or zero or multiple tree branches have been provided, there's now a nice error also explaining about `--output-format`. (The only error handled so far was `data` not being there in the first tree branch, but without that `--output-format` hint.)

Fixes #596.

### DIFF
--- a/cmd/sops/decrypt.go
+++ b/cmd/sops/decrypt.go
@@ -1,13 +1,18 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/getsops/sops/v3"
 	"github.com/getsops/sops/v3/cmd/sops/codes"
 	"github.com/getsops/sops/v3/cmd/sops/common"
 	"github.com/getsops/sops/v3/keyservice"
+	"github.com/getsops/sops/v3/stores/json"
 )
+
+const notBinaryHint = ("This is likely not an encrypted binary file?" +
+	" If not, use --output-type to select the correct output type.")
 
 type decryptOpts struct {
 	Cipher      sops.Cipher
@@ -45,6 +50,9 @@ func decrypt(opts decryptOpts) (decryptedFile []byte, err error) {
 		return extract(tree, opts.Extract, opts.OutputStore)
 	}
 	decryptedFile, err = opts.OutputStore.EmitPlainFile(tree.Branches)
+	if errors.Is(err, json.BinaryStoreEmitPlainError) {
+		err = fmt.Errorf("%s\n\n%s", err.Error(), notBinaryHint)
+	}
 	if err != nil {
 		return nil, common.NewExitError(fmt.Sprintf("Error dumping file: %s", err), codes.ErrorDumpingTree)
 	}
@@ -59,6 +67,9 @@ func extract(tree *sops.Tree, path []interface{}, outputStore sops.Store) (outpu
 	if newBranch, ok := v.(sops.TreeBranch); ok {
 		tree.Branches[0] = newBranch
 		decrypted, err := outputStore.EmitPlainFile(tree.Branches)
+		if errors.Is(err, json.BinaryStoreEmitPlainError) {
+			err = fmt.Errorf("%s\n\n%s", err.Error(), notBinaryHint)
+		}
 		if err != nil {
 			return nil, common.NewExitError(fmt.Sprintf("Error dumping file: %s", err), codes.ErrorDumpingTree)
 		}
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -3,6 +3,7 @@ package json //import "github.com/getsops/sops/v3/stores/json"
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 
@@ -42,15 +43,24 @@ func (store BinaryStore) EmitEncryptedFile(in sops.Tree) ([]byte, error) {
 	return store.store.EmitEncryptedFile(in)
 }
 
+var BinaryStoreEmitPlainError = errors.New("error emitting binary store")
+
 // EmitPlainFile produces plaintext json file's bytes from its corresponding sops.TreeBranches object
 func (store BinaryStore) EmitPlainFile(in sops.TreeBranches) ([]byte, error) {
+	if len(in) != 1 {
+		return nil, fmt.Errorf("%w: there must be exactly one tree branch", BinaryStoreEmitPlainError)
+	}
 	// JSON stores a single object per file
 	for _, item := range in[0] {
 		if item.Key == "data" {
-			return []byte(item.Value.(string)), nil
+			if value, ok := item.Value.(string); ok {
+				return []byte(value), nil
+			} else {
+				return nil, fmt.Errorf("%w: 'data' key in tree does not have a string value", BinaryStoreEmitPlainError)
+			}
 		}
 	}
-	return nil, fmt.Errorf("No binary data found in tree")
+	return nil, fmt.Errorf("%w: no binary data found in tree", BinaryStoreEmitPlainError)
 }
 
 // EmitValue extracts a value from a generic interface{} object representing a structured set
--- a/stores/json/store_test.go
+++ b/stores/json/store_test.go
@@ -3,8 +3,8 @@ package json
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/getsops/sops/v3"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestDecodeJSON(t *testing.T) {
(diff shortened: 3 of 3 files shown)


stores/json/store.go
```diff
@@ -42,15 +42,25 @@ func (store BinaryStore) EmitEncryptedFile(in sops.Tree) ([]byte, error) {
 	return store.store.EmitEncryptedFile(in)
 }
 
+const NOT_BINARY_HINT = ("This is likely not an encrypted binary file?" +
```
hiddeco, line 45: https://google.github.io/styleguide/go/decisions.html#constant-names

stores/json/store.go
```diff
@@ -42,15 +42,25 @@ func (store BinaryStore) EmitEncryptedFile(in sops.Tree) ([]byte, error) {
...
+	}
 	// JSON stores a single object per file
 	for _, item := range in[0] {
 		if item.Key == "data" {
-			return []byte(item.Value.(string)), nil
+			if value, ok := item.Value.(string); ok {
+				return []byte(value), nil
+			} else {
+				return nil, fmt.Errorf("'data' key in tree does not have a string value. " + NOT_BINARY_HINT)
+			}
```
hiddeco, line 60:
```suggestion
			if value, ok := item.Value.(string); ok {
				return []byte(value), nil
			}
			return nil, fmt.Errorf("'data' key in tree does not have a string value. " + NOT_BINARY_HINT)
```

stores/json/store.go
```diff
@@ -42,15 +42,25 @@ func (store BinaryStore) EmitEncryptedFile(in sops.Tree) ([]byte, error) {
 	return store.store.EmitEncryptedFile(in)
 }
 
+const NOT_BINARY_HINT = ("This is likely not an encrypted binary file?" +
+	" If not, use --output-type to select the correct output type.")
```
hiddeco, line 46: As the store can be used while using SDK, mentioning of a flag can be misleading.

stores/json/store.go
felixfontein: I guess I'm too used to const notation in C++, Java, Python, ... ;-)

stores/json/store.go
felixfontein: I guess then it's better to use a specialized error, which allows to print the message when returned in the CLI.

I'll update the PR accordingly (though likely not today :) ).

stores/json/store.go
```diff
@@ -42,15 +42,30 @@ func (store BinaryStore) EmitEncryptedFile(in sops.Tree) ([]byte, error) {
...
+}
+
+func (e BinaryStoreEmitPlainError) Error() string {
+	return e.message
+}
+
 // EmitPlainFile produces plaintext json file's bytes from its corresponding sops.TreeBranches object
 func (store BinaryStore) EmitPlainFile(in sops.TreeBranches) ([]byte, error) {
+	if len(in) != 1 {
+		return nil, BinaryStoreEmitPlainError{message: "There must be exactly one tree branch"}
```
hiddeco, line 56: Just for future reference for other Go projects, and not applicable here:

1. Normally in Go, error strings should not be capitalized (https://github.com/golang/go/wiki/CodeReviewComments#error-strings). However, that ship has sailed in this code base :-).
2. The more idiomatic name would be something along the lines of `ErrEmitPlain`. However, I can see you probably based this on `NotImplementedError`, `decryptGroupError`, etc. already existing in the code base.

hiddeco (approved):
Last commit requires as sign-off, and typically commit summaries do not end with a dot. Other than these nitpicks, it looks good to me :bow:

cmd/sops/decrypt.go
```diff
@@ -45,6 +49,9 @@ func decrypt(opts decryptOpts) (decryptedFile []byte, err error) {
 		return extract(tree, opts.Extract, opts.OutputStore)
 	}
 	decryptedFile, err = opts.OutputStore.EmitPlainFile(tree.Branches)
+	if bsepe, ok := err.(json.BinaryStoreEmitPlainError); ok {
+		err = fmt.Errorf("%s. %s", bsepe.Error(), notBinaryHint)
+	}
```
holiman, line 54: Why not use `errors.Is(err, json.BinaryStoreEmitPlainError)`, and define the latter as , for example
```
var BinaryStoreEmitPlainError = errors.New("error emitting binary store")
```
And when you throw it, you do ,for example
```
return nil, fmt.Errorf("%w: 'data' key in tree does not have a string value", BinaryStoreEmitPlainError)
```
Then you don't need a custom error-type

felixfontein: The why answer is easy: because I've never seen that before :) This looks a lot nicer, I've switched to it in 51956d9034498632957cc878c3ef8205f68b58a5.

cmd/sops/decrypt.go
```diff
@@ -59,6 +67,9 @@ func extract(tree *sops.Tree, path []interface{}, outputStore sops.Store) (outpu
 	if newBranch, ok := v.(sops.TreeBranch); ok {
 		tree.Branches[0] = newBranch
 		decrypted, err := outputStore.EmitPlainFile(tree.Branches)
+		if errors.Is(err, json.BinaryStoreEmitPlainError) {
+			err = fmt.Errorf("%s. %s", err.Error(), notBinaryHint)
```
hiddeco, line 71:
```suggestion
			err = fmt.Errorf("%s\n\n%s", err.Error(), notBinaryHint)
```

maybe?

In addition, I just noticed that in combination with the `common.NewExitError` below, this becomes `Error dumping file: Error emitting binary store: ...`. Which makes me wonder if we should prefer a Go idiomatic error as `BinaryStoreEmitPlainError` (resulting in `Error dumping file: error ...`), even if this would diverge from other errors in the code base.

But this is just nitpicking at this point really.

felixfontein: Yes, sounds good. Adjusted in 49a4145d34bea2cd9cfb3fdfa1518c6102eb1094. The CLI output now looks like:
```
Error dumping file: error emitting binary store: no binary data found in tree

This is likely not an encrypted binary file? If not, use --output-type to select the correct output type.
```

hiddeco (approved):
Awesome, thank you @felixfontein 🥇

felixfontein:
@hiddeco should I squash the commits?

hiddeco:
Probably best, yes.

felixfontein:
@hiddeco @holiman thanks a lot for reviewing!

PR #1291, opened 2023-09-16
On macOS, prefer XDG_CONFIG_HOME over os.UserConfigDir()

### LINKED ISSUE #1263: sops does not respect explicit XDG_CONFIG_HOME
I am on MacOS, and I have explicitly set `XDG_CONFIG_HOME` in my environment because `/Users/me/Library/Application Support` is an awkward location to store config files and `$HOME/.config` is much better. By setting this explicitly, applications should respect it.

However:

```
$ sops  <encrypted file>

Group 0: FAILED
  age<pukey>: FAILED
    - | failed to open file: open /Users/me/Library/Application
      | Support/sops/age/keys.txt: no such file or directory

$ echo $XDG_CONFIG_HOME
/Users/me/.config
```

Sadly, the [ship has sailed](https://github.com/golang/go/issues/29960#issuecomment-505321146) for `os.UserConfigDir()`, which is what sops uses, but that doesn't mean there can't be a fallback, so it would be nice if sops could support it.

felixfontein:
Fixes #1263.

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"filippo.io/age"
@@ -23,8 +24,10 @@ const (
 	// age keys file.
 	SopsAgeKeyFileEnv = "SOPS_AGE_KEY_FILE"
 	// SopsAgeKeyUserConfigPath is the default age keys file path in
-	// os.UserConfigDir.
+	// getUserConfigDir().
 	SopsAgeKeyUserConfigPath = "sops/age/keys.txt"
+	// On macOS, os.UserConfigDir() ignores XDG_CONFIG_HOME. So we handle that manually.
+	xdgConfigHome = "XDG_CONFIG_HOME"
 )
 
 // log is the global logger for any age MasterKey.
@@ -222,6 +225,15 @@ func (key *MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+func getUserConfigDir() (string, error) {
+	if runtime.GOOS == "darwin" {
+		if userConfigDir, ok := os.LookupEnv(xdgConfigHome); ok && userConfigDir != "" {
+			return userConfigDir, nil
+		}
+	}
+	return os.UserConfigDir()
+}
+
 // loadIdentities attempts to load the age identities based on runtime
 // environment configurations (e.g. SopsAgeKeyEnv, SopsAgeKeyFileEnv,
 // SopsAgeKeyUserConfigPath). It will load all found references, and expects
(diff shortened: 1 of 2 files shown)


hiddeco:
Think inclusion of iOS is weird here because of two reasons:

1. We do not actually release any SOPS binary for iOS.
2. If someone were to use SOPS as a SDK on iOS, I am doubtful about them being able to set the environment variable. But I may be wrong here, as I am not an iPad user, nor have I ever attempted to do anything shell-like on such a device.

felixfontein:
I fully agree, I mainly added `ios` here since it appeared in age/keysource_test.go. I'm not sure it does any good there as well, though. Maybe it's best to simply remove it completely. (We can still add it if we notice it's needed.)

age/keysource.go
```diff
@@ -222,6 +225,15 @@ func (key *MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+func getUserConfigDir() (string, error) {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "ios" {
```
felixfontein, line 229:
```suggestion
	if runtime.GOOS == "darwin" {
```

felixfontein:
(I'll squash once everyone is happy.)

age/keysource_test.go
```diff
@@ -380,8 +380,6 @@ func overwriteUserConfigDir(t *testing.T, path string) {
 	switch runtime.GOOS {
 	case "windows":
 		t.Setenv("AppData", path)
-	case "darwin", "ios": // This adds "/Library/Application Support" as a suffix to $HOME
```
hiddeco, line 383: Would maintain this, and add an individual test for `getUserConfigDir` to ensure the custom `darwin` behavior works as expected.

felixfontein: I added a test. I cannot run it on macOS though, only on Linux.

hiddeco: You added a test for `overwriteUserConfigDir`, and not for `getUserConfigDir` :smile:

felixfontein: Well, it's testing both. To test `getUserConfigDir()` we somehow need to set the correct environment variable. And there's already a helper for doing that, called `overwriteUserConfigDir`, so why not use that one instead of repeating the same code?

hiddeco: But we now do not appear to test the Darwin specific behavior around preferring the environment variable if set?

felixfontein: We do: on Darwin `overwriteUserConfigDir` now sets `XDG_CONFIG_HOME` (`default` case).

hiddeco: Ah _doh_, I am suffering from broken nights :-).

I think maintaining the logic in `overwriteUserConfigDir` and adding a specific test case for `XDG_CONFIG_HOME` on Darwin (testing `getUserConfigDir`) results in better maintainable tests and is easier to grasp (~for example, for people with limited sleep~), in which the test case is descriptive for the feature you are adding. Instead of it being indirectly tested via a certain construction.

felixfontein: I adjusted the test to be more explicit. I didn't restore the `overwriteUserConfigDir` behavior though since it would now be incorrect if the user has `XDG_CONFIG_HOME` set.

hiddeco (approved):

felixfontein:

PR #1298, opened 2023-09-19
pgp: improve handling of GnuPG home dir

### LINKED ISSUE #1294: sops 3.8.0 doesn't work with qubes split gpg
Cross posting from https://github.com/getsops/sops/pull/1054#issuecomment-1722630295

After upgrading to the new release, qubes split gpg (https://www.qubes-os.org/doc/split-gpg/ / https://github.com/QubesOS/qubes-app-linux-split-gpg) is no longer working.
I think this is due to the new release passing `--no-default-keyring` and `--homedir` which are both broken with qubes split gpg for different reasons:
  - `--no-default-keyring` is handled [here](https://github.com/QubesOS/qubes-app-linux-split-gpg/blob/ef64f62a882a2e62f31c06ab5e02cbe7e0dd1643/gpg-client-wrapper#L39-L40)
  - `--homedir` isn't handled at all, and it fails with `qubes-gpg-client: unrecognized option '--homedir'`

Also filed this issue with qubes: https://github.com/QubesOS/qubes-issues/issues/8523

hiddeco:
Fixes #1294

### DIFF
--- a/pgp/keysource.go
+++ b/pgp/keysource.go
@@ -318,7 +318,7 @@ func (key *MasterKey) encryptWithGnuPG(dataKey []byte) error {
 		fingerprint,
 		"--no-encrypt-to",
 	}
-	err, stdout, stderr := gpgExec(key.gnuPGHome(), args, bytes.NewReader(dataKey))
+	err, stdout, stderr := gpgExec(key.gnuPGHomeDir, args, bytes.NewReader(dataKey))
 	if err != nil {
 		return fmt.Errorf("failed to encrypt sops data key with pgp: %s", strings.TrimSpace(stderr.String()))
 	}
@@ -407,7 +407,7 @@ func (key *MasterKey) decryptWithGnuPG() ([]byte, error) {
 	args := []string{
 		"-d",
 	}
-	err, stdout, stderr := gpgExec(key.gnuPGHome(), args, strings.NewReader(key.EncryptedKey))
+	err, stdout, stderr := gpgExec(key.gnuPGHomeDir, args, strings.NewReader(key.EncryptedKey))
 	if err != nil {
 		return nil, fmt.Errorf("failed to decrypt sops data key with pgp: %s",
 			strings.TrimSpace(stderr.String()))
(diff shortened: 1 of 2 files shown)


pgp/keysource.go
```diff
@@ -587,7 +566,7 @@ func fingerprintIndex(ring openpgp.EntityList) map[string]openpgp.Entity {
 // When the command fails, an error is returned.
 func gpgExec(gnuPGHome string, args []string, stdin io.Reader) (err error, stdout bytes.Buffer, stderr bytes.Buffer) {
 	if gnuPGHome != "" {
-		args = append([]string{"--no-default-keyring", "--homedir", gnuPGHome}, args...)
+		args = append([]string{"--homedir", gnuPGHome}, args...)
```
devstein, line 569: I was confused by this argument name with the same name as new function `gnuPGHome`. Maybe rename to `gnuPGHomeDir`?

pgp/keysource.go
```diff
@@ -318,7 +318,7 @@ func (key *MasterKey) encryptWithGnuPG(dataKey []byte) error {
 		fingerprint,
 		"--no-encrypt-to",
 	}
-	err, stdout, stderr := gpgExec(key.gnuPGHome(), args, bytes.NewReader(dataKey))
+	err, stdout, stderr := gpgExec(key.gnuPGHomeDir, args, bytes.NewReader(dataKey))
```
devstein, line 321: Just confirming that this shouldn't be:  `gnuPGHome(key.gnuPGHomeDir)`

pgp/keysource.go
```diff
@@ -407,7 +407,7 @@ func (key *MasterKey) decryptWithGnuPG() ([]byte, error) {
 	args := []string{
 		"-d",
 	}
-	err, stdout, stderr := gpgExec(key.gnuPGHome(), args, strings.NewReader(key.EncryptedKey))
+	err, stdout, stderr := gpgExec(key.gnuPGHomeDir, args, strings.NewReader(key.EncryptedKey))
```
devstein, line 410: Same

pgp/keysource.go
```diff
@@ -609,6 +588,27 @@ func gpgBinary() string {
...
+			usr, err := user.Current()
+			if err != nil {
+				return filepath.Join(os.Getenv("HOME"), ".gnupg")
+			}
+			return filepath.Join(usr.HomeDir, ".gnupg")
+		}
+		return dir
+	}
+	return customPath
+}
```
devstein, line 610: Nit: Any reason not to use early returns to match the order preference? IMO it improves readability 
```go
func gnuPGHome(customPath string) string {
  if customPath != "" {
    return customPath
  }

  dir := os.Getenv("GNUPGHOME")
  if dir != "" {
    return dir
  }

  usr, err := user.Current()
  if err != nil {
    return filepath.Join(os.Getenv("HOME"), ".gnupg")
  }

  return filepath.Join(usr.HomeDir, ".gnupg")
}
```

devstein (commented):
I don't understand the impact of this, but few minor code comments and clarifications

pgp/keysource.go
hiddeco: It shouldn't, because that would re-introduce the behavior of always supplying a home directory (resulting in `--homedir` getting configured).

pgp/keysource.go
hiddeco: No particular reason. I am in a weird camp that actually finds the current thing more readable as-is, but that is very subjective :-)

pgp/keysource.go
hiddeco: Will actually just change it to `homeDir` to match the flag.

devstein (commented):
LGTM

devstein (approved):

PR #1304, opened 2023-09-26
decrypt: fix dropped error

alrs:
This picks up a dropped `err` variable in the `decrypt` package.

### DIFF
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -59,6 +59,9 @@ func DataWithFormat(data []byte, format Format) (cleartext []byte, err error) {
 		key,
 		tree.Metadata.LastModified.Format(time.RFC3339),
 	)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to decrypt original mac: %w", err)
+	}
 	if originalMac != mac {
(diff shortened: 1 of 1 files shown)


decrypt/decrypt.go
```diff
@@ -59,6 +59,9 @@ func DataWithFormat(data []byte, format Format) (cleartext []byte, err error) {
 		key,
 		tree.Metadata.LastModified.Format(time.RFC3339),
 	)
+	if err != nil {
```
felixfontein, line 62: In case `err != nil`, very likely the check `originalMac != mac` also fails, so *an* error would have already been returned - but the message would not be so helpful (probably something like `"...: expected mac xxx, got "`). So handling the error here is definitely a good thing!

felixfontein (approved):

hiddeco (approved):

hiddeco:
There actually is a list of (many) more swallowed errors when you run `make staticcheck`, which should all eventually be looked at some time (which is a tiresome job).

---

@alrs if you rebase this yourself, we can maintain your commit signature :-). Thanks for the contribution :bow:

PR #1308, opened 2023-09-29
ci: add FOSSA license scan workflow

hiddeco:


felixfontein:
@hiddeco what's the current state here?

felixfontein:
There's also #1458.

PR #1311, opened 2023-09-29
Handle errors

felixfontein:
Fixes some things reported by `make staticcheck`. This needs more testing and potentially tests, thus I marked it as a draft for now.

### DIFF
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -89,6 +89,9 @@ func DecryptTree(opts DecryptTreeOpts) (dataKey []byte, err error) {
 	}
 	fileMac, err := opts.Cipher.Decrypt(opts.Tree.Metadata.MessageAuthenticationCode, dataKey, opts.Tree.Metadata.LastModified.Format(time.RFC3339))
 	if !opts.IgnoreMac {
+		if err != nil {
+			return nil, NewExitError(fmt.Sprintf("Cannot decrypt MAC: %s", err), codes.MacMismatch)
+		}
 		if fileMac != computedMac {
 			// If the file has an empty MAC, display "no MAC" instead of not displaying anything
 			if fileMac == "" {
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -321,10 +321,10 @@ func FixAWSKMSEncryptionContextBug(opts GenericDecryptOpts, tree *sops.Tree) (*s
 	}
 
 	file, err := os.Create(opts.InputPath)
-	defer file.Close()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/sops/common/common.go
```diff
@@ -89,6 +89,9 @@ func DecryptTree(opts DecryptTreeOpts) (dataKey []byte, err error) {
 	}
 	fileMac, err := opts.Cipher.Decrypt(opts.Tree.Metadata.MessageAuthenticationCode, dataKey, opts.Tree.Metadata.LastModified.Format(time.RFC3339))
 	if !opts.IgnoreMac {
+		if err != nil {
+			return nil, NewExitError(fmt.Sprintf("Cannot decrypt MAC: %s", err), codes.MacMismatch)
```
felixfontein, line 93: If `err != nil`, then decryption failed for some reason we should tell the user, instead of just saying that there was a MAC mismatch, with the file having a MAC of `nil`.

stores/yaml/store.go
```diff
@@ -131,6 +131,10 @@ func (store Store) appendYamlNodeToTreeBranch(node *yaml.Node, branch sops.TreeB
 		return nil, fmt.Errorf("YAML documents that are values are not supported")
 	case yaml.AliasNode:
 		branch, err = store.appendYamlNodeToTreeBranch(node.Alias, branch, false)
+		if err != nil {
+			// This should never happen since node.Alias was already successfully decoded before
+			return nil, err
```
felixfontein, line 136: I'm pretty sure this should never happen, but it definitely won't hurt to still pass on the error if it actually does.

stores/yaml/store.go
```diff
@@ -204,9 +208,9 @@ func (store *Store) appendSequence(in []interface{}, sequence *yaml.Node) {
 	}
 	if len(comments) > 0 {
 		if beginning {
-			comments = store.addCommentsHead(sequence, comments)
+			store.addCommentsHead(sequence, comments)
```
felixfontein, line 211: The return value is always `nil`. It's mainly returned so you can write `comments = store.addCommentsHead(xx, comments)` and continue using `comments` later on. But we have no "later" here, so removing the assignment is best (and makes the linter happy).

stores/yaml/store_test.go
```diff
@@ -48,6 +48,59 @@ var BRANCHES = sops.TreeBranches{
...
 
+var ALIASES = []byte(`---
+key1: &foo
+  - foo
+key2: *foo
+key3: &bar
+  foo: bar
+  baz: bam
+key4: *bar
+`)
```
felixfontein, line 59: Having a test with anchors and aliases surely doesn't hurt.

felixfontein (commented):
I've now went through all ignored errors. This is ready for reviewing!

hiddeco (approved):
Thanks @felixfontein 🙏

PR #1317, opened 2023-10-08
Report key rotation errors

### LINKED ISSUE #855: sops truncates the file to zero after a failed key rotation
Sops version: 3.7.1 (Installed from deb package)

How to reproduce:
1. create example.yaml using one KMS and one PGP key
2. turn off your aws profile / credentials
3. try to rotate the file in place: `sops -r -i example.yaml`

What happens:
1. example.yaml file gets truncated to zero bytes!
2. sops prints no error messages
3. sops returns exit code zero as if everything was ok

What I would like to happen:
1. sops detects that not all master keys are available
2. sops prints error message: you cannot rotate because you are missing key X
3. sops exits with non-zero code.

felixfontein:
Fixes #855.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -904,6 +904,11 @@ func main() {
 				AddMasterKeys:    addMasterKeys,
 				RemoveMasterKeys: rmMasterKeys,
 			})
+			// While this check is also done below, the `err` in this scope shadows
+			// the `err` in the outer scope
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Thanks @felixfontein :bow:

felixfontein:
@hiddeco thanks for reviewing and merging!

PR #1318, opened 2023-10-08
cmd/sops/main.go: make sure to wrap raw errors with toExitError()

felixfontein:
While working on #1317 I noticed that there are quite a few `return err`s which do not use `toExitError()` like most of the other errors that are handled. Some of the errors in questions might already be `cli.ExitError`s, but some might not.


hiddeco (approved):
Thank you once again @felixfontein :1st_place_medal: :cake:

felixfontein:
@hiddeco thanks here as well! :)

PR #1272, opened 2023-09-06
Enrich AWS authentication documentation

nsantiago2719:
Update README.srt to mention about using environment variables of AWS credentials

PR #1320, opened 2023-10-09
Improve RST and MD files

felixfontein:
Subset of #1287 that doesn't add linting.

#1272 should be merged first and this PR then updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,7 +96,7 @@ separated, in the **SOPS_PGP_FP** env variable.
 
 Note: you can use both PGP and KMS simultaneously.
 
-Then simply call SOPS with a file path as argument. It will handle the
+Then simply call ``sops`` with a file path as argument. It will handle the
 encryption/decryption transparently and open the cleartext file in an editor
 
 .. code:: sh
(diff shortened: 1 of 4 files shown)


hiddeco:
@felixfontein any chance you would be up to addressing the mixed usage of Sops / SOPS / sops throughout the document in this PR? In my opinion, it should be either "SOPS" (the name) or `sops` (the binary) depending on the context.

felixfontein:
@hiddeco good idea, I added a commit which normalize the SOPS usages. I converted quite a few `sops` to SOPS as well. Please take a look whether you think that's OK.

README.rst
```diff
@@ -331,79 +333,79 @@ We assume you have an instance (or more) of Vault running and you have privilege
...
+          hc_vault_transit_uri: "$VAULT_ADDR/v1/sops/keys/thirdkey"
+    EOF
+
+    $ sops --verbose -e prod/raw.yaml > prod/encrypted.yaml
 
 Adding and removing keys
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-When creating new files, ``sops`` uses the PGP, KMS and GCP KMS defined in the
+When creating new files, SOPS uses the PGP, KMS and GCP KMS defined in the
```
hiddeco, line 392: As we later make mention of flags, maintaining the command reference is better here I think.

There are a couple of other places where this applies as well.

felixfontein: I hope I manged to fix the right ones you meant.

hiddeco (approved):

felixfontein:
@hiddeco thanks for reviewing and merging!

PR #1332, opened 2023-10-29
Allow to override fileName with different value

### LINKED ISSUE #594: creation_rules from can't be matched when encrypting from stdin
given a project with following structure:
```
.
├── .sops.yaml
└── secrets
    └── test.json
```

with  `.sops.yaml` content:
```ymal
creation_rules:
  - path_regex: secrets/.*
    gcp_kms: projects/YOUR_POJECT/locations/global/keyRings/YOUR_RING/cryptoKeys/YOUR_KEY
``` 

I am not able to encrypt data from stdin into the `secrets` directory, for example
```
cat secrets/test.json | sops --input-type json --output-type json --output secrets/test.encrypted.json -e /dev/stdin
```

gives me:
```
error loading config: no matching creation rules found
```

I am guessing that `sops` is trying to match `/dev/stdin` against the `creation_rules` from the config file (changing `path_regex: secrets/.*`  `to path_regex: .*` makes it work, but that is obviously not a solution).

Here's a sample repo which can be used to reproduce the error:
https://github.com/adrian-gierakowski/sops-encrypt-from-stdin

### Possible solutions
1. match against value of --output (converted to absolute path, or path relative to config file) if present
2. add a new flag which would take a path that could be used to match against creation rules

The latter would be more universal as it would allow use cases where --output is not specified

As a side note: I stumbled upon this issue while trying to write a script to re-encrypt all files in a directory after changing the `creation_rules`, and wanted to avoid temporarily storing unencrypted data to disk while doing it.

felixfontein:
This adds a `--filename-override` parameter which can be used when providing data to encrypt (or decrypt) from `/dev/stdin`.

For example: `echo 'foo: bar' | ~/go/bin/sops --encrypt --filename-override foobar.yaml /dev/stdin` will take the data to encrypt from the `echo` command's output, and act as if `foobar.yaml` has been supplied instead of `/dev/stdin`. This is relevant for finding creation rules in the SOPS configuration, and for determining the input and output type if these are not explicitly specified via `--input-type` and `--output-type`.

Fixes #594.


cmd/sops/main.go
```diff
@@ -733,12 +737,16 @@ func main() {
 				return common.NewExitError("Error: cannot operate on non-existent file", codes.NoFileSpecified)
 			}
 		}
+		fileNameOverride := c.String("filename-override")
+		if fileNameOverride == "" {
+			fileNameOverride = fileName
+		}
```
devstein, line 743: Curious why choose `fileNameOverride` vs modify `fileName`?

```go
		fileNameOverride := c.String("filename-override")
		if fileNameOverride != "" {
			fileName = fileNameOverride
		}
```

Related do we need to update this `fileName` reference on lines 976-982?

```go
		// We open the file *after* the operations on the tree have been
		// executed to avoid truncating it when there's errors
		if c.Bool("in-place") || isEditMode || c.String("set") != "" {
			file, err := os.Create(fileName)
			if err != nil {
				return common.NewExitError(fmt.Sprintf("Could not open in-place file for writing: %s", err), codes.CouldNotWriteOutputFile)
			}
```

devstein (commented):
Nice work. One clarifying comment. Should we add an example for documentation? Something like "Encrypting data from stdout and writing to a new file"

cmd/sops/main.go
felixfontein: > Curious why choose `fileNameOverride` vs modify `fileName`?

Because we still want to read from / write to `fileName`, to we need the original value for these operations.

> Related do we need to update this `fileName` reference on lines 976-982?

No, because `fileName` is the file we actually read from (and write to, if `--output` isn't used).

felixfontein:
> Nice work. One clarifying comment. Should we add an example for documentation? Something like "Encrypting data from stdout and writing to a new file"

Yeah, documentation is definitely needed. I'll work on that next.

README.rst
```diff
@@ -323,7 +323,39 @@ Now you can encrypt a file using::
...
 And decrypt it using::
 
-     $ sops --decrypt test.enc.yaml
+    $ sops --decrypt test.enc.yaml
+
+
+Encrypting and decrypting from other programs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using ``sops`` in scripts or from other programs, there are often situations where you do not want to write encrypted or decrypted data to disk. The best way to avoid this is to pass data to SOPS via stdin, and to let SOPS write data to stdout. By default, the encrypt and decrypt operations write data to stdout already. To pass data via stdin, you need to pass ``/dev/stdin`` as the input filename. Please note that this only works on Unix-like operating systems such as macOS and Linux. On Windows, you have to use named pipes.
```
felixfontein, line 332: I'm not sure about Windows here. Is it true that you need named pipes for this? Or is there another mechanism?

hiddeco: With https://github.com/getsops/sops/pull/1104, it would simply be a `-`. I would need to revise this to incorporate my own feedback, as it appears the contributor gave up.

hiddeco: I'll update this section once I have recreated the above.

README.rst
```diff
@@ -323,7 +323,39 @@ Now you can encrypt a file using::
 
 And decrypt it using::
 
-     $ sops --decrypt test.enc.yaml
+    $ sops --decrypt test.enc.yaml
+
+
+Encrypting and decrypting from other programs
```
hiddeco, line 329: In think we need to line wrap the documentation, as that seems to be the case for the majority of the sections in the README and eases reading in e.g. a terminal.

hiddeco (approved):
Besides my comment around linewraping, this looks good to me. Thanks @felixfontein :bow:

README.rst
felixfontein: I'd actually argue that explicitly wrapping lines is bad, since terminals by default wrap lines and thus you use the full terminal width, as opposed to explicit line wrapping, which looks mostly bad except if it is applied consistently *and* your terminal is wider than the limit used in the file. But that's something we should look at at another point in time :)

I've wrapped lines to a similar length as other parts of this file for now.

felixfontein:
@hiddeco thanks a lot for reviewing this!

Issue #1259, opened 2023-08-14
`--output-type` must appear before `-d`


spencerwilson:
On sops 3.7.3:

## Expected

Given an encrypted .env file,
```
% sops -d secrets.enc.env --output-type json
<command prints secrets in the env file, but in a JSON object container instead>
```

## Actual
```
% sops -d secrets.enc.env --output-type json
<command behaves as though `--output-type` was not given
```

## Workaround
Moving `--output-type json` in front of `-d <path>` _does_ work:
```
% sops --output-type json -d secrets.enc.env
<expected output as described above>
```

Not sure if this is expected behavior and thus a docs gap, or a bug. I'd guess the latter since `sops --help` says that everything is just a "global option" with no relative order required for them to work.

felixfontein:
I guess this is related to #598. There's a bunch of similar issues, most (or all?) of them closed linking to that one.

felixfontein:
Also please note that `-d path` are **two separate** options: a flag (`-d`) and a positional argument (`path`). All options must come before the positional arguments. So `sops -d --output-type json secrets.enc.env` will work as well.

felixfontein:
#1342 improves the situation by at least warning when this happens.

Issue #598, opened 2019-12-20
Argument order for --extract


heydonovan:
It took me a while to figure this out, but should argument order matter here? 🤔 

```
$ sops -d secrets.yaml
envSecrets:
    deploy-key: 12345

$ sops -d secrets.yaml --extract '["envSecrets"]["deploy-key"]'
envSecrets:
    deploy-key: 12345

$ sops -d --extract '["envSecrets"]["deploy-key"]' secrets.yaml
12345
```

autrilla:
Yeah, this is not ideal and it’s an artifact of the CLI library we use. The
file name must go last, essentially, since it’s an argument and not a flag.

This is also not specific to —extract.

patricknelson:
I think this should be detected and fail in case any parameters are passed after the file name.

The reason why: I was **also** stuck on this for a bit until I realized my parameter order was incorrect. I found this issue since I accidentally used the same order as @heydonovan `sops -d secrets.yaml --extract '["envSecrets"]["deploy-key"]'`. Since it didn't fail, it took too long assuming maybe my `--extract` syntax was invalid.  Only once you fix the order to `sops -d --extract '["envSecrets"]["deploy-key"]' secrets.yaml` will it begin to only only extract the specified key, but also notify that the key is valid or not.

Should we open a new issue? Is that a bug or considered expected behavior?

autrilla:
It's the same issue. It's considered a bug, and we definitely will welcome PRs that fix it, but it's a tricky one, since it'll affect lots of things and therefore will require some evidence that nothing was broken because of it.

felixfontein:
#1342 improves the situation by at least warning when this happens.

PR #1342, opened 2023-11-05
Warn/fail if the wrong number of arguments is provided

felixfontein:
When warning, also inform about potentially ignored flags.

This improves UX because right now, many users find out that some of their flags are "ignored" (because the CLI library treats them as positional arguments, and sops only uses the first one of them). See for example #1259 or #598.

Either #1274 or this PR needs to be adjusted, whatever is merged first, since this PR is adding a warning for `updatekeys`, while that PR allows `updatekeys` to process multiple filenames.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -48,6 +48,21 @@ func init() {
 	log = logging.NewLogger("CMD")
 }
 
+func warnMoreThanOnePositionalArgument(c *cli.Context) {
+	if c.NArg() > 1 {
+		log.Warn("More than one positional argument provided. Only the first one will be used!")
+		potentialFlag := ""
+		for i, value := range c.Args() {
+			if i > 0 && strings.HasPrefix(value, "-") {
+				potentialFlag = value
+			}
+		}
+		if potentialFlag != "" {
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -48,6 +48,21 @@ func init() {
...
+	if c.NArg() > 1 {
+		log.Warn("More than one positional argument provided. Only the first one will be used!")
+		potentialFlag := ""
+		for i, value := range c.Args() {
+			if i > 0 && strings.HasPrefix(value, "-") {
+				potentialFlag = value
+			}
+		}
+		if potentialFlag != "" {
+			log.Warn(fmt.Sprintf("Note that one of the ignored positional argument is %q, which looks like a flag. Flags must always be provied before the first positional argument!", potentialFlag))
```
hiddeco, line 61:
```suggestion
			log.Warn(fmt.Sprintf("Note that one of the ignored positional argument is %q, which looks like a flag. Flags must always be provided before the first positional argument!", potentialFlag))
```

hiddeco (approved):

felixfontein:

PR #1345, opened 2023-11-07
Sort masterkeys according to decryption-order

bkreitch:
This is more generic way to address https://github.com/getsops/sops/issues/305. As suggested in https://github.com/getsops/sops/issues/305#issuecomment-1793766331 and https://github.com/getsops/sops/issues/305#issuecomment-1793768768 it supports `--decryption-order` and **SOPS_DECRYPTION_ORDER** to specify the order in which decryption method are tried. Default order is `age,pgp`, so it covers functionality of #1121 too.

Couple of additional thoughts:
-  Sorting just indices as suggested in https://github.com/getsops/sops/pull/1121#discussion_r1382581918 together with priorities makes code a bit cryptic. Maybe using copy() to clone the group slice and sort the cloned slice could be acceptable alternative (since we don't modify the elements, but only the order)?
- Passing DecryptionOrder through number of options and function calls is a bit cumbersome. Would it make sense to include DecryptionOrder and KeyServices together in a struct since they seem to be used mostly together?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -170,6 +170,11 @@ Given that, the only command a SOPS user needs is:
 encrypted if modified, and saved back to its original location. All of these
 steps, apart from the actual editing, are transparent to the user.
 
+The order in which available decryption methods are tried can be specified with
+``--decryption-order`` option or **SOPS_DECRYPTION_ORDER** environment variable
+as a comma separated list. The default order is ``age,pgp``. Offline methods are
+tried first and then the remaining ones.
+
 Test with the dev PGP key
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -12,8 +12,9 @@ import (
 
 	"filippo.io/age"
 	"filippo.io/age/armor"
-	"github.com/getsops/sops/v3/logging"
 	"github.com/sirupsen/logrus"
+
+	"github.com/getsops/sops/v3/logging"
 )
 
 const (
@@ -28,6 +29,8 @@ const (
 	SopsAgeKeyUserConfigPath = "sops/age/keys.txt"
 	// On macOS, os.UserConfigDir() ignores XDG_CONFIG_HOME. So we handle that manually.
 	xdgConfigHome = "XDG_CONFIG_HOME"
+	// KeyTypeIdentifier is the string used to identify an age MasterKey.
+	KeyTypeIdentifier = "age"
 )
 
 // log is the global logger for any age MasterKey.
@@ -223,6 +226,11 @@ func (key *MasterKey) ToMap() map[string]interface{} {
 	out["recipient"] = key.Recipient
 	out["enc"] = key.EncryptedKey
 	return out
+}
+
+// TypeToIdentifier returns the string identifier for the MasterKey type.
+func (key *MasterKey) TypeToIdentifier() string {
+	return KeyTypeIdentifier
 }
 
 func getUserConfigDir() (string, error) {
--- a/azkv/keysource.go
+++ b/azkv/keysource.go
@@ -22,6 +22,11 @@ import (
 	"github.com/getsops/sops/v3/logging"
 )
 
+const (
+	// KeyTypeIdentifier is the string used to identify an Azure Key Vault MasterKey.
+	KeyTypeIdentifier = "azure_kv"
+)
+
 var (
 	// log is the global logger for any Azure Key Vault MasterKey.
 	log *logrus.Logger
@@ -213,6 +218,11 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	out["created_at"] = key.CreationDate.UTC().Format(time.RFC3339)
 	out["enc"] = key.EncryptedKey
 	return out
+}
+
+// TypeToIdentifier returns the string identifier for the MasterKey type.
+func (key *MasterKey) TypeToIdentifier() string {
+	return KeyTypeIdentifier
 }
 
 // getTokenCredential returns the tokenCredential of the MasterKey, or
--- a/cmd/sops/codes/codes.go
+++ b/cmd/sops/codes/codes.go
@@ -25,6 +25,7 @@ const (
 	NoFileSpecified                        int = 100
 	CouldNotRetrieveKey                    int = 128
 	NoEncryptionKeyFound                   int = 111
+	DuplicateDecryptionKeyType             int = 112
 	FileHasNotBeenModified                 int = 200
 	NoEditorFound                          int = 201
 	FailedToCompareVersions                int = 202
--- a/cmd/sops/common/common.go
+++ b/cmd/sops/common/common.go
@@ -72,6 +72,8 @@ type DecryptTreeOpts struct {
 	Tree *sops.Tree
 	// KeyServices are the key services to be used for decryption of the data key
 	KeyServices []keyservice.KeyServiceClient
+	// DecryptionOrder is the order in which available decryption methods are tried
+	DecryptionOrder []string
 	// IgnoreMac is whether or not to ignore the Message Authentication Code included in the SOPS tree
 	IgnoreMac bool
 	// Cipher is the cryptographic cipher to use to decrypt the values inside the tree
@@ -80,7 +82,7 @@ type DecryptTreeOpts struct {
 
 // DecryptTree decrypts the tree passed in through the DecryptTreeOpts and additionally returns the decrypted data key
 func DecryptTree(opts DecryptTreeOpts) (dataKey []byte, err error) {
-	dataKey, err = opts.Tree.Metadata.GetDataKeyWithKeyServices(opts.KeyServices)
+	dataKey, err = opts.Tree.Metadata.GetDataKeyWithKeyServices(opts.KeyServices, opts.DecryptionOrder)
 	if err != nil {
 		return nil, NewExitError(err, codes.CouldNotRetrieveKey)
 	}
@@ -222,11 +224,12 @@ func GetKMSKeyWithEncryptionCtx(tree *sops.Tree) (keyGroupIndex int, keyIndex in
 
 // GenericDecryptOpts represents decryption options and config
 type GenericDecryptOpts struct {
-	Cipher      sops.Cipher
-	InputStore  sops.Store
-	InputPath   string
-	IgnoreMAC   bool
-	KeyServices []keyservice.KeyServiceClient
+	Cipher          sops.Cipher
+	InputStore      sops.Store
+	InputPath       string
+	IgnoreMAC       bool
+	KeyServices     []keyservice.KeyServiceClient
+	DecryptionOrder []string
 }
 
 // LoadEncryptedFileWithBugFixes is a wrapper around LoadEncryptedFile which includes
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -1,4 +1,4 @@
-package main //import "github.com/getsops/sops/v3/cmd/sops"
+package main // import "github.com/getsops/sops/v3/cmd/sops"
 
 import (
 	"context"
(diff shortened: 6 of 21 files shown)


cmd/sops/codes/codes.go
```diff
@@ -25,6 +25,8 @@ const (
 	NoFileSpecified                        int = 100
 	CouldNotRetrieveKey                    int = 128
 	NoEncryptionKeyFound                   int = 111
+	InvalidDecryptionKeyType               int = 112
+	DupicateDecryptionKeyType              int = 113
```
GMartinez-Sisti, line 29: typo!
```suggestion
	DuplicateDecryptionKeyType              int = 113
```

cmd/sops/main.go
```diff
@@ -1251,3 +1281,21 @@ func extractSetArguments(set string) (path []interface{}, valueToInsert interfac
...
+		return sops.SopsDecryptionOrderDefault, nil
+	}
+	orderList := strings.Split(decryptionOrder, ",")
+	unique := make(map[string]bool)
+	for _, v := range orderList {
+		if !slices.Contains(sops.MasterKeyTypes, v) {
+			return nil, common.NewExitError(fmt.Sprintf("Invalid decryption key type: %s", v), codes.InvalidDecryptionKeyType)
+		}
+		if _, ok := unique[v]; ok {
+			return nil, common.NewExitError(fmt.Sprintf("Duplicate decryption key type: %s", v), codes.InvalidDecryptionKeyType)
```
GMartinez-Sisti, line 1296: Wrong error:
```suggestion
			return nil, common.NewExitError(fmt.Sprintf("Duplicate decryption key type: %s", v), codes.DuplicateDecryptionKeyType)
```

keyservice/keyservice.go
```diff
@@ -82,3 +82,22 @@ func KeyFromMasterKey(mk keys.MasterKey) Key {
...
+	case *kms.MasterKey:
+		return "kms"
+	case *azkv.MasterKey:
+		return "azure_kv"
+	case *age.MasterKey:
+		return "age"
+	default:
+		panic(fmt.Sprintf("Tried to convert unknown MasterKey type %T to string", mk))
+	}
+}
```
GMartinez-Sisti, line 103: I would suggest to add a new field or a method to the `MasterKey` struct instead to keep it closer to the source. When new key types are added this might be missed.

bkreitch: Good idea. Method seems to fit nicely.

sops_test.go
```diff
@@ -758,3 +761,10 @@ func TestEmitAsMap(t *testing.T) {
 		assert.Equal(t, expected, data)
 	}
 }
+
+func TestSortKeyGroupIndices(t *testing.T) {
+	group := KeyGroup{&hcvault.MasterKey{}, &age.MasterKey{}, &pgp.MasterKey{}}
+	expected := []int{1, 2, 0}
+	indices := sortKeyGroupIndices(group, SopsDecryptionOrderDefault)
+	assert.Equal(t, expected, indices)
```
felixfontein, line 769: I would do some more tests, like also having `pgp.MasterKey{}` before `age.MasterKey{}`.

felixfontein (commented):
Looks good on a very first glance!

sops_test.go
bkreitch: > I would do some more tests, like also having `pgp.MasterKey{}` before `age.MasterKey{}`.

@felixfontein sure, I've added more tests.

felixfontein:
The failing sanity test will get fixed by #1350.

It's probably best to remove the changes from go.mod and go.sum to avoid conflicts.

bkreitch:
> It's probably best to remove the changes from go.mod and go.sum to avoid conflicts.

@felixfontein Ok, sure, removed.

bkreitch:
> It's probably best to remove the changes from go.mod and go.sum to avoid conflicts.

@felixfontein Looks like tests won't pass without committed changes to go.mod and go.sum

cmd/sops/codes/codes.go
```diff
@@ -25,6 +25,8 @@ const (
 	NoFileSpecified                        int = 100
 	CouldNotRetrieveKey                    int = 128
 	NoEncryptionKeyFound                   int = 111
+	InvalidDecryptionKeyType               int = 112
+	DuplicateDecryptionKeyType              int = 113
```
felixfontein, line 29:
```suggestion
	DuplicateDecryptionKeyType             int = 113
```

felixfontein (commented):
I think this looks good! @getsops/maintainers what do you think?

felixfontein:
> > It's probably best to remove the changes from go.mod and go.sum to avoid conflicts.
> 
> @felixfontein Looks like tests won't pass without committed changes to go.mod and go.sum

I'm really confused why these are needed. :confused:

bkreitch:
> > > It's probably best to remove the changes from go.mod and go.sum to avoid conflicts.
> > 
> > 
> > @felixfontein Looks like tests won't pass without committed changes to go.mod and go.sum
> 
> I'm really confused why these are needed. 😕

@felixfontein For now I've just added a small function instead of importing "golang.org/x/exp/slices". Later with Go 1.21 it could be replaced with build-in "slices" package.

felixfontein (approved):

Ph0tonic (changes requested):
Juste a small misspell

README.rst
```diff
@@ -170,6 +170,10 @@ Given that, the only command a SOPS user needs is:
 encrypted if modified, and saved back to its original location. All of these
 steps, apart from the actual editing, are transparent to the user.
 
+Order in which availible decryption methods are tried can by specified by ``--decryption-order`` option
```
Ph0tonic, line 173:
```suggestion
Order in which available decryption methods are tried can by specified by ``--decryption-order`` option
```

README.rst
```diff
@@ -170,6 +170,10 @@ Given that, the only command a SOPS user needs is:
...
+or **SOPS_DECRYPTION_ORDER** environment variable as comma separated list.
```
Ph0tonic, line 174:
```suggestion
or **SOPS_DECRYPTION_ORDER** environment variable as a comma separated list.
```

README.rst
```diff
@@ -170,6 +170,10 @@ Given that, the only command a SOPS user needs is:
 encrypted if modified, and saved back to its original location. All of these
 steps, apart from the actual editing, are transparent to the user.
 
+The order in which available decryption methods are tried, can be specified with ``--decryption-order`` option
+or **SOPS_DECRYPTION_ORDER** environment variable as a comma separated list.
+Default order is ``age,pgp`` - meaning offline methods are tried first and then the remaining ones.
+
```
Ph0tonic, line 176: A last suggestion, sorry for not having done it in one shot.

```suggestion
The order in which available decryption methods are tried can be specified with ``--decryption-order`` option
or **SOPS_DECRYPTION_ORDER** environment variable as a comma separated list.
The default order is ``age,pgp``. Offline methods are tried first and then the remaining ones.

```

README.rst
```diff
@@ -170,6 +170,9 @@ Given that, the only command a SOPS user needs is:
 encrypted if modified, and saved back to its original location. All of these
 steps, apart from the actual editing, are transparent to the user.
 
+The order in which available decryption methods are tried can be specified with ``--decryption-order`` option
+or **SOPS_DECRYPTION_ORDER** environment variable as a comma separated list.
+The default order is ``age,pgp``. Offline methods are tried first and then the remaining ones.
```
hiddeco, line 175:
```suggestion
The order in which available decryption methods are tried can be specified with
``--decryption-order`` option or **SOPS_DECRYPTION_ORDER** environment variable
as a comma separated list. The default order is ``age,pgp``. Offline methods are
tried first and then the remaining ones.

```

age/keysource.go
```diff
@@ -225,6 +225,11 @@ func (key *MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+// TypeToString converts key type to a string
```
hiddeco, line 228:
```suggestion
// TypeToString converts key type to a string.
```

age/keysource.go
```diff
@@ -225,6 +225,11 @@ func (key *MasterKey) ToMap() map[string]interface{} {
...
+func (key *MasterKey) TypeToString() string {
+	return "age"
```
hiddeco, line 230: This could be exported as a `const`, making it programmatically easier to check against.

azkv/keysource.go
```diff
@@ -215,6 +215,11 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+// TypeToString converts key type to a string
```
hiddeco, line 218: Same comments as above.

cmd/sops/common/common.go
```diff
@@ -70,7 +70,8 @@ type DecryptTreeOpts struct {
 	// Tree is the tree to be decrypted
 	Tree *sops.Tree
 	// KeyServices are the key services to be used for decryption of the data key
-	KeyServices []keyservice.KeyServiceClient
+	KeyServices     []keyservice.KeyServiceClient
+	DecryptionOrder []string
```
hiddeco, line 74: This should have a doc block describing what the field is for.

gcpkms/keysource.go
```diff
@@ -196,6 +196,11 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+// TypeToString converts key type to a string
```
hiddeco, line 199: Same comments as above.

hcvault/keysource.go
```diff
@@ -216,6 +216,11 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+// TypeToString converts key type to a string
```
hiddeco, line 219: Same comments as above.

kms/keysource.go
```diff
@@ -297,6 +297,11 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+// TypeToString converts key type to a string
```
hiddeco, line 300: Same comments as above.

pgp/keysource.go
```diff
@@ -449,6 +449,11 @@ func (key MasterKey) ToMap() map[string]interface{} {
 	return out
 }
 
+// TypeToString converts key type to a string
```
hiddeco, line 452: Same comments as above.

sops.go
```diff
@@ -58,6 +59,11 @@ import (
 // DefaultUnencryptedSuffix is the default suffix a TreeItem key has to end with for sops to leave its Value unencrypted
 const DefaultUnencryptedSuffix = "_unencrypted"
 
+var (
+	MasterKeyTypes             = []string{"pgp", "gcp_kms", "hc_vault", "kms", "azure_kv", "age"}
+	SopsDecryptionOrderDefault = []string{"age", "pgp"}
+)
```
hiddeco, line 65: Am _really_ on the fence about this implementation detail. The problem I have with this introduction is that it worsens the scattering of string identifiers throughout the code base, and can easily be forgotten to be updated when a new key source is introduced.

The first solution to mitigate this that comes to mind would be to start making use of the identifier constants exposed in the key source implementation (as I suggested above), but this would cause the abstraction to further leak details as we would now have to import every key implementation within this module. Which isn't a great solution either.

Which makes me think we should either:

1. Factor this out (but I have no suggestion on how to do this yet...).
2. As a short-term solution, not maintain a list of _all_ key types, but rather only sort based on what's provided (which means we can either rely on our own defaults with a limited set of identifiers, or rely on the user their input).

bkreitch: @hiddeco I agree, this list is a bit problematic. I've updated the sort to rely only on provided list.

felixfontein:
I think everything's good now (as good as it can get without solving the big issues @hiddeco mentioned, but these are out of scope for this PR). Except that now there's a conflict since I merged another PR. Sorry for that :( Should be pretty easy to resolve though...

hiddeco (approved):
Thank you very much for your contribution @bkreitch. Made some small changes after looking at it for a 4th time, please let me know if you agree with them and I will merge :bow:

bkreitch:
> Thank you very much for your contribution @bkreitch. Made some small changes after looking at it for a 4th time, please let me know if you agree with them and I will merge 🙇

@hiddeco Yep, sure, thanks :)

felixfontein:
Thanks to everyone involved getting this ready and merged :)

PR #1349, opened 2023-11-13
Overwrite home dir in vault test

bkreitch:
This will overwrite home dir in additional test to ensure local .vault-token is not picked up.

### DIFF
--- a/hcvault/keysource_test.go
+++ b/hcvault/keysource_test.go
@@ -359,7 +359,14 @@ func Test_dataKeyFromSecret(t *testing.T) {
 
 func Test_vaultClient(t *testing.T) {
 	t.Run("client", func(t *testing.T) {
+		tmpDir := t.TempDir()
+
+		// Reset before and after to make sure the override is taken into
+		// account, and restored after the test.
+		homedir.Reset()
+		t.Cleanup(func() { homedir.Reset() })
 		t.Setenv("VAULT_TOKEN", "")
+		t.Setenv("HOME", tmpDir)
 
 		got, err := vaultClient(testVaultAddress, "")
 		assert.NoError(t, err)


hcvault/keysource_test.go
```diff
@@ -359,7 +359,14 @@ func Test_dataKeyFromSecret(t *testing.T) {
...
 func Test_vaultClient(t *testing.T) {
 	t.Run("client", func(t *testing.T) {
+		tmpDir := t.TempDir()
+
+		// Reset before and after to make sure the override is taken into
+		// account, and restored after the test.
+		homedir.Reset()
+		t.Cleanup(func() { homedir.Reset() })
 		t.Setenv("VAULT_TOKEN", "")
+		t.Setenv("HOME", tmpDir)
```
felixfontein, line 369: Maybe it's better to use `overwriteUserConfigDir()` from `age/keysource_test.go`.

bkreitch: @felixfontein I wanted to add `overwriteHomeDir()` (since it should change **HOME** variable and not **AppData/XDG_CONFIG_HOME**), but then I saw other tests in the file just use `t.Setenv("HOME", tmpDir)` - which seems to be good enough for Windows and Unix (including MacOS), but probably won't work on **plan9**. But I can add the function and replace all `t.Setenv("HOME", tmpDir)` if you think it will be better.

felixfontein: I guess we should adjust all tests at once then at some point in the future.

(And no, I'm not aware of anyone using `plan9` :) )

felixfontein (approved):

felixfontein:
@bkreitch thanks for this one as well!

PR #1350, opened 2023-11-13
README: Correctly tag code as yaml-stream

felixfontein:
Right now the rstcheck test fails because a YAML stream with two documents is labelled as a YAML file.


hiddeco (approved):
We should probably prioritize just making this Markdown to save ourselves from an immense amount of pain (except the automatic ToC generation, that's actually pretty nice).

felixfontein:

PR #1356, opened 2023-11-21
Consolidate Flatten/Unflatten pre/post processing

### LINKED ISSUE #1353: --mac-only-encrypted panics on .env files.
Thank you for --mac-only-encrypted  Works great on my .yaml files

However it panics on .env

```shell
$ sops -e --mac-only-encrypted sample.env
panic: interface conversion: interface {} is bool, not string

goroutine 1 [running]:
github.com/getsops/sops/v3/stores/dotenv.(*Store).EmitPlainFile(0xc00058e200?, {0xc000011110?, 0xc0005aae45?, 0xc0005aae45?})
        /home/lance/go/src/github.com/getsops/sops/stores/dotenv/store.go:130 +0x2bf
github.com/getsops/sops/v3/stores/dotenv.(*Store).EmitEncryptedFile(0xc0004e8680?, {{{0xdd1be8f, 0xedce9ddb7, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, ...}, ...})
        /home/lance/go/src/github.com/getsops/sops/stores/dotenv/store.go:115 +0x3fc
main.encrypt({{0x1231d98, 0xc0004e4690}, {0x12389b0, 0x19bc1a0}, {0x12389b0, 0x19bc1a0}, {0xc000049ad0, 0x24}, {0xc0004daac0, 0x1, ...}, ...})
        /home/lance/go/src/github.com/getsops/sops/cmd/sops/encrypt.go:107 +0x575
main.main.func8(0xc0001898c0)
        /home/lance/go/src/github.com/getsops/sops/cmd/sops/main.go:808 +0xab8
github.com/urfave/cli.HandleAction({0xeb5880?, 0x110f8b0?}, 0xc0004756c0?)
        /home/lance/go/pkg/mod/github.com/urfave/cli@v1.22.14/app.go:524 +0x50
github.com/urfave/cli.(*App).Run(0xc0004756c0, {0xc00003e0a0, 0x5, 0x5})
        /home/lance/go/pkg/mod/github.com/urfave/cli@v1.22.14/app.go:286 +0x766
main.main()
        /home/lance/go/src/github.com/getsops/sops/cmd/sops/main.go:1014 +0x3530
```

I'm running 0cc1d116c47f0311c66f4bb348cf4026c130d275  (main HEAD from Nov 14)

lancerushing:
Fixes #1353

Consolidate the common json marshalling and \n handing to stores/flatten.go 
Make existing functions unexported.
Add casts for MACOnlyEncrypted.

When I was investigating #1353 I noticed code duplication in `dotenv/store.go` and `ini/store.go`.  It seemed cleaner to consolidate the the code.


stores/flatten.go
```diff
@@ -43,10 +44,10 @@ func flattenValue(value interface{}) interface{} {
 	return output
 }
 
-// Flatten flattens a map with potentially nested maps into a flat
+// flattenMap flattens a map with potentially nested maps into a flat
 // map. Only string keys are allowed on both the top-level map and
 // child maps.
-func Flatten(in map[string]interface{}) map[string]interface{} {
+func flattenMap(in map[string]interface{}) map[string]interface{} {
```
felixfontein, line 50: IMO this funcion should not be renamed.

stores/flatten.go
```diff
@@ -60,6 +61,28 @@ func Flatten(in map[string]interface{}) map[string]interface{} {
 	return newMap
 }
 
+func Flatten(md Metadata) (map[string]interface{}, error) {
```
felixfontein, line 64: Instead, this function should be called `FlattenMetadata`.

stores/flatten.go
```diff
@@ -173,8 +196,8 @@ func unflatten(currentNode interface{}, currentToken, nextToken token, value int
 	return nil
 }
 
-// Unflatten unflattens a map flattened by Flatten
-func Unflatten(in map[string]interface{}) map[string]interface{} {
+// unflattenMap unflattens a map flattened by flattenMap
+func unflattenMap(in map[string]interface{}) map[string]interface{} {
```
felixfontein, line 200: Same here...

stores/flatten.go
```diff
@@ -185,3 +208,42 @@ func Unflatten(in map[string]interface{}) map[string]interface{} {
 	}
 	return newMap
 }
+
+// Unflatten unflattens a map flattened by Flatten into Metadata
+func Unflatten(in map[string]interface{}) (Metadata, error) {
```
felixfontein, line 213: ...and here (`UnflattenMetadata`).

stores/flatten.go
```diff
@@ -60,6 +61,28 @@ func Flatten(in map[string]interface{}) map[string]interface{} {
...
+		return nil, err
+	}
+
+	flat := flattenMap(mdMap)
+	macOnlyEncryptedToString(flat)
+	for k, v := range flat {
+		if s, ok := v.(string); ok {
+			flat[k] = strings.Replace(s, "\n", "\\n", -1)
+		}
+	}
```
felixfontein, line 81: IMO this function (and its unflattening correspond) should be split up into three different functions:
1. a function that converts a metadata object to a flattened map;
2. a function which converts known boolean (or other non-string fields) to strings;
3. a function which replaces `\n` by `\\n`.

Then the first function can be optionally combined - or not combined - with the other two in case of potential other (future) file formats that only need some of these transformations.

felixfontein (commented):
Thanks for your contribution! I like the general idea, though I would avoid breaking backwards compatibility (`Flatten`/`Unflatten` signatures) and would make it more modular. What do you (and also others) think?

hiddeco (commented):
I am in favor of this solution, but would prefer it not breaking signatures of existing functions.

lancerushing:
@felixfontein I'll make the suggested changes. Should take me about an hour.

@hiddeco I'll rename the functions per felix's suggestions.  Thus the signatures won't be changing for the same function name.  But the existing functions will be removed b/c they have no callers.  (dead code)

stores/flatten.go
lancerushing: method names... :thinking: 
1. FlattenMetadata / UnflattenMetadata
2. encodeNonStrings / decodeNonStrings  :thinking:  packNonStrings / unpackNonStrings
3. encodeNewLines / decodeNewLines  .. pack__ / unpack__  

Any opinions?  I'll start with encode__/decode__ for now.  Easy to refactor if we want something different.

stores/flatten.go
lancerushing: Do we still want to leave the functions exported (capital function name)? There are no callers outside of the package.

stores/flatten.go
lancerushing: done

stores/flatten.go
lancerushing: done

stores/flatten.go
lancerushing: done

stores/flatten.go
lancerushing: done

lancerushing:
:thinking: Thinking more about the Metadata encoding for the `ini` and `dotenv` formats.  It's an interesting goal to encode the Metadata into the target format.  There are limitations.  Notice how the `dotenv` store checking for `isComplexValue`.  

:thinking: :thinking:  I wonder if it might be advantageous to encode it as one line of json?  or even json base64 encoded?  It would eliminate the `\n` and non-string handling.  But it would make the resulting file less "readable".  meh.

I'm not advocating the json (w/ or w/o base64).  Just thinking of alternatives that would be more robust.  Just something to think about for the next major version.

Anyway, thank you for a great tool.  I'm making a lot of use it in container images.

lancerushing:
@felixfontein / @hiddeco Do I need to keep this branch up to date with main?  (I did not see any recommendations in CONTRIBUTING.md)

felixfontein:
@lancerushing I think that's only necessary in case of conflicts; we can simply rebase before merging with the UI (at least as long as there are no conflicts and rebase just runs through).

felixfontein:
> 🤔 Thinking more about the Metadata encoding for the `ini` and `dotenv` formats. It's an interesting goal to encode the Metadata into the target format. There are limitations. Notice how the `dotenv` store checking for `isComplexValue`.
> 
> 🤔 🤔 I wonder if it might be advantageous to encode it as one line of json? or even json base64 encoded? It would eliminate the `\n` and non-string handling. But it would make the resulting file less "readable". meh.

One addition to "less readable" is that diffs are no longer useful. Right now if some metadata changes (for example you add a new key, or remove one), you can easily see what happens from the diff. That wouldn't be possible anymore.

Also changing the way how metadata is stored would break backwards compatibility, so for some time we would have to support both the old and new way how metadata is stored.

Having generic code for such stores that allow to flatten/unflatten metadata and do other operations (like handling formats where every value is a string) is probably the best way forward.

lancerushing:
Thanks for the all the discussion.  Do we have an ETA when we this could be merged?  Thank you again.

stores/flatten.go
```diff
@@ -60,6 +61,23 @@ func Flatten(in map[string]interface{}) map[string]interface{} {
...
+	err = json.Unmarshal(jsonBytes, &mdMap)
+	if err != nil {
+		return nil, err
+	}
+
+	flat := Flatten(mdMap)
+	encodeNonStrings(flat)
+	encodeNewLines(flat)
+	return flat, nil
```
felixfontein, line 78: IMO the calls to `encodeNonStrings` and `encodeNewLines` should not be part of this function. They should be called from the stores directly. (Same for the reverse function.)

I can imagine that in the future, some stores might need some of these transformations, but not all of them. Having to create new wrappers that are mostly copy'n'paste of this function (and the reverse function) sounds unnecessarly complex to me.

lancerushing: Done.  I moved the encode/decode NonStrings/NewLines calls out of this function

stores/flatten.go
```diff
@@ -185,3 +201,53 @@ func Unflatten(in map[string]interface{}) map[string]interface{} {
...
+
+func EncodeNewLines(m map[string]interface{}) {
+	for k, v := range m {
+		if s, ok := v.(string); ok {
+			m[k] = strings.Replace(s, "\n", "\\n", -1)
+		}
+	}
+}
+
+// DecodeNonStrings will look for known keys that are not strings and decode to the appropriate type
```
felixfontein, line 233:
```suggestion
// DecodeNonStrings will look for known metadata keys that are not strings and decode to the appropriate type
```

stores/flatten.go
```diff
@@ -185,3 +201,53 @@ func Unflatten(in map[string]interface{}) map[string]interface{} {
...
+func DecodeNonStrings(m map[string]interface{}) {
+	if v, ok := m["mac_only_encrypted"]; ok {
+		m["mac_only_encrypted"] = false
+		if v == "true" {
+			m["mac_only_encrypted"] = true
+		}
+	}
+}
+
+// EncodeNonStrings will look for known keys that are not strings and will encode it to strings
```
felixfontein, line 243:
```suggestion
// EncodeNonStrings will look for known metadata keys that are not strings and will encode it to strings
```

felixfontein (commented):
I would mention 'metadata' in these comments. Besides that, LGTM!

stores/flatten.go
lancerushing: Done.

stores/flatten.go
lancerushing: Done.

felixfontein (approved):

hiddeco (commented):
Besides the lack of doc blocks, I am happy with this implementation.

lancerushing:
Anything I need to do to get the PR merged?

hiddeco:
@lancerushing if you could add doc blocks to the functions which do not have one, then this should be good to go.

lancerushing:
> @lancerushing if you could add doc blocks to the functions which do not have one, then this should be good to go.

Done.  Added comments for the new functions that were missing comments.  -Thank you, Lance

felixfontein:
@lancerushing thank you very much for fixing this!

lancerushing:
:tada: Thank you for the merge and a great project!  Let me know if I can help with anything else.

PR #1358, opened 2023-11-25
Added support for access tokens in gcpkms

christoffer-eide:
This PR adds support for access token (via the `GOOGLE_CREDENTIALS` env var).

If the env var `GOOGLE_CREDENTIALS` is _not_ set, the gcloud sdk fetches an access token from the [instance metadata endpoint](https://cloud.google.com/appengine/docs/legacy/standard/java/accessing-instance-metadata#identifying_which_metadata_endpoint_to_use) `http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token`. 

In our workload on GKE, we don't want to use the access token returned by the metadata endpoint directly. We use this access token to [impersonate another service account](https://cloud.google.com/iam/docs/create-short-lived-credentials-direct#sa-credentials-oauth), which has the minimum of permissions required for the kms decrypt. 

There is no facility to set access tokens in sops, only static (long lived) credentials via the credentials file/`GOOGLE_CREDENTIALS`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -235,6 +235,44 @@ you can enable application default credentials using the sdk:
 
     $ gcloud auth application-default login
 
+It's also possible to configure the credentials using the env var ``GOOGLE_CREDENTIALS``.
+This can either point to a file or JSON content.
+
+.. code:: sh
+
+    export GOOGLE_CREDENTIALS=$(cat credentials.json)
+    export GOOGLE_CREDENTIALS=credentials.json
+
+Supported formats are:
+
+Credentials file:
+
+.. code:: json
+
+    {
+        "type": "service_account",
+        "project_id": "my-project",
+        "private_key_id": "66a4119f8aefbe8687ef0e14c6e7e0e1844b7950",
+        "private_key": "***",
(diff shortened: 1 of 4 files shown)


devstein (approved):
Thanks for contributing @christoffer-eide 🥇 

Looks good to me. Can you quickly fix the merge conflicts?

cc @getsops/maintainers if anyone else wants to take a look

devstein (commented):
(Meant to comment, not approve yet)

devstein (commented):
Can you add a note to the [GCP KMS docs](https://github.com/getsops/sops?tab=readme-ov-file#encrypting-using-gcp-kms) for others to discover this feature?

devstein:
Hey @christoffer-eide 👋  - someone else just opened a PR https://github.com/getsops/sops/pull/1578 for the same functionality. 

Are you able to address the comments, so we can get this merged?

devstein:
All good @christoffer-eide! I agree that using a separate env var  `CLOUDSDK_AUTH_ACCESS_TOKEN`, which is the same behavior as the GCP libraries, makes more sense.

Appreciate the contribution!


Closing in favor of #1578

PR #1372, opened 2023-12-10
Docs: indent is used both for encryption and decryption

felixfontein:
Right now the documenation of the `--indent` option says it's only used for encryption, while it is also used for decryption.

While testing this I noticed that for indenting YAML files, only indents of 2 to 9 are used. If you specify 1, or 10 or larger, the indent is ignored. This comes from go-yaml: https://github.com/go-yaml/yaml/blob/f6f7691b1fdeb513f56608cd2c32c51f8194bf51/emitterc.go#L328-L330

CC @Ph0tonic

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1192,6 +1192,11 @@ by configuring ``.sops.yaml`` with:
       yaml:
           indent: 2
 
+.. note::
+
+  The YAML emitter used by sops only supports values between 2 and 9. If you specify 1,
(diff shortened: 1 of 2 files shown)


felixfontein:
(When I tested it I started with `--indent 10` to have a ridiculous value that's easy to spot. I was surprised that it resulted in an indent of 2, and then started probing around...)

hiddeco (approved):

PR #1343, opened 2023-11-05
Add separate subcommands for encryption and decryption

felixfontein:
This is a Work in Progress for exploring how to implement #1333. Right now the new subcommands are basically a stripped-down copy'n'paste of the current "main" command.

Fixes #1333

Issue #1333, opened 2023-10-29
Make encrypt, decrypt, rotate, set, and edit proper commands


felixfontein:
Right now there are a few explicit commands (`exec-env`, `exec-file`, `publish`, `keyservice`, `groups`, `updatekeys`, `help`), and several implicit ones (`--encrypt`, `--decrypt`, `--rotate`, `--set`, edit), some pseudo-commands which only work with other commands (`--extract` with `--decrypt`, `--rm-*` and `--add-*` with `--rotate`).

This is pretty confusing, both from the `sops --help` output and also from the code (it is not trivial to figure out which of the flags are commands), *and* we have a long list of top-level flags which are only used in *some* of the commands, or even just one (like `--extract`, `--rm-*`, `--add-*`).

I think it would be better long term to migrate all "flag commands" (`--encrypt`, `--decrypt`, `--rotate`, `--set`, and the fallback 'edit') to proper subcommands. I would do this in three steps:
1. Add proper commands (additionally to the existing flag commands).
2. Eventually deprecate the use of global flags outside of the (flag) commands where they are really needed.
3. Eventually deprecate the flag commands themselves.
4. Eventually remove the flag commands, and remove global flags for command-specific options.

1\. should be done soon, 2. better sooner than later (this also makes the situation less confusing by telling users that the option they provided isn't actually used by the command they try to use), and 3. should only happen in a longer time (some years from now), and 4. even further in the future (say, at least 4-5 years?).

What do you think?

PR #1387, opened 2023-12-26
Move extraction of encryption and rotation options to separate functions

felixfontein:
Preparation for a rewrite of #1343, which aims at implementing #1333 (Make encrypt, decrypt, rotate, set, and edit proper commands).

Currently contains a rebased version of #974, since that adds more encryption options.

This simplifies the main command handler to a more manageable ~180 lines of code, and heavily reduces the copy'n'paste that would happen in an implementation like #1343.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1544,9 +1544,18 @@ that match the supplied regular expression. For example, this command:
 will not encrypt the values under the ``description`` and ``metadata`` keys in a YAML file
 containing kubernetes secrets, while encrypting everything else.
 
+For YAML files, another method is to use ``--encrypted-comment-regex`` which will
+only encrypt comments and values which have a preceding comment matching the supplied
+regular expression.
+
+Conversely, you can opt in to only left certain keys without encrypting by using the
+``--unencrypted-comment-regex`` option, which will leave the values and comments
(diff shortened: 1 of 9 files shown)


felixfontein:
An unintended side-effect: if before, `--encrypt --set` has been specified together, the `--encrypt` fails (since the file is already encrypted), but the error returned by it is *ignored* by the old code and overwritten by `--set`'s error (which is `nil`). 4e8796e24007b4c3b92d654a8b68d8d6e94a0422 restores the old behavior. (I'll create a PR to issue a warning if such commands are combined.)

felixfontein:
I modified that commit to 05ff0873e721610a2e99f4ed886e6654de59c33b: first, `--decrypt` and `--rotate` have the same effect as `--set` on `--encrypt` (see #1388 which adds a warning of more than one command is used). Also the change for `--rotate` was already introduced in #1317 (and is already released), so I don't want to undo that behavior change. (It was fixing something even more broken.)

felixfontein:
#1389 is this PR without #974, so it can be reviewed and merged sooner.

felixfontein:
I moved the main commit to #1392, rebased upon #1389. I'm closing this one in favor of #1392 (and of course #1389).

PR #1390, opened 2023-12-26
Improve integration tests: reformat test code, make sure exit codes are always checked

felixfontein:


devstein (approved):
Makes sense to me. What you think about adding a CI check to enforce `rustfmt`? Happy to make an issue for tracking

felixfontein:
Adding a CI check definitely makes sense! It's so easy to apply formatting (simply run `cargo fmt`) if you remember you have to do it...

PR #1391, opened 2023-12-26
Add separate subcommands for encryption, decryption, rotating, editing, and setting values

### LINKED ISSUE #1333: Make encrypt, decrypt, rotate, set, and edit proper commands

felixfontein:
This is another Work in Progress for exploring how to implement #1333, similar to the original WIP #1343. Right now the new subcommands are basically a stripped-down copy'n'paste of the current "main" command. This PR based upon #1389, which reduces the copy'n'paste a lot.

Fixes #1333

~~This PR includes the PRs #1389 and #1390. I will rebase once these are merged.~~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,12 +96,12 @@ separated, in the **SOPS_PGP_FP** env variable.
 
 Note: you can use both PGP and KMS simultaneously.
 
-Then simply call ``sops`` with a file path as argument. It will handle the
+Then simply call ``sops edit`` with a file path as argument. It will handle the
 encryption/decryption transparently and open the cleartext file in an editor
 
 .. code:: sh
 
-    $ sops mynewtestfile.yaml
+    $ sops edit mynewtestfile.yaml
     mynewtestfile.yaml doesn't exist, creating it.
     please wait while an encryption key is being generated and stored in a secure fashion
     file written to mynewtestfile.yaml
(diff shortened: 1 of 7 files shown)


felixfontein:
Now that #1389 has been merged, this is ready for review! :tada:

hiddeco (approved):
Nice work @felixfontein! :bow:

felixfontein:
@hiddeco thanks a lot for reviewing this!

PR #1392, opened 2023-12-27
Support using comments to select parts to encrypt

felixfontein:
This PR adds support to annotate comments with a string (e.g., `sops:enc`) which can then be matched with a regex. If it matches, the corresponding value (the one which follows the comment) is encrypted while other values are not. (There is also the opposite regex available, to select those values which should not be encrypted.)

This enables the YAML file to have the same structure encrypted and decrypted, without having to add suffixes or manage complex regexes to match keys. See #543 for more discussion.

~~(This PR currently contains #1389 and will be rebased once that's merged. It is a lot more likely that #1389 will be merged before this PR or any of its variants, so I decided to use that one as a basis.)~~

This PR continues the work of #974 by rebasing #974 upon #1389 and adding some final touches (see https://github.com/getsops/sops/pull/974#issuecomment-1862766339).

Closes #974.

### DIFF
--- a/sops.go
+++ b/sops.go
@@ -285,24 +285,24 @@ func (branch TreeBranch) Truncate(path []interface{}) (interface{}, error) {
 	return current, nil
 }
 
-func (branch TreeBranch) walkValue(in interface{}, path []string, onLeaves func(in interface{}, path []string) (interface{}, error)) (interface{}, error) {
+func (branch TreeBranch) walkValue(in interface{}, path []string, commentsStack [][]string, onLeaves func(in interface{}, path []string, commentsStack [][]string) (interface{}, error)) (interface{}, error) {
 	switch in := in.(type) {
 	case string:
-		return onLeaves(in, path)
+		return onLeaves(in, path, commentsStack)
 	case []byte:
-		return onLeaves(string(in), path)
+		return onLeaves(string(in), path, commentsStack)
 	case int:
-		return onLeaves(in, path)
+		return onLeaves(in, path, commentsStack)
 	case bool:
-		return onLeaves(in, path)
+		return onLeaves(in, path, commentsStack)
 	case float64:
-		return onLeaves(in, path)
+		return onLeaves(in, path, commentsStack)
 	case Comment:
-		return onLeaves(in, path)
+		return onLeaves(in, path, commentsStack)
 	case TreeBranch:
-		return branch.walkBranch(in, path, onLeaves)
+		return branch.walkBranch(in, path, commentsStack, onLeaves)
 	case []interface{}:
-		return branch.walkSlice(in, path, onLeaves)
+		return branch.walkSlice(in, path, commentsStack, onLeaves)
 	case nil:
 		// the value returned remains the same since it doesn't make
 		// sense to encrypt or decrypt a nil value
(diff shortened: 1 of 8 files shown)


felixfontein:
@mitar I think the two comments I added address the issues from my review in https://github.com/getsops/sops/pull/974#pullrequestreview-1642951900. Can you take a look at the commits?

sops.go
```diff
@@ -441,6 +441,14 @@ func (tree Tree) Encrypt(key []byte, cipher Cipher) (string, error) {
 				if err != nil {
 					return nil, fmt.Errorf("Could not encrypt value: %s", err)
 				}
+				if ok && tree.Metadata.UnencryptedCommentRegex != "" {
+					// If an encrypted comment matches tree.Metadata.UnencryptedCommentRegex, decryption will fail
+					// as the MAC does not match, and the commented value will not be decrypted.
+					matched, _ := regexp.Match(tree.Metadata.UnencryptedCommentRegex, []byte(in.(string)))
```
mitar, line 447: Why not use here `MatchString`?

Didn't we [say we will also try to match against `EncryptedCommentRegex`](https://github.com/getsops/sops/pull/974#issuecomment-1793750862)? So both of them should not match the resulting encrypted string?

mitar:
Thanks. I think this looks good as it is already. Thank you for all the work. I made a small comment, but I think it is not critical to do it.

sops.go
felixfontein: I originally implemented this, and then when trying to test it I removed the code again. As opposed to `UnencryptedCommentRegex`, encrypted comments matching `EncryptedCommentRegex` aren't a problem since comments are only encrypted/decrypted if they are part of an encrypted subtree that's encrypted due to a comment higher up.

For example, you can encrypt
```yaml
---
#ENC[AES256_GCM,data:MI3A,iv:c9lHpZCbz3W9ooGuGbmQTcWcXAH7JEL/5JyJdC41dec=,tag:LrQLBEgQayguId5RrNiOCA==,type:comment]
foo:
  bar: barbar #ENC[AES256_GCM,data:MI3A,iv:c9lHpZCbz3W9ooGuGbmQTcWcXAH7JEL/5JyJdC41dec=,tag:LrQLBEgQayguId5RrNiOCA==,type:comment]
# something else
baz:
  bam: bambam #ENC[AES256_GCM,data:MI3A,iv:c9lHpZCbz3W9ooGuGbmQTcWcXAH7JEL/5JyJdC41dec=,tag:LrQLBEgQayguId5RrNiOCA==,type:comment]
```
with `--encrypted-comment-regex ENC` and then decrypt it again without any problems.

We could still implement the check to be extra cautious, but it isn't needed (or I'm missing something :) ).

mitar: OK. I think it is then fine and we can revisit this if anyone finds a realistic case where it is breaking some user flow.

felixfontein:
Now that #1389 has been merged, this is ready for review! :tada:

sops.go
```diff
@@ -255,21 +255,38 @@ func (branch TreeBranch) walkValue(in interface{}, path []string, onLeaves func(
...
+func (branch TreeBranch) walkSlice(in []interface{}, path []string, commentsStack [][]string, onLeaves func(in interface{}, path []string, commentsStack [][]string) (interface{}, error)) ([]interface{}, error) {
+	// Because append returns a new slice, the original stack is not changed.
+	commentsStack = append(commentsStack, []string{})
 	for i, v := range in {
-		newV, err := branch.walkValue(v, path, onLeaves)
+		c, vIsComment := v.(Comment)
+		if vIsComment {
+			// If v is a comment, we add it to the slice of active comments.
+			// This allows us to also encrypt comments themselves by enabling encryption in a prior comment.
+			commentsStack[len(commentsStack)-1] = append(commentsStack[len(commentsStack)-1], c.Value)
```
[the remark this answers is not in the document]
felixfontein: That can't happen since in the first line of the function, one element is appended to `commentsStack`, and during the execution of this function no value is removed from `commentsStack`.

Gui13:
This looks good, any idea when this could be merged ?

felixfontein:
It's currently waiting for further reviews / approval by maintainer(s).

mitar:
Is there anything to do to push this further?

felixfontein:
Like several other PRs, this is waiting for a review / approval from someone else in @getsops/maintainers.

devstein (commented):
Looks good to me, but would be nice to get another review from @hiddeco @onedr0p if possible; otherwise, I can merge

sabre1041 (approved):
LGTM

felixfontein:
@mitar thank you very much for your contribution! I'm glad it finally got merged...
Also thanks to everyone who helped reviewing and testing this!

mitar:
Thank you everyone who was involved into getting this in, and especially you @felixfontein for all the help getting it in.

How could I bring #1404 further?

c33s:
@felixfontein 

thank you very much for this awesome feature!

two questions: 
- can you please add a simple example to the readme on how to use this new feature?
- is it possible to invert the features behavior (encrypt by default and use a comment to keep the value unencrypted)?

mitar:
README does have some text about this [here](https://github.com/getsops/sops/pull/1392/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168f).

> is it possible to invert the features behavior (encrypt by default and use a comment to keep the value unencrypted)?

Yes, you can use `unencrypted-regex` for that.

c33s:
i meant some examples for the config file like:
.sops.yaml
```
- encrypted_regex: '.'
  unencrypted_comment_regex: 'keep_decrypted'
  path_regex: '*\.yaml$'
```

file.yaml
```
foo:
  # keep_decrypted
  bar: 1234
  password: will be enrcypted
```

oh in missed the `--unencrypted-comment-regex`. really awesome it can work both ways

PR #1393, opened 2023-12-28
Make check whether file contains invalid keys for encryption dependent on output store

### LINKED ISSUE #1384: Encrypt dotenv file inplace more than once should be failsafe.
Oh no It hit me again. After changing the configuration in *.sops.yaml* I usually run `sops -i -d; sops -i -e` on the encrypted files to apply the changes. (Maybe there is a better way to do this ?)

But if I run `sops -i -e` twice on a file, then the file can not be encrypted afterwards and sops returns the error  **MAC mismatch. File has ...**. Any succeeding `sops -i -e` call on a file should abort if the file is already encrypted.

felixfontein:
Depending on the output store, the keys used for storing SOPS metadata vary. For YAML, JSON, and INI files, the `sops` top-level key is used. For DotEnv files, keys starting with `sops_` are used.

Fixes #1384.

### DIFF
--- a/stores/stores.go
+++ b/stores/stores.go
@@ -506,3 +506,13 @@ var ExampleFlatTree = sops.Tree{
 		},
 	},
 }
+
+// HasSopsTopLevelKey returns true if the given branch has a top-level key called "sops".
+func HasSopsTopLevelKey(branch sops.TreeBranch) bool {
+	for _, b := range branch {
+		if b.Key == "sops" {
+			return true
+		}
+	}
+	return false
+}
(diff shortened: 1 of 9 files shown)


stores/stores.go
```diff
@@ -506,3 +506,13 @@ var ExampleFlatTree = sops.Tree{
 		},
 	},
 }
+
+// HasSopsTopLevelKey returns true if the given branch has a top-level key called "sops".
+func HasSopsTopLevelKey(branch sops.TreeBranch) bool {
+	for _, b := range branch {
+		if b.Key == "sops" {
```
devstein, line 513: Not worth it for this PR, but thoughts on adding a constant for `"sops"`? 

https://github.com/search?q=repo:getsops/sops+sops%22+language:Go&type=code

devstein (approved):
👍

stores/stores.go
felixfontein: Yeah, that would be a very good idea!

felixfontein:
@devstein thanks for reviewing this!

stores/stores.go
felixfontein: #1398

PR #1396, opened 2023-12-29
Add basic tests for exec-env and exec-file

felixfontein:
~~Contains a rebased version of #1390.~~

### DIFF
--- a/functional-tests/src/lib.rs
+++ b/functional-tests/src/lib.rs
@@ -840,4 +840,113 @@ b: ba"#
             .expect("couldn't read output file contents");
         assert_ne!(contents, "", "Output file is empty");
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


felixfontein:
@getsops/maintainers I'm really confused. If I run the tests locally, even with exactly the same sops binary that is used in CI, the tests pass. But in CI they fail. I can really see no reason why they should fail. (Failing means that #784 is already fixed.) Does anyone of you have an idea what's going wrong here?

felixfontein:
Looks like GH's Ubuntu image has a funny `/bin/sh`. With `/bin/bash` the behavior is a lot more uniform. Anyway, the test now passes :tada:

onedr0p (approved):

felixfontein:

PR #1400, opened 2023-12-30
support for age identity with passphrase

tomaszduda23:
Solves https://github.com/getsops/sops/issues/933. It is basically rip of from https://github.com/FiloSottile/age/blob/101cc8676386b0503571a929a88618cae2f0b1cd/cmd/age/parse.go#L143. It looks that sops already support reading passwords similar way for [PGP](https://github.com/getsops/sops/blob/0bceaf42b834f254cf2c6a6f61e7121de8eb9c52/pgp/keysource.go#L520)

It also support password prompt by gpg-agent. It requires merge of https://github.com/getsops/gopgagent/pull/1


age/keysource.go
```diff
@@ -9,12 +10,15 @@ import (
...
 	"strings"
+	"syscall"
 
 	"filippo.io/age"
 	"filippo.io/age/armor"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/term"
 
 	"github.com/getsops/sops/v3/logging"
+	gpgagent "github.com/tomaszduda23/gopgagent"
```
felixfontein, line 21: I guess this should be changed to https://github.com/getsops/gopgagent/.

felixfontein (commented):
Converting to draft until the agent from getsops is used.

tomaszduda23:
@felixfontein would you mind review again?

age/keysource.go
```diff
@@ -284,11 +287,105 @@ func (key *MasterKey) loadIdentities() (ParsedIdentities, error) {
...
+								err = nil
+							}
+						}
+						if err != nil {
+							return "", fmt.Errorf("could not read passphrase: %v", err)
+						}
+
+						fmt.Fprintln(os.Stderr)
+
+						return pass, nil
```
felixfontein, line 352: #1692 adds `tui.go` in `age/` with some code vendored from age, namely the `readPassphrase` function (signature `func readPassphrase(prompt string) ([]byte, error)`). With that, I think this can be replaced by:
```suggestion
						passphrase, err := readPassphrase("Enter passphrase for identity '%s':")
						if err != nil {
							return "", err
						}
						return string(passphrase), nil
```

felixfontein (commented):
Mering #1692 introduced conflicts in keysource_test.go; my guess is that both PRs added something at the end which caused this, so it should be easy to fix.

age/tui.go
```diff
@@ -18,9 +18,18 @@ import (
...
 )
 
+const (
+	SopsAgePasswordEnv = "SOPS_AGE_PASSWORD"
+)
+
 // readPassphrase reads a passphrase from the terminal. It does not read from a
 // non-terminal stdin, so it does not check stdinInUse.
 func readPassphrase(prompt string) ([]byte, error) {
+	password := os.Getenv(SopsAgePasswordEnv)
```
felixfontein, line 28: I don't think this should be here. In fact I would remove it completely from this PR; having this as a separate PR for consideration after this PR and #1641 is merged is OK, but right now is not the time to add new features here.

age/keysource.go
```diff
@@ -371,3 +450,83 @@ func parseIdentities(identity ...string) (ParsedIdentities, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read passphrase: %v", err)
+	}
+	ii, err := age.NewScryptIdentity(pass)
+	if err != nil {
+		return nil, err
+	}
+	fileKey, err = ii.Unwrap(stanzas)
+	return fileKey, err
+}
```
felixfontein, line 532: This is basically copied from https://github.com/FiloSottile/age/blob/101cc8676386b0503571a929a88618cae2f0b1cd/cmd/age/encrypted_keys.go. Can you create a new file `encrypted_keys.go` with this content and add a similar license header as for `tui.go`?

age/keysource.go
```diff
@@ -326,11 +328,88 @@ func (key *MasterKey) loadIdentities() (ParsedIdentities, error) {
...
+					log.Warnf("encrypted identity '%s' didn't match file's recipients", n)
+				},
+			}}
+			identities = append(identities, ids...)
+		default:
+			ids, err := age.ParseIdentities(b)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse '%s' age identities: %w", n, err)
+			}
+			identities = append(identities, ids...)
```
felixfontein, line 411: Since this code is also heavily copied from https://github.com/FiloSottile/age/blob/101cc8676386b0503571a929a88618cae2f0b1cd/cmd/age/parse.go, it's better to move it to a separate file (also `encrypted_keys.go`?) Also moving this part to its own function (that can be moved to that file) will improve the code anyway, the function is getting quite long.

felixfontein (commented):
Thanks for updating the PR!

age/tui.go
tomaszduda23: It is not easy to test without it. The previous change had this code to side load password from test.
```
reader := bufio.NewReader(os.Stdin)
pass, err = reader.ReadString('\n')
if err == io.EOF {
	err = nil
}
```
Any idea how would I supply password from tests for /dev/tty?

felixfontein: Hmm, in that case, use https://stackoverflow.com/a/78532310 so that it only works in testing, and not in production builds.

age/keysource.go
tomaszduda23: done

age/keysource.go
tomaszduda23: done

age/tui.go
tomaszduda23: done

felixfontein (approved):

felixfontein:
(I'll wait until tomorrow with merging this to give others a chance to comment. In case anyone else wants to take a look, and doesn't have time til then, please write it here, then I'll wait longer. CC @getsops/maintainers.)

felixfontein:
@tomaszduda23 thanks a lot for your contribution and patience!

PR #1403, opened 2024-01-02
chore: add rust code formatting check

### LINKED ISSUE #1397: CI: add Rust code formatting check (for functional tests)
Reformatting can easily be done with `cd functional-tests && cargo fmt`.

Ref: https://github.com/getsops/sops/pull/1390#pullrequestreview-1799218671

shikharvashistha:
This commit adds rust code formatting to the CI

closes #1397

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -116,3 +116,7 @@ jobs:
       - name: Run tests
         run: cargo test
         working-directory: ./functional-tests
+
+      - name: Run Formatting
+        run: cargo fmt
+        working-directory: ./functional-tests
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -117,6 +117,6 @@ jobs:
         run: cargo test
         working-directory: ./functional-tests
 
-      - name: Run Formatting
-        run: cargo fmt
+      - name: Run Formatting Check
+        run: cargo fmt --check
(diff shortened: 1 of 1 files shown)


.github/workflows/cli.yml
```diff
@@ -116,3 +116,7 @@ jobs:
       - name: Run tests
         run: cargo test
         working-directory: ./functional-tests
+
+      - name: Run Formatting
+        run: cargo fmt
```
felixfontein, line 121: If you don't specify `--check`:
```suggestion
        run: cargo fmt --check
```
it will simply reformat and not output anything.

shikharvashistha: I think I've missed it. Thanks for pointing it out 👍.

felixfontein:
The test looks good, though I'm not sure whether the placement is good. Right now it is run as part of the functional tests.

I'm wondering whether we should create a new workflow for such linting-style tests. I think it won't hurt to add more static code checks in the future, like [staticcheck](https://github.com/dominikh/go-tools). What do you think @getsops/maintainers?

devstein:
@felixfontein I agree that creating a separate workflow for linting would be preferable. I don't have a strong opinion whether it's done in this PR or a future one.

@shikharvashistha thanks for contributing!

felixfontein:
Re-used in #1604, which has now been merged. Thanks!

PR #1434, opened 2024-02-09
Decryption: do not fail if no matching creation_rule is present in config file

### LINKED ISSUE #868: Decryption fails if file is not in the creation_rules
When decrypting a file with `sops -d file.yaml` sops fails with `error loading config: no matching creation rules found` even if the file contains the correct sops metadata. If I add this file to one `creation_rules` or remove the `.sops.yaml` file this is working as intended. 

I was expecting sops not to complain on decryption phase since everything that it needs is present in the metadata. Moreover the key is named `creation_rules`. It's confusing that you need a _creation_ rule to decrypt an _existing_ file.

felixfontein:
Since the config file loading is only done for backwards compatibility (and the result isn't used), we can simply skip it when only decrypting.

(This is not a problem for new-style decryption - `sops decrypt` subcommand - since that doens't load the config for the file.)

Fixes #868.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -1500,6 +1500,12 @@ func main() {
 	}, keyserviceFlags...)
 
 	app.Action = func(c *cli.Context) error {
+		isDecryptMode := c.Bool("decrypt")
+		isEncryptMode := c.Bool("encrypt")
+		isRotateMode := c.Bool("rotate")
+		isSetMode := c.String("set") != ""
+		isEditMode := !isEncryptMode && !isDecryptMode && !isRotateMode && !isSetMode
+
 		if c.Bool("verbose") {
 			logging.SetLevel(logrus.DebugLevel)
 		}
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -1511,26 +1517,30 @@ func main() {
...
 			log.Warn("More than one command (--encrypt, --decrypt, --rotate, --set) has been specified. Only the changes made by the last one will be visible. Note that this behavior is deprecated and will cause an error eventually.")
 		}
 
-		// Load configuration here for backwards compatibility (error out in case of bad config files)
-		_, err = loadConfig(c, fileNameOverride, nil)
-		if err != nil {
-			return toExitError(err)
+		// Load configuration here for backwards compatibility (error out in case of bad config files),
+		// but only when not just decrypting (https://github.com/getsops/sops/issues/868)
+		needsCreationRule := isEncryptMode || isRotateMode || isSetMode || isEditMode
```
devstein, line 1538: If I understand correctly, this would only be false if

```
isEncryptMode = false
isRotateMode = false
isSetMode = false
isDecryptMode = true
```

If this is correct, it might be easier to reason about if this is written as 
```suggestion
		// Load configuration here for backwards compatibility (error out in case of bad config files),
		// but only when not just decrypting (https://github.com/getsops/sops/issues/868)
		needsCreationRule := !(isEncryptMode || isRotateMode || isSetMode) && isDecryptMode
```

devstein (commented):
One readability comment, but otherwise looks good to me (code-wise)

cmd/sops/main.go
felixfontein: > If I understand correctly, this would only be false if
> 
> ```
> isEncryptMode = false
> isRotateMode = false
> isSetMode = false
> isDecryptMode = true
> ```

Yes, due to the way `isEditMode` works (it's `true` if no other mode is set).

> If this is correct, it might be easier to reason about if this is written as

I find the new expression harder to understand, and the intended behavior - the creation rule needs to be checked here if one of the modes is active that needs it later - less clear.

Also your expression has a boolean error, it evaluates to `true` in case `needsCreationRule` should be `false`, so it would have to be `needsCreationRule := isEncryptMode || isRotateMode || isSetMode || !isDecryptMode`.

sabre1041 (approved):
LGTM

felixfontein:
@devstein @sabre1041 thanks for reviewing this!

PR #1450, opened 2024-03-04
feat: improve format detection

rdeusser:
A common convention among sops users is to add `.enc` to the end of a file that's just been encrypted (mainly through a script of some sort; e.g. `sops -e [file].yaml > [file].yaml.enc`). Doing this causes an issue when trying to decrypt it: `sops -d [file].yaml.enc > [file].yaml`. Sops will error out when doing this because now the input and output types are to binary as the default unless you specify `--input-type` and `--output-type`. That's all well and good, but:

1. I shouldn't have to do this.
2. Not all encrypted files are going to have `.yaml`/`.yml` extensions, otherwise all sops commands could just set `--input-type` and `--output-type` and forget about it.

This PR adds a small quality of life improvement to sops that checks all the file extensions with binary as the default.

### DIFF
--- a/cmd/sops/formats/formats.go
+++ b/cmd/sops/formats/formats.go
@@ -1,6 +1,9 @@
 package formats
 
-import "strings"
+import (
+	"path/filepath"
+	"strings"
+)
 
 // Format is an enum type
 type Format int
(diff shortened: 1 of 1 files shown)


felixfontein:
Thanks for your contribution!

I'm ambivalent about this change. While this is clearly useful, it is a **breaking change** and might break existing workflows, and will even refuse to decrypt files which previously decrypted fine. Say for example that someone encrypted a file `foo.yaml.bar`, which wasn't a YAML file (but for some reason has `.yaml` in the filename), to `foo.yaml.bar.enc`. This would have been treated as a binary file by default, and thus the encrypted file would be in JSON, with tabs. With the code from this PR, it would try to load the encrypted file as YAML (because of the `.yaml` in there), which will fail since YAML does not allow tab indentation. But also if the user wants to encrpyt `foo.yaml.bar` again, which worked fine so far, SOPS will now fail since now it assumes its a YAML file.

(The README's main convention is to add the `.enc` extension *before* the main extension, like `file.enc.yaml` instead of `file.yaml.enc` - that way both SOPS and other tools will know that this is still a YAML file.)

rdeusser:
Hmm, I see your point. While it works great for my use-case, I can see how it could potentially break others.

The README is in .rst format for some reason and fails to load and so I didn't think to open it locally or view it raw. I don't see anything that says `.enc.[yaml|json|ini]` is the preferred convention, in fact both `.enc.yaml` and `.json.enc` are used so that might be a bit confusing.

Appreciate you taking the time to look at this PR! Since this isn't a good solution for the majority, I'll be closing it. Thanks!

felixfontein:
That the README isn't rendering is unfortunately due to a change by GitHub, they don't seem to want to fix it... We're currently looking at converting it (though IMO we first should merge some more PRs so we don't end up with a conflict fest...)

I wouldn't say `.enc.yaml`/`.enc.json`/... is the *preferred* convention, but it has some advantages, namely that most common tools (including SOPS and editors) detect the right format. (I'm personally using `.sops.yaml`/`.sops.json` instead of `.enc.yaml`/`.enc.json` so it's clearer that sops is involved :) )

PR #1474, opened 2024-03-27
exec-file: if `--filename` is used, use the provided filename without random suffix

### LINKED ISSUE #1473: `sops execfile` filename should not have a random suffix appended in --no-fifo mode
Hello,

Some programs (in this case, Chef `knife`) expect the file extension to reflect the data type of the file.

When performing `sops exec-file --filename tmp.json --no-fifo my-role.json 'knife role from file {}'`, `knife` breaks because the file name is not tmp.json, as requested, but something along the lines of `/private/var/folders/vy/_x8ql9xb6ehd634h9q00000gn/T/.sops2839518645/tmp.json3758396011`.

In my opinion, there is no real benefit for sops to essentially randomize the extension, as the whole path is essentially security by obscurity anyhow and adding one more unnecessary "layer" to it won't prevent shell globbing from catching the file name anyway.

When using a FIFO (default), the temporary file name is as expected, with no random suffix appended to the file name.

Thanks.

felixfontein:
Fixes #1473.

### DIFF
--- a/cmd/sops/subcommand/exec/exec.go
+++ b/cmd/sops/subcommand/exec/exec.go
@@ -3,12 +3,17 @@ package exec
 import (
 	"bytes"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 
 	"github.com/getsops/sops/v3/logging"
 
 	"github.com/sirupsen/logrus"
+)
+
+const (
+	FallbackFilename = "tmp-file"
 )
 
 var log *logrus.Logger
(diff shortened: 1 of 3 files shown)


cmd/sops/subcommand/exec/exec.go
```diff
@@ -55,7 +73,11 @@ func ExecWithFile(opts ExecOpts) error {
...
 		// will block if there is no reader present
-		filename = GetPipe(dir, opts.Filename)
+		filename = opts.Filename
+		if filename == "" {
+			filename = FallbackFilename
+		}
+		filename = GetPipe(dir, filename)
 		go WritePipe(filename, opts.Plaintext)
 	} else {
 		handle := GetFile(dir, opts.Filename)
```
devstein, line 83: Might be worth leaving a comment as to why this is `opts.Filename` vs `filename` to avoid future confusion (I could see myself changing this in the future 😅 )

Alternatively, you could pass `filename` to `GetFile` and check for `filename == FallbackFilename`, but then there is the edge-case where a user intentionally passes `"tmp-file"` as the filename

devstein (commented):
LGTM - one comment. I think it's good to merge regardless

cmd/sops/subcommand/exec/exec.go
felixfontein: Good point, I've updated the PR accordingly (and rebased while at it).

> Alternatively, you could pass `filename` to `GetFile` and check for `filename == FallbackFilename`, but then there is the edge-case where a user intentionally passes `"tmp-file"` as the filename

Yeah, that's why I did it this way ;-)

felixfontein:
@devstein thanks for reviewing and merging!

PR #1477, opened 2024-03-31
sops: fix typo in interface name

duthils:
Just came across this typo and fixed it :)


felixfontein (approved):
Thanks for noticing and fixing this!

PR #1493, opened 2024-04-23
Feature: `merge` key for key groups and *make keys unique*

jonasbadstuebner:
This closes #1123 

The test is not very great (as in: "It doesn't check very much"), but since it was not too hard to implement, I think this is okay.

If you have any comments, please let me know.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -94,6 +94,7 @@ type configFile struct {
 }
 
 type keyGroup struct {
+	Merge   []keyGroup
 	KMS     []kmsKey
 	GCPKMS  []gcpKmsKey  `yaml:"gcp_kms"`
 	AzureKV []azureKVKey `yaml:"azure_keyvault"`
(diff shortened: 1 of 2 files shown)


jonasbadstuebner:
Just so I have said it: This does not ensure that a key is included only once. Maybe this is a desirable feature when people start merging `key_groups` now?

But that's is not any different from the overall `key_groups` behavior. And as far as I can tell, sops can handle it if there are multiple `recipient`s matching. (I think it just takes the first matching one, I haven't checked this any futher)

jonasbadstuebner:
Should also mention:
Fixes #878

jonasbadstuebner:
@felixfontein Sorry to bother you again, but what is the current state of sops, it seems to be very slow moving and the last patch release has been a long time ago.

Why does a small MR like this take so much time to review and merge and when is the next patch release planned? 3.9.0 is on the road map for a very long time already, do you need help or is it something else?

felixfontein:
It seems that most maintainers currently have no time to review PRs. Since I don't want to merge PRs that I don't fully understand without a second review, this means that rarely any PRs get merged currently.

I was hoping for a soonish 3.9.0 release for some time already, I have no idea when it will happen.

jonasbadstuebner:
If you need maintainers, I would be happy to support sops. Who would be in charge of deciding this and what would I have to do?

felixfontein:
I'm not sure how the process works, but joining the #sops-dev channel on the CNCF Slack (https://github.com/getsops/sops/blob/main/CONTRIBUTING.md#communication, the second link allows you to get an invite for the CNCF Slack, and the former then allows you to join that channel) is probably a good idea.

config/config.go
```diff
@@ -94,6 +94,7 @@ type configFile struct {
 }
 
 type keyGroup struct {
+	Merge   []keyGroup
 	KMS     []kmsKey
```
felixfontein, line 98: I'm wondering whether you can create an infinite loop by using YAML anchors. I don't think you can, but YAML anchors are a feature of YAML I try to avoid, so maybe I'm missing a trick here.

config/config_test.go
```diff
@@ -120,6 +120,40 @@ creation_rules:
...
+        - qux
+        gcp_kms:
+        - resource_id: bar
+        - resource_id: baz
+        azure_keyvault:
+        - vaultUrl: https://bar.vault.azure.net
+          key: bar-key
+          version: barversion
+        hc_vault:
+        - 'https://baz.vault:8200/v1/baz/keys/baz-key'
```
felixfontein, line 154: Maybe add another keygroup to `merge` that itself has a `merge` key with more key groups.

felixfontein (commented):
Some thoughts about recursion ;)

> Just so I have said it: This does not ensure that a key is included only once. Maybe this is a desirable feature when people start merging `key_groups` now?

I tend to say yes. We definitely have to make sure to avoid breaking backwards compatibility, though. Maybe it should only be used for everything inside `merge`, but not for the ones added on top-level? I.e. remove duplicates in line 193, after processing `group.Merge` and before processing `group.Age`.

config/config_test.go
jonasbadstuebner: Will do.
I will rename the keys a little bit, so they are easier to see through.

config/config.go
jonasbadstuebner: I also don't think it is possible, because YAML anchors only resolve references when parsed (and the parser supports it). But I will double check this.

jonasbadstuebner:
> Some thoughts about recursion ;)
> 
> > Just so I have said it: This does not ensure that a key is included only once. Maybe this is a desirable feature when people start merging `key_groups` now?
> 
> I tend to say yes. We definitely have to make sure to avoid breaking backwards compatibility, though. Maybe it should only be used for everything inside `merge`, but not for the ones added on top-level? I.e. remove duplicates in line 193, after processing `group.Merge` and before processing `group.Age`.

This implementation would allow
```
key_groups:
- key1
- merge:
  - key1
```
Which would result in duplicated `key1`. Is this the desired behavior?

config/config.go
jonasbadstuebner: Tested:
```
creation_rules:
  - path_regex: ""
    key_groups:
    - merge: &merge1
      - merge: *merge1
```
Result:
```
Expected nil, but got: &errors.errorString{s:"Could not unmarshal config file: yaml: anchor 'merge1' value contains itself"}
```
So I think it is safe to say that an infinite loop like this could not happen.

config/config_test.go
jonasbadstuebner: Please confirm that the updated test is more complete.

jonasbadstuebner:
I implemented the uniqueness as requested. The top-level is not touched.

config/config.go
```diff
@@ -180,38 +181,73 @@ type Config struct {
...
+	}
+
+	var keyGroup sops.KeyGroup
+
+	// Add only unique values from mergeGroup to keyGroup
+	unique := make(map[string]bool)
+	for _, v := range mergeKeyGroup {
+		// keytype + toMap make a unique combination per key type
+		valueMap := v.ToMap()
+		delete(valueMap, "created_at")
```
felixfontein, line 201: Why not use `v.ToString()` instead of a string representation of `v.ToMap()`?

The only problem I see with that is that `kms.ToString()` only uses `arn` and not `role`, `context`, and `awsProfile`. I have no idea whether that's OK or not... From https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-id it *seems* that the ARN uniquely identifies the key.

Does anyone from @getsops/maintainers have some insight here?

config/config_test.go
```diff
@@ -324,6 +435,13 @@ func TestLoadConfigFileWithGroups(t *testing.T) {
 	assert.Equal(t, expected, conf)
 }
 
+func TestLoadConfigFileWithMerge(t *testing.T) {
+	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithMergeType, t), "/conf/path", "whatever", nil)
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(conf.KeyGroups[0]))
```
felixfontein, line 441: I'd also add a test for the length of `conf.Keygroups` itself:
```suggestion
	assert.Equal(t, 2, len(conf.KeyGroups))
	assert.Equal(t, 1, len(conf.KeyGroups[0]))
```

config/config.go
```diff
@@ -180,38 +181,73 @@ type Config struct {
...
+
+		key := fmt.Sprintf("%T/%v", v, valueMap)
+		if _, ok := unique[key]; ok {
+			// key already contained, therefore not unique
+			continue
+		}
+
+		keyGroup = append(keyGroup, v)
+		unique[key] = true
+	}
```
felixfontein, line 211: How about moving the deduplication code to a function
```go
func deduplicateKeygroup(group keyGroup) sops.KeyGroup {
  ...
}
```
? That would make `extractMasterKeys` easier to read, and the deduplication code easier to move around.

felixfontein:
> This implementation would allow
> 
> ```
> key_groups:
> - key1
>   merge:
>   - key1
> ```
> 
> Which would result in duplicated `key1`. Is this the desired behavior?

I wouldn't say "desired", but I think it's a good compromise between backwards compatibility (you were able to duplicate keys before) and not deduplicating at all.

Another possibility would be to deduplicate everything once `group.Merge` contains something. In that case no deduplication is done for backwards compatibility if the `merge` feature isn't used, but once it is used the result is always deduplicated.

config/config.go
jonasbadstuebner: This difference on (at least) the `kms` keys is the reason.

I found this other place, where keyGroups are compared to each other:
https://github.com/getsops/sops/blob/1c46d2492192cc55bc33baf6b9b78889fe3e795a/cmd/sops/common/common.go#L401
and here `ToString` is used:
https://github.com/getsops/sops/blob/1c46d2492192cc55bc33baf6b9b78889fe3e795a/cmd/sops/common/common.go#L423
To have the same behavior, I have changed my code to use `ToString` too.

jonasbadstuebner:
To me, changing the behavior of the top-level based on a key being present would be unintuitive.
Does removing duplicates on the top-level really break backwards compatibility though? If you have 2 keys in this list that are basically the same, it should not break your en- or decryption to only have the key included once. But I did not test this very deeply.

felixfontein:
Having thought some more about this, I tend to agree that removing duplication in generally should be OK.

It's mainly breaking in the sense that if you re-encrypt a file that has duplicated keys (like when editing the encrypted content), suddenly some keys are removed, which can be quite confusing for users. But as long as we explicitly announce deduplication in the changelog, I think it should be OK.

(I would also mention it in the PR's title, so it shows up prominently in the git commit log.)

felixfontein (approved):
Looks good to me. Let's wait a bit more whether someone comments on https://github.com/getsops/sops/pull/1493#discussion_r1625318840.

sabre1041 (approved):
LGTM

felixfontein:
@jonasbadstuebner thanks for your contribution!

PR #1494, opened 2024-04-23
correct `decrypt`/`encrypt` to `-d`/`-e` in README

jmhodges:
The README described the use of `sops decrypt` and `sops encrypt` commands that no longer exist in modern versions.

The correct commands seem to be `sops -d` and `sops -e`.

I ran into this while trying to get the `git diff`ing working as described in this doc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,7 +154,7 @@ To decrypt a file in a ``cat`` fashion, use the ``-d`` flag:
 
 .. code:: sh
 
-    $ sops decrypt mynewtestfile.yaml
+    $ sops -d mynewtestfile.yaml
 
 SOPS encrypted files contain the necessary information to decrypt their content.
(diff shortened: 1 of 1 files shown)


felixfontein (changes requested):
The README refers to the version in the `main` branch, not to any yet released version.

jmhodges:
What is the actually accurate documentation link, then? Folks have been linking to this and leading folks astray

felixfontein:
The README for 3.8.1 is at https://github.com/getsops/sops/blob/v3.8.1/README.rst. Right now there is unfortunately no way to always view the docs for the latest released version, you have to use the README.rst from the corresponding tag...

felixfontein:
Closing since 3.9.0 has been released. Thanks for your contribution!

PR #1501, opened 2024-04-30
Fix wrong parameter in README

chrgraefe:


felixfontein (changes requested):
This parameter is not wrong, it only does not yet correspond to a released version of sops. See also #1494.

@getsops/maintainers we REALLY need to start getting forward to a 3.9.0 release.

felixfontein:
Closing since 3.9.0 has been released. Thanks for your contribution!

PR #1513, opened 2024-05-24
feat: follow best practices with regards to naming .env files

DrakeEvans:
Many libraries and dev tools use .env files.  Typically different files are used for different environments like dev, staging, and production.  Typically, these are named .env.production with the designation happening _**after**_ the .env.

This change allows sops to recognize dotenv files named with .env at the beginning of the filename in addition to the suffix (current functionality)

### DIFF
--- a/cmd/sops/formats/formats.go
+++ b/cmd/sops/formats/formats.go
@@ -1,6 +1,9 @@
 package formats
 
-import "strings"
+import (
+	"path/filepath"
+	"strings"
+)
 
 // Format is an enum type
 type Format int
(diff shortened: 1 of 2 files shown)


felixfontein:
Thanks for your contribution! This is unfortunately a **breaking change**, see https://github.com/getsops/sops/pull/1450#issuecomment-1978049930 for details.

DrakeEvans:
@felixfontein any docs on the sops.yaml configuration file?  I would be happy to achieve a similar result by specifying the --input-type and --output-type in the config instead.  I just cant seem to find any docs on it but happy to make the change myself

felixfontein:
@DrakeEvans it's always a breaking space, since it affects how SOPS identifies the type of some files. It will suddenly use the env store instead of the binary store for some files.

felixfontein:
(A possible solution would be to embed - if asked to - the original store type in the encrypted file, so SOPS doesn't have to guess it when decrypting. The 'if asked to' part here is important since otherwise it's also breaking backwards compatibility. But that's a more complex change that first needs some discussion...)

DrakeEvans:
Yeah Im wondering if its easier to just specify the input-type/output-type in the config file.  Does that functionality already exist?

felixfontein:
I don't think it exists so far.

PR #1522, opened 2024-06-06
pgp: don't shorten key fingerprints

### LINKED ISSUE #1365: unable to force specific gpg subkey
## situation

I use sops to manage secrets for my projects, all using GPG for encryption.

I have a GPG master key with several encryption subkeys:

- One on my YubiKey (oldest)
- One on my desktop at home
- One on my desktop at work (newest)

I just recently added the desktop and work subkeys, and am hoping to be able to configure sops to encrypt my secrets with all of these subkeys so that I can recover in the eventuality that I lose the YubiKey. 

## state of the world

As far as I understand:

- sops shells out to gpg to encrypt/decrypt
- When given a master key fingerprint to use for encryption, gpg will always use the newest encryption subkey
- By default, any subkey fingerprint passed to gpg will cause it to lookup the master key first, then select an encryption subkey (which will always be the newest one per above)
- It is possible to override this behavior by passing a specific subkey fingerprint followed by an exclamation mark (disables master key lookup and forces the use of the specified fingerprint)
- sops truncates key fingerprints to 16 characters during encryption for compatibility reasons

As a result, it isn't possible to tell sops to use a specific key fingerprint, since for a keyid like `0123456789ABCDEF!`, sops will produce `123456789ABCDEF!` , which is rejected by gpg as malformed.

In my case, this means that everything sops encrypts that points at my gpg key can be encrypted only with my work encryption key, as it's the newest.

## proof of concept

The following patch solves this problem for me and lets me specify an exclamation-mark fingerprint to sops:

```diff
diff --git a/pgp/keysource.go b/pgp/keysource.go
--- pgp/keysource.go
+++ pgp/keysource.go
@@ -625,9 +625,6 @@
 // shortenFingerprint returns the short ID of the given fingerprint.
 // This is mostly used for compatibility reasons, as older versions of GnuPG
 // do not always like long IDs.
 func shortenFingerprint(fingerprint string) string {
-	if offset := len(fingerprint) - 16; offset > 0 {
-		fingerprint = fingerprint[offset:]
-	}
 	return fingerprint
 }
```

## proposed upstream

In `shortenFingerprint`, special-case strings with terminal `!`, chopping the string to 17 characters rather than 16, retaining the full key and the `!`. If amenable, I could PR this.

### related

#808

tilpner:
As @mammothbane already identified in #1365, the pgp module is stripping the trailing exclamation mark from fingerprints that pgp uses to identify specific subkeys.
Because the shortened fingerprint refers to the whole key instead of just the subkey, I can't decrypt any secrets I encrypt for that subkey.

According to the doc comment, this was meant for compatibility with older GPG versions. I don't know which incompatibilities @hiddeco was referring to here, or if they are still relevant.

https://github.com/getsops/sops/blob/1c46d2492192cc55bc33baf6b9b78889fe3e795a/pgp/keysource.go#L633-L635

Fixes #1365

### DIFF
--- a/pgp/keysource.go
+++ b/pgp/keysource.go
@@ -321,8 +321,6 @@ func (key *MasterKey) encryptWithOpenPGP(dataKey []byte) error {
 // PGP key that belongs to Fingerprint. It sets EncryptedDataKey, or returns
 // an error.
 func (key *MasterKey) encryptWithGnuPG(dataKey []byte) error {
-	fingerprint := shortenFingerprint(key.Fingerprint)
-
 	args := []string{
 		"--no-default-recipient",
 		"--yes",
@@ -331,7 +329,7 @@ func (key *MasterKey) encryptWithGnuPG(dataKey []byte) error {
 		"-r",
 		key.Fingerprint,
 		"--trusted-key",
-		fingerprint,
+		key.Fingerprint,
 		"--no-encrypt-to",
 	}
 	stdout, stderr, err := gpgExec(key.gnuPGHomeDir, args, bytes.NewReader(dataKey))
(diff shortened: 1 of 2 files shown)


felixfontein:
> According to the doc comment, this was meant for compatibility with older GPG versions. I don't know which incompatibilities @hiddeco was referring to here, or if they are still relevant.

I also don't know whether this is still needed, but to avoid breaking changes for users I've created a PR which keeps the trimming, but makes sure to handle `!` correctly: #1720

I think eventually we should remove the trimming function (as implemented in this PR), but I'm not sure when it's time to do that. Though definitely not in a bugfix release :)

Luflosi:
Since the original problem this PR aimed to fix was solved by https://github.com/getsops/sops/pull/1720, I think this can be closed.

PR #1523, opened 2024-06-07
feat: normalize path for creation rules

### LINKED ISSUE #1036: path_regex doesn't normalize paths with filepath.ToSlash
Currently to support path_regexes that should work on windows and other systems you need to manually handle the different separators with `(/|\\)` like this. 

```  
- path_regex: staging(/|\\).*\.secret\.(yaml|json)$
  gcp_kms: projects/.../sops-key-staging
```

This is a bit cumbersome and a bit error prone to use and would be nicer if we could just use slashes like this:

```
- path_regex: staging/.*\.secret\.(yaml|json)$
  gcp_kms: projects/.../sops-key-staging
```

Prince213:
Fixes #1036.

This may be a breaking change. 🤔


config/config.go
```diff
@@ -368,6 +368,8 @@ func parseCreationRuleForFile(conf *configFile, confPath, filePath string, kmsEn
 
 	// compare file path relative to path of config file
 	filePath = strings.TrimPrefix(filePath, configDir+string(filepath.Separator))
+	// replace \ with /
+	filePath = filepath.ToSlash(filePath)
```
felixfontein, line 372: For everyone using regexes that support both Windows and Unix paths, this is not a breaking change. But for users that use Windows paths exclusively and wrote their regexes that way, it's a breaking change since their regexes won't work anymore.

To avoid breaking changes, we'll have to (for some time) keep the old behavior, but warn the user if it will change in the future. So better do something like:
```suggestion
	filePathAlt := filepath.ToSlash(filePath)
```
here, and below before calling `reg.MatchString(filePath)` do `newMatch := reg.MatchString(filePathAlt)` and then emit a warning if it's value is different from `reg.MatchString(filePath)`.

felixfontein (commented):
Thanks for your contribution!

config/config.go
Prince213: Let's also add a option to silence that warning (by opting in to the new behavior).

felixfontein: In Ansible we usually proceed as follows:
1. Add a new option that allows to select the behavior. The default is the old behavior.
2. A version later, deprecate the option's default value - i.e. start warning the user if they didn't specify a value for that option and if the change to the new default would produce a change. Then users have time to figure out whether they prefer the old or new behavior, and can explicitly set the option to the value they prefer (and as soon as it has a value, there will be no more warnings).
3. Eventually flip the default of the option to the new value.

@getsops/maintainers what do you think, and what would you prefer?

PR #1527, opened 2024-06-11
chore(deps): Update Go dependencies

reneleonhardt:
### Features
* Update Go dependencies
* Release with Go 1.22
* Update alpine


Makefile
```diff
@@ -106,11 +105,11 @@ install-staticcheck:
 
 .PHONY: install-goreleaser
 install-goreleaser:
-	$(call go-install-tool,$(GORELEASER),github.com/goreleaser/goreleaser@$(GORELEASER_VERSION),$(GORELEASER_VERSION))
+	$(call go-install-tool,$(GORELEASER),github.com/goreleaser/goreleaser/v2@$(GORELEASER_VERSION),$(GORELEASER_VERSION))
 
 .PHONY: install-syft
 install-syft:
-	$(call go-install-tool,$(SYFT),github.com/anchore/syft/cmd/syft@$(SYFT_VERSION),$(SYFT_VERSION))
+	curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- $(SYFT_VERSION)
```
felixfontein, line 112: I don't think the combination `curl | sh` is acceptable. This is a very bad practice.

felixfontein (commented):
Thanks for your contribution! I think your PR is doing way too many things in one PR that should happen in multiple smaller PRs. Can you please split it up into smaller parts, such as:

1. Updating Go deps.
2. Extending CI to build multiple deps.
3. Updating the release process and using dependabot for updating Dockerfile.
4. Updating Rust dependencies and using dependabot for updating Cargo.toml.
5. Updating Makefile.

Thanks!

Makefile
reneleonhardt: ? https://github.com/anchore/syft#recommended

> ## Installation
> 
> Syft binaries are provided for Linux, macOS and Windows.
> 
> ### Recommended
>
> ```bash 
> curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
> ```

felixfontein: In that case the project itself is giving extremely bad suggestions of how to install itself. `curl | sh` should **never** be done if you don't control the shell file that's downloaded yourself and are 100% sure it's ok.

.github/workflows/release.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         id: goreleaser
         uses: goreleaser/goreleaser-action@286f3b13b1b49da4ac219696163fb8c1c93e1200 # v6.0.0
         with:
-          version: 1.21.x
+          version: 1.22.x
```
felixfontein, line 68: This is **not** the Go version BTW, but the goreleaser version.

Ref: https://github.com/getsops/sops/issues/1547#issuecomment-2196460328

reneleonhardt: Interesting, this is confusing, why would you use a goreleaser version from Sept 2023 if 1.26.x is the last 1.x? 🤔
I suggest "updating" to 1.26 first, but 2.1 is around the corner...

felixfontein:
Closing since this is pretty outdated. Thanks for your contribution.

PR #1530, opened 2024-06-16
chore(deps): Update Rust dependencies

reneleonhardt:
### Features
* Update Rust dependencies (tempdir was deprecated)

### DIFF
--- a/functional-tests/Cargo.toml
+++ b/functional-tests/Cargo.toml
@@ -5,9 +5,9 @@ edition = "2021"
 authors = ["Adrian Utrilla <adrianutrilla@gmail.com>"]
 
 [dependencies]
-tempdir = "0.3.5"
+tempfile = "3"
 serde = "1.0"
-serde_json = "1.0.99"
-serde_yaml = "0.9.22"
+serde_json = "1.0"
+serde_yaml = "0.9"
 serde_derive = "1.0"
-lazy_static = "1.4.0"
+lazy_static = "1.5"
--- a/functional-tests/Cargo.toml
+++ b/functional-tests/Cargo.toml
@@ -5,9 +5,9 @@ edition = "2021"
 authors = ["Adrian Utrilla <adrianutrilla@gmail.com>"]
 
 [dependencies]
-tempfile = "3"
+tempfile = "3.12.0"
 serde = "1.0"
-serde_json = "1.0"
-serde_yaml = "0.9"
+serde_json = "1.0.128"
+serde_yaml = "0.9.34"
 serde_derive = "1.0"
-lazy_static = "1.5"
+lazy_static = "1.5.0"
--- a/functional-tests/src/lib.rs
+++ b/functional-tests/src/lib.rs
@@ -1,11 +1,9 @@
+#[cfg_attr(test, macro_use)]
+extern crate lazy_static;
 extern crate serde;
 extern crate serde_json;
 extern crate serde_yaml;
-extern crate tempdir;
-#[macro_use]
-extern crate lazy_static;
-#[macro_use]
-extern crate serde_derive;
+extern crate tempfile;
 
 #[cfg(test)]
 mod tests {
@@ -19,7 +17,8 @@ mod tests {
     use std::io::{Read, Write};
     use std::path::Path;
     use std::process::Command;
-    use tempdir::TempDir;
+    use tempfile::Builder;
+    use tempfile::TempDir;
     const SOPS_BINARY_PATH: &'static str = "./sops";
     const KMS_KEY: &'static str = "FUNCTIONAL_TEST_KMS_ARN";
 
@@ -36,8 +35,10 @@ mod tests {
     }
 
     lazy_static! {
-        static ref TMP_DIR: TempDir =
-            TempDir::new("sops-functional-tests").expect("Unable to create temporary directory");
+        static ref TMP_DIR: TempDir = Builder::new()
+            .prefix("sops-functional-tests")
+            .tempdir()
+            .expect("Unable to create temporary directory");
     }
 
     fn prepare_temp_file(name: &str, contents: &[u8]) -> String {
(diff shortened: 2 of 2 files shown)


functional-tests/Cargo.toml
```diff
@@ -5,9 +5,9 @@ edition = "2021"
 authors = ["Adrian Utrilla <adrianutrilla@gmail.com>"]
 
 [dependencies]
-tempdir = "0.3.5"
+tempfile = "3"
 serde = "1.0"
-serde_json = "1.0.99"
-serde_yaml = "0.9.22"
+serde_json = "1.0"
+serde_yaml = "0.9"
```
felixfontein, line 11: Why you are removing the patch versions for most of the dependencies? Did you verify that the tests also work fine with lower patch versions?

felixfontein (commented):
Thanks for your contribution!

felixfontein:
@reneleonhardt ping

.github/dependabot.yaml
```diff
@@ -2,7 +2,7 @@ version: 2
 
 updates:
   - package-ecosystem: "docker"
-    directory: "/"
+    directory: "/.release"
```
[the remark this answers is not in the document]
reneleonhardt: Unrelated? Did you ever wonder why there have been no updates in https://github.com/getsops/sops/commits/main/.release? 😉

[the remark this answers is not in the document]
felixfontein: This has nothing to do with Rust. This PR is for updating Rust dependencies.

functional-tests/Cargo.toml
reneleonhardt: It seems best practice for Rust dependencies to require always the latest patch version in every build.
How does this affect your tests results?
I tested what I could, your project doesn't provide e2e test scripts for developers, so the 2 vault tests failed, therefore contributors have to rely on your CI pipelines.

Wouldn't it be much better for your users if you would care more for build security than allegedly incompatible patch versions of test dependencies? 🤔
https://blog.rust-lang.org/2023/08/03/cve-2023-38497.html
Why are you using an ancient rust 1.70.0 toolchain since 2 days?

felixfontein: > It seems best practice for Rust dependencies to require always the latest patch version in every build.

Sounds good to me.

> Wouldn't it be much better for your users if you would care more for build security than allegedly incompatible patch versions of test dependencies? 🤔

I'm not sure what you're trying to imply here. I asked a simple question here, I didn't say you have to revert to the previous state. I personally do care for build security, but I'm not that familiar with all the tools used in the SOPS build pipeline.

> https://blog.rust-lang.org/2023/08/03/cve-2023-38497.html Why are you using an ancient rust 1.70.0 toolchain since 2 days?

How come you assume we're using it since 2 days? We have been using Rust 1.70.0 for a long time already, the only thing that changed is the place where 1.70.0 is configured. (Which makes it easier to update that version to something less ancient.)

felixfontein: Also in CI one often pins all dependencies, and uses things like Dependabot to update them. That prevents random CI failures due to updated dependencies that sneaked in without explicit approval.

Since you added a Dependabot config for Cargo in this PR, I don't really understand why you remove the patch versions here.

felixfontein:
Can you please either remove the Dependabot config for Rust, or re-add the full version numbers in `Cargo.toml`? And remove the Docker dependabot config from this PR? Thanks.

felixfontein:
Why did you remove the Dependabot config for Rust/Cargo?

felixfontein:
(Also note that with commit messages like that, the chance that I will merge this are pretty much zero. Maybe someone else will merge it, but I definitely will not.)

reneleonhardt:
You required me to remove the Dependabot config for Rust, even if it makes no sense for an `Update Rust dependencies
` PR, if I want to have a slight chance to get my contribution to your project merged.
So I did against my better judgement, that's why I had to phrase those commits accordingly about the consequences for the maintainers who didn't configure working Dependabot rules yet and made this PR necessary in the first place because no one stepped up and replaced tempdir for example which has been deprecated 6 years ago.
Then you copied my contribution to rip it out-of-context without asking for my permission or if that even would make sense, an Update PR should update *and* configure Dependabot to update automatically int the future to avoid bugs and security vulnerabilities 🤷

felixfontein:
> You required me to remove the Dependabot config for Rust, even if it makes no sense for an `Update Rust dependencies ` PR, if I want to have a slight chance to get my contribution to your project merged.

I hope you're not actively trying to misinterpret what I wrote:

> Can you please either remove the Dependabot config for Rust, or re-add the full version numbers in Cargo.toml (the second is preferable from my side)?

There is an EITHER-OR condition. That means, either do the first, or the second. I also explicitly wrote that I would prefer the second, not the first.

You decided to do both, which is not what I asked you to do, and what we both agree is not a good idea.

> Then you copied my contribution to rip it out-of-context without asking for my permission

I did **not** copy your contribution, I simply recreated parts of it since a) it was no longer part of this PR, and b) it was something I would have liked to merge already in June. Also if you compare the Cargo part of your commit (11ebfc206023d7ffba631803652ba1e81dc2a48b) with my commit (576f809787e0a27c96bac41726220f5acae0dd24), you can see that I did not copy your contribution, but based it on the existing entry for GH Actions.

felixfontein:
Also, according to the official Cargo documentation (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), `serde_json = "1.0.128"` allows all versions of serde_json that are `>= 1.0.128` and `< 2.0.0`, while `serde_json = "1.0"` allows all versions that are `>= 1.0` and `< 2.0.0`. So there's no good reason to shorten version numbers, unless you want to allow even older versions of the dependencies to match.

PR #1544, opened 2024-06-26
Update sops.go comment

rjeczkow:
Keys (not values) in cleartext.


felixfontein (approved):
Thanks a lot for fixing that!

PR #1427, opened 2024-02-01
build with go 1.21 and update dependencies

b-dean:
also run `go get -u ./...` to update all the dependencies

in particular, this fixes these vulnerabilities that show up in our scans:

| Component | Version | Vulnerability | Severity |
|-----------|---------|---------------|----------|
| github.com/cloudflare/circl | v1.3.3 | [GHSA-9763-4f94-gfch](https://github.com/advisories/GHSA-9763-4f94-gfch) | high |
| go | 1.20.8 | [CVE-2023-45285](https://nvd.nist.gov/vuln/detail/CVE-2023-45285) | high |
| go | 1.20.8 | [CVE-2023-45283](https://nvd.nist.gov/vuln/detail/CVE-2023-45283) | high |
| go | 1.20.8 | [CVE-2023-39325](https://nvd.nist.gov/vuln/detail/CVE-2023-39325) | high |
| go | 1.20.8 | [CVE-2023-39323](https://nvd.nist.gov/vuln/detail/CVE-2023-39323) | high |

PR #1548, opened 2024-06-28
Explicitly mark goreleaser's version, to avoid confusion with the Go version

felixfontein:
Prevent mistakes such as in #1427 and #1527 where this is assumed to be a Go version.

Ref: https://github.com/getsops/sops/issues/1547#issuecomment-2196460328


onedr0p (approved):

reneleonhardt:
Why didn't you update your ancient goreleaser version too instead of only adding a comment while you're waiting to upgrade to 2.x?
1.26.2 was the last 1.x.

felixfontein:
Because I'd like to add tests first (https://github.com/getsops/sops/issues/1547#issuecomment-2196411365) before upgrading it.

PR #1553, opened 2024-07-07
Fix the URL used for determining the latest SOPS version

felixfontein:
Right now the code does a HEAD to https://github.com/mozilla/sops/releases/latest and then follows the 301 redirect and does a HEAD to https://github.com/getsops/sops/releases/latest to obtain the version from the 302 redirect that returns.

Since the repo has been renamed, I think we can skip the mozilla/sops step and directly ask getsops/sops.

(The original URL was added in #1261, I'm not sure why we used mozilla/sops there instead of getsops/sops.)


onedr0p (approved):

felixfontein:
@onedr0p thanks for reviewing this!

PR #1569, opened 2024-07-30
Note on the name of the configuration file

### LINKED ISSUE #1567: Sops cannot find config file if it's named .sops.yml (instead of .sops.yaml)
I don't know if it is a hard requirement that the sops config file is exactly `.sops.yaml` for it to be auto-discovered, or if it should accept the alternative `.yml` extension too, but it took me a while to realize this was the issue I was having.

So, it might be good to have a warning in the docs stating that it should match exactly to save some time to anyone facing the same issue. Or even better would be to allow the file to have either of those extensions.

I think the code where the name is defined to match that specific file name is [here](https://github.com/getsops/sops/blob/c921bf5e4d9339ce1e593fce7a575f55f0c262b4/config/config.go#L49) just for reference.

Thank you for the great work!

ErraticMaker:
Making explicit that only files named `.sops.yaml` will be auto-discovered. And alternative names (like `.sops.yml`) need to be used with the `--config` flag.

Closes #1567

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -669,6 +669,9 @@ of all new files. If your secrets are stored under a specific directory, like a
 ``git`` repository, you can create a ``.sops.yaml`` configuration file at the root
 directory to define which keys are used for which filename.
 
+Note: The file needs to be named ``.sops.yaml``. Other names (i.e. ``.sops.yml``) won't be automatically
+discovered by sops. You'll need to pass the ``--config .sops.yml`` option for it to be picked up.
+
 Let's take an example:
 
 * file named **something.dev.yaml** should use one set of KMS A, PGP and age
--- a/README.rst
+++ b/README.rst
@@ -669,8 +669,10 @@ of all new files. If your secrets are stored under a specific directory, like a
 ``git`` repository, you can create a ``.sops.yaml`` configuration file at the root
 directory to define which keys are used for which filename.
 
-Note: The file needs to be named ``.sops.yaml``. Other names (i.e. ``.sops.yml``) won't be automatically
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -669,6 +669,9 @@ of all new files. If your secrets are stored under a specific directory, like a
 ``git`` repository, you can create a ``.sops.yaml`` configuration file at the root
 directory to define which keys are used for which filename.
 
+Note: The file needs to be named ``.sops.yaml``. Other names (i.e. ``.sops.yml``) won't be automatically
+discovered by sops. You'll need to pass the ``--config .sops.yml`` option for it to be picked up.
```
felixfontein, line 673:
```suggestion
.. note::

  The file needs to be named ``.sops.yaml``. Other names (i.e. ``.sops.yml``) won't be automatically
  discovered by SOPS. You'll need to pass the ``--config .sops.yml`` option for it to be picked up.
```

felixfontein (commented):
Thanks for your contribution!

README.rst
ErraticMaker: I was gonna use rst note admonition syntax, but none of the other notes in the file are using it and also. It doesn't renders nicely in GitHub, it juts adds one extra line break 😢. See [this issue](https://github.com/github/markup/issues/1682) for reference.

Anyways, I'll do the change if you still think is fine. Such a minor thing, I don't want to waste your valuable time.

zubieta (approved):

felixfontein (approved):

felixfontein:
@ErraticMaker thanks for improving the docs!

PR #1571, opened 2024-07-31
chore: omitempty metadata to reduce the size of stored config

cgetzen:
This reduces the sops configuration stored in files to its bare minimum:
```yaml
# example with GCP KMS

someConfig:
...
sops:
    gcp_kms:
        - resource_id: projects/some-project/locations/global/keyRings/some-keyring/cryptoKeys/some-key
          created_at: "2024-08-31T01:09:49Z"
          enc: someenc=
    lastmodified: "2024-08-31T21:53:36Z"
    mac: ENC[AES256_GCM,data:/somemac=,tag:sometag==,type:str]
    encrypted_regex: someregex
    mac_only_encrypted: true
    version: 3.9.0

```


felixfontein:
While this is nice, it's also causing a lot of changes to exisitng SOPS-encrypted files when these are changed (edited, keys updated, rotated, etc.).

@getsops/maintainers we have to figure out when to include this (it definitely should not go into a bugfix release) - assuming nobody objects to having this at all?

cgetzen:
Thanks for the review @felixfontein. What do you think about adding the [3.10](https://github.com/getsops/sops/milestone/7) label so that it could make it into the next minor version?

felixfontein:
@cgetzen I added it to the Discussion milestone so it can be moved to the appropriate milestone once it has been discussed. (GitHub only allows to assign one milestone, so adding the 3.10 milestone will remove the Discussion milestone.)

cgetzen:
@felixfontein I'm curious what qualities would prevent this from going out in a minor release? It is backwards compatible.

felixfontein:
I think having it in a new minor release is fine, but I'd like to hear opinions from the other maintainers as well. This is potentially creating larger diffs for users, that could be a reason to wait for a new major release.

felixfontein:
Ping @getsops/maintainers

hiddeco (approved):

felixfontein:
@cgetzen can you please rebase or merge `main` into this so we can merge? Or check the checkbox that allows us maintainers to push commits, so we can push a merge commit for latest `main` ourselves? Thanks.

felixfontein (approved):

felixfontein:
@cgetzen thanks for your contribution!

PR #1576, opened 2024-08-05
keyservice: update protobuf to an recent release

Foxboron:
I'm unable to build this project with a recent version of protobuf and it seems like this hasn't been updated in a while.

In an effort to try make it buildable here is the needed patches.


Makefile
```diff
@@ -78,7 +78,7 @@ generate: keyservice/keyservice.pb.go
 	$(GO) generate
 
 %.pb.go: %.proto
-	protoc --go_out=plugins=grpc:. $<
+	protoc --go_grpc_opt=require_unimplemented_servers=false --go_grpc_out=. --go_out=. $<
```
felixfontein, line 81: I can't get this working with `go_grpc`, I have to use `go-grpc`:
```suggestion
	protoc --go-grpc_opt=require_unimplemented_servers=false --go-grpc_out=. --go_out=. $<
```
I installed it with `go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest` (and `go install google.golang.org/protobuf/cmd/protoc-gen-go@latest`).

(With the latest version the generated files also look differently.)

felixfontein (commented):
Thanks for your contribution! Sorry it took me so long to take a closer look...

Makefile
Foxboron: I'll take another stab at this then.

PR #1578, opened 2024-08-06
Support `GOOGLE_OAUTH_ACCESS_TOKEN` for Google Cloud Platform 

marensofier:
We are facing a situation where we want to use short lived access tokens with a minimum of permissions for the kms, and not use static long lived credentials ✨ 

That is why we really want to add support for the use of gcp access tokens in sops using the environment variable ```GOOGLE_OAUTH_ACCESS_TOKEN ``` (also  used by Pulumi/Terrafrom). 

Closes #1188.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -266,8 +266,12 @@ It is also possible to use ``updatekeys``, when adding or removing age recipient
   
 Encrypting using GCP KMS
 ~~~~~~~~~~~~~~~~~~~~~~~~
-GCP KMS uses `Application Default Credentials
-<https://developers.google.com/identity/protocols/application-default-credentials>`_.
+GCP KMS has support for authorization with the use of `Application Default Credentials
+<https://developers.google.com/identity/protocols/application-default-credentials>`_ and using an OAuth 2.0 token.
+Application default credentials precedes the use of access token.
+
+Using Application Default Credentials you can authorize by doing this:
+
 If you already logged in using
 
 .. code:: sh
@@ -279,6 +283,18 @@ you can enable application default credentials using the sdk:
 .. code:: sh
 
     $ gcloud auth application-default login
+
+Using OAauth tokens you can authorize by doing this:
+
+.. code:: sh
+    
+    $ export GOOGLE_OAUTH_ACCESS_TOKEN=<your access token>
+
+Or if you are logged in you can authorize by generating an access token:
+
+.. code:: sh
+
+    $ export GOOGLE_OAUTH_ACCESS_TOKEN="$(gcloud auth print-access-token)"
 
 Encrypting/decrypting with GCP KMS requires a KMS ResourceID. You can use the
 cloud console the get the ResourceID or you can create one using the gcloud
--- a/gcpkms/keysource.go
+++ b/gcpkms/keysource.go
@@ -24,6 +24,9 @@ const (
 	// a path to a credentials file, or directly as the variable's value in JSON
 	// format.
 	SopsGoogleCredentialsEnv = "GOOGLE_CREDENTIALS"
+	// SopsGoogleCredentialsOAuthTokenEnv is the environment variable used for the
+	// GCP OAuth 2.0 Token.
+	SopsGoogleCredentialsOAuthTokenEnv = "GOOGLE_OAUTH_ACCESS_TOKEN"
 	// KeyTypeIdentifier is the string used to identify a GCP KMS MasterKey.
 	KeyTypeIdentifier = "gcp_kms"
 )
@@ -245,12 +248,19 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
 	default:
 		credentials, err := getGoogleCredentials()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("credentials: failed to obtain credentials from %q: %w", SopsGoogleCredentialsEnv, err)
 		}
 		if credentials != nil {
 			opts = append(opts, option.WithCredentialsJSON(credentials))
+			break
+		}
+
+		if atCredentials := getGoogleOAuthTokenFromEnv(); atCredentials != nil {
+			opts = append(opts, option.WithTokenSource(atCredentials))
+			break
 		}
 	}
+
 	if key.grpcConn != nil {
 		opts = append(opts, option.WithGRPCConn(key.grpcConn))
 	}
@@ -266,8 +276,8 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
 
 // getGoogleCredentials returns the SopsGoogleCredentialsEnv variable, as
 // either the file contents of the path of a credentials file, or as value in
-// JSON format. It returns an error if the file cannot be read, and may return
-// a nil byte slice if no value is set.
+// JSON format.
+// It returns an error and a nil byte slice if the file cannot be read.
 func getGoogleCredentials() ([]byte, error) {
 	if defaultCredentials, ok := os.LookupEnv(SopsGoogleCredentialsEnv); ok && len(defaultCredentials) > 0 {
 		if _, err := os.Stat(defaultCredentials); err == nil {
@@ -276,4 +286,17 @@ func getGoogleCredentials() ([]byte, error) {
 		return []byte(defaultCredentials), nil
 	}
 	return nil, nil
+}
+
+// getGoogleOAuthTokenFromEnv returns the SopsGoogleCredentialsOauthTokenEnv variable,
+// as the OAauth 2.0 token.
+// It returns an error and a nil byte slice if the envrionment variable is not set.
+func getGoogleOAuthTokenFromEnv() oauth2.TokenSource {
+	if token, ok := os.LookupEnv(SopsGoogleCredentialsOAuthTokenEnv); ok && len(token) > 0 {
+		tokenSource := oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: token},
+		)
+		return tokenSource
+	}
+	return nil
 }
--- a/gcpkms/keysource_test.go
+++ b/gcpkms/keysource_test.go
@@ -61,8 +61,9 @@ func TestMasterKey_Encrypt(t *testing.T) {
 	})
 
 	key := MasterKey{
-		grpcConn:   newGRPCServer("0"),
-		ResourceID: testResourceID,
+		grpcConn:       newGRPCServer("0"),
+		ResourceID:     testResourceID,
+		credentialJSON: []byte("arbitrary credentials"),
 	}
 	err := key.Encrypt([]byte("encrypt"))
 	assert.NoError(t, err)
@@ -88,9 +89,10 @@ func TestMasterKey_Decrypt(t *testing.T) {
 		Plaintext: []byte(decryptedData),
 	})
 	key := MasterKey{
-		grpcConn:     newGRPCServer("0"),
-		ResourceID:   testResourceID,
-		EncryptedKey: "encryptedKey",
+		grpcConn:       newGRPCServer("0"),
+		ResourceID:     testResourceID,
+		EncryptedKey:   "encryptedKey",
+		credentialJSON: []byte("arbitrary credentials"),
 	}
 	data, err := key.Decrypt()
 	assert.NoError(t, err)
@@ -124,7 +126,7 @@ func TestMasterKey_ToMap(t *testing.T) {
 	}, key.ToMap())
 }
 
-func TestMasterKey_createCloudKMSService(t *testing.T) {
+func TestMasterKey_createCloudKMSService_withCredentialsFile(t *testing.T) {
 	tests := []struct {
 		key       MasterKey
 		errString string
@@ -144,6 +146,12 @@ func TestMasterKey_createCloudKMSService(t *testing.T) {
 		"type": "authorized_user"}`),
 			},
 		},
+		{
+			key: MasterKey{
+				ResourceID: testResourceID,
+			},
+			errString: `credentials: failed to obtain credentials from "SOPS_GOOGLE_CREDENTIALS"`,
+		},
 	}
 
 	for _, tt := range tests {
@@ -155,6 +163,29 @@ func TestMasterKey_createCloudKMSService(t *testing.T) {
 		}
 		assert.NoError(t, err)
 	}
+}
+
+func TestMasterKey_createCloudKMSService_withOauthToken(t *testing.T) {
+	t.Setenv(SopsGoogleCredentialsOAuthTokenEnv, "token")
+
+	masterKey := MasterKey{
+		ResourceID: testResourceID,
+	}
+
+	_, err := masterKey.newKMSClient()
+
+	assert.NoError(t, err)
+}
+
+func TestMasterKey_createCloudKMSService_withoutCredentials(t *testing.T) {
+	masterKey := MasterKey{
+		ResourceID: testResourceID,
+	}
+
+	_, err := masterKey.newKMSClient()
+
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "credentials: could not find default credentials")
 }
 
 func newGRPCServer(port string) *grpc.ClientConn {


gcpkms/keysource.go
```diff
@@ -253,3 +261,20 @@ func getGoogleCredentials() ([]byte, error) {
 	}
 	return nil, nil
 }
+
+func getGoogleAccessToken() (oauth2.TokenSource, error) {
```
devstein, line 265: can you add a go doc comment?

README.rst
```diff
@@ -235,6 +240,21 @@ you can enable application default credentials using the sdk:
...
 
+
+
+Using oauth tokens you can authorize by doing this:
+
+.. code:: sh
+
+    $ export CLOUDSDK_AUTH_ACCESS_TOKEN=<your access token>
+
+If you are already logged in
```
devstein, line 251: Nit:
```suggestion
Or if you are logged in you can authorize by generating an access token:
```

gcpkms/keysource.go
```diff
@@ -220,12 +222,18 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
-		if err != nil {
-			return nil, err
-		}
 		if credentials != nil {
 			opts = append(opts, option.WithCredentialsJSON(credentials))
 		}
+
+		at_credentials, at_err := getGoogleAccessToken()
+		if at_credentials != nil {
+			opts = append(opts, option.WithTokenSource(at_credentials))
```
devstein, line 231: Will it cause any issues if _both_ the default credentials and access token credentials are added to `opts`?

gcpkms/keysource.go
```diff
@@ -220,12 +222,18 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
+		}
+
+		if err != nil && at_err != nil {
+			return nil, err
+		}
```
devstein, line 236: We will never return `at_err`. Is there a way to surface both? Or the more relevant one?

gcpkms/keysource.go
```diff
@@ -220,12 +222,18 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 		credentials, err := getGoogleCredentials()
-		if err != nil {
-			return nil, err
-		}
 		if credentials != nil {
 			opts = append(opts, option.WithCredentialsJSON(credentials))
 		}
+
+		at_credentials, at_err := getGoogleAccessToken()
+		if at_credentials != nil {
```
devstein, line 230: Related to my other comment, given precedence to the default credentials is it even worth running this if we already found credentials?

devstein (changes requested):
@marensofier Thank you for your contribution! This has been a frequently requested feature. 

I left a few comments, but can you also please add test coverage?

devstein:
Related to https://github.com/getsops/sops/pull/1358

README.rst
```diff
@@ -235,6 +240,21 @@ you can enable application default credentials using the sdk:
...
+
+.. code:: sh
+
+    $ export CLOUDSDK_AUTH_ACCESS_TOKEN=<your access token>
+
+If you are already logged in 
+
+.. code:: sh
+
+    $ export CLOUDSDK_AUTH_ACCESS_TOKEN=$(gcloud auth print-access-token)
```
felixfontein, line 255:
```suggestion
    $ export CLOUDSDK_AUTH_ACCESS_TOKEN="$(gcloud auth print-access-token)"
```

gcpkms/keysource.go
```diff
@@ -22,7 +23,8 @@ const (
 	// SopsGoogleCredentialsEnv can be set as an environment variable as either
 	// a path to a credentials file, or directly as the variable's value in JSON
 	// format.
-	SopsGoogleCredentialsEnv = "GOOGLE_CREDENTIALS"
+	SopsGoogleCredentialsEnv         = "GOOGLE_CREDENTIALS"
+	SopsGoogleCredentialsAccessToken = "CLOUDSDK_AUTH_ACCESS_TOKEN"
```
felixfontein, line 27: I think `SopsGoogleCredentialsAccessToken` should also have some documentation.

README.rst
```diff
@@ -235,6 +240,21 @@ you can enable application default credentials using the sdk:
...
+
```
felixfontein, line 256: Shouldn't you mention that `CLOUDSDK_AUTH_ACCESS_TOKEN` can also point to a file?

Also, does the Google Cloud SDK also support this double usage of `CLOUDSDK_AUTH_ACCESS_TOKEN`? If not, why support it here (but don't document it)?

marensofier: Yes, of course, i agree - i will remove the possibility to add it as a reference to a file.

gcpkms/keysource.go
```diff
@@ -249,7 +261,23 @@ func getGoogleCredentials() ([]byte, error) {
 		if _, err := os.Stat(defaultCredentials); err == nil {
 			return os.ReadFile(defaultCredentials)
 		}
+
 		return []byte(defaultCredentials), nil
 	}
-	return nil, nil
+	return nil, fmt.Errorf("could not find Google credential file")
```
devstein, line 267: Can you update the function doc comment to reflect behavior change? Right now, it incorrectly reads `and may return a nil byte slice if no value is set.`

devstein (commented):
Looks good! Thanks for incorporating the feedback. One last minor change

gcpkms/keysource.go
marensofier: I also changed the description of the SopsGoogleCredentialsOAuthToken, where i forgot to update after removing the possibility of using a file path! 

Thank you for all of your comments, and time! 🌻

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/getsops/sops/v3
 
-go 1.22
+go 1.23.0
+
```
felixfontein, line 4: Can you undo this change? Thanks.

felixfontein (commented):
The PR now contains a lot of unnecessary files that should be removed. Also `--signoff` is missing (for DCO). Can you please fix this? Thanks!

sabre1041 (approved):
LGTM

sabre1041:
There is a fair amount of overlap with #1794 where there does need to be some assessment between the two enhancements

gcpkms/keysource.go
```diff
@@ -219,37 +223,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
-		if err != nil {
-			return nil, err
-		}
+		credentials, err_credentials_file := getGoogleCredentials()
 		if credentials != nil {
 			opts = append(opts, option.WithCredentialsJSON(credentials))
+			break
+		}
+
+		at_credentials, err_credentials_token := getGoogleOAuthToken()
```
matheuscscp, line 232:
```suggestion
		atCredentials, errCredentialsToken := getGoogleOAuthToken()
```

gcpkms/keysource.go
```diff
@@ -219,37 +223,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 	if key.grpcConn != nil {
 		opts = append(opts, option.WithGRPCConn(key.grpcConn))
 	}
 
 	ctx := context.Background()
-	client, err := kms.NewKeyManagementClient(ctx, opts...)
-	if err != nil {
-		return nil, err
+	client, err_credentials := kms.NewKeyManagementClient(ctx, opts...)
+	if err_credentials != nil {
```
matheuscscp, line 248:
```suggestion
	if errCredentials != nil {
```

gcpkms/keysource.go
```diff
@@ -219,37 +223,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
+		return nil, err_credentials
```
matheuscscp, line 249:
```suggestion
		return nil, errCredentials
```

gcpkms/keysource.go
```diff
@@ -23,6 +24,9 @@ const (
 	// a path to a credentials file, or directly as the variable's value in JSON
 	// format.
 	SopsGoogleCredentialsEnv = "GOOGLE_CREDENTIALS"
+	// SopsGoogleCredentialsOAuthToken is the environment variable used for the
+	// GCP Oauth2 Token.
+	SopsGoogleCredentialsOAuthToken = "GOOGLE_OAUTH_ACCESS_TOKEN"
```
matheuscscp, line 29:
```suggestion
	// SopsGoogleCredentialsOAuthTokenEnv is the environment variable used for the
	// GCP Oauth2 Token.
	SopsGoogleCredentialsOAuthTokenEnv = "GOOGLE_OAUTH_ACCESS_TOKEN"
```

gcpkms/keysource.go
```diff
@@ -203,8 +207,8 @@ func (key *MasterKey) TypeToIdentifier() string {
 	return KeyTypeIdentifier
 }
 
-// newKMSClient returns a GCP KMS client configured with the credentialJSON
-// and/or grpcConn, falling back to environmental defaults.
+// newKMSClient returns a GCP KMS client configured with the credentialJSON,
+// tokenSource and/or grpcConn, falling back to environmental defaults.
```
matheuscscp, line 211: I don't think the GCP authentication method introduced in this PR is quite what you would call "token source". A token source method would be similar to the equivalents of AWS and Azure:


https://github.com/getsops/sops/blob/5355c249aa8ea413bea6b2ea20a029cb21048e99/kms/keysource.go#L72-L76

https://github.com/getsops/sops/blob/5355c249aa8ea413bea6b2ea20a029cb21048e99/azkv/keysource.go#L58-L62

gcpkms/keysource.go
```diff
@@ -219,37 +223,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 			return os.ReadFile(defaultCredentials)
 		}
+
 		return []byte(defaultCredentials), nil
 	}
-	return nil, nil
+	return nil, fmt.Errorf("could not find Google credential file")
+}
+
+// getGoogleOAuthToken returns the SopsGoogleCredentialsOauthToken variable,
```
matheuscscp, line 271:
```suggestion
// getGoogleOAuthToken returns the SopsGoogleCredentialsOauthTokenEnv variable,
```

matheuscscp:
> There is a fair amount of overlap with #1794 where there does need to be some assessment between the two enhancements

Hey @sabre1041 I disagree, there's no overlap here. The two PRs are introducing two different methods of authentication and can be merged independently. This one just needs to rename the method to emphasize the fact that it retrieves the token from an environment variable, and not from a generic `oauth2.TokenSource` passed in memory like the other one.

In the end the switch-case from `newKMSClient()` should look like this:

```go
	switch {
	case key.tokenSource != nil:
		opts = append(opts, option.WithTokenSource(key.tokenSource))
	case key.credentialJSON != nil:
		opts = append(opts, option.WithCredentialsJSON(key.credentialJSON))
	default:
		credentials, errCredentialsFile := getGoogleCredentials()
		if credentials != nil {
			opts = append(opts, option.WithCredentialsJSON(credentials))
			break
		}

		atCredentials, errCredentialsToken := getGoogleOAuthTokenFromEnv()
		if atCredentials != nil {
			opts = append(opts, option.WithTokenSource(atCredentials))
		}

		if errCredentialsFile != nil && errCredentialsToken != nil {
			return nil, fmt.Errorf("credentials: failed to get credentials for gcp kms, add default credentials or oauth access token from env")
		}
	}
```

felixfontein:
While going through all PRs I noticed that #1188 seems to do the same thing (from the general functionality point of view) as this one. So once this is done I guess we can close #1188. (If that's not correct, please tell me :) )

felixfontein:
@marensofier can you please resolve the conflicts and address the comments? Thanks.

gcpkms/keysource.go
felixfontein:
```suggestion
		atCredentials, errCredentialsToken := getGoogleOAuthTokenFromEnv()
```

go.mod
felixfontein:
```suggestion
```

gcpkms/keysource_test.go
```diff
@@ -157,6 +165,29 @@ func TestMasterKey_createCloudKMSService(t *testing.T) {
 	}
 }
 
+func TestMasterKey_createCloudKMSService_withOauthToken(t *testing.T) {
+	t.Setenv(SopsGoogleCredentialsOAuthToken, "token")
```
felixfontein, line 169:
```suggestion
	t.Setenv(SopsGoogleCredentialsOAuthTokenEnv, "token")
```

README.rst
```diff
@@ -266,8 +266,12 @@ It is also possible to use ``updatekeys``, when adding or removing age recipient
   
 Encrypting using GCP KMS
 ~~~~~~~~~~~~~~~~~~~~~~~~
-GCP KMS uses `Application Default Credentials
-<https://developers.google.com/identity/protocols/application-default-credentials>`_.
+GCP KMS has support for authorization with the use of `Application Default Credentials
+<https://developers.google.com/identity/protocols/application-default-credentials>`_ and using an oauth2 token.
```
felixfontein, line 270:
```suggestion
<https://developers.google.com/identity/protocols/application-default-credentials>`_ and using an OAuth 2.0 token.
```

README.rst
```diff
@@ -280,6 +284,19 @@ you can enable application default credentials using the sdk:
 
     $ gcloud auth application-default login
 
+Using oauth tokens you can authorize by doing this:
```
felixfontein, line 287:
```suggestion
Using OAauth tokens you can authorize by doing this:
```

gcpkms/keysource.go
```diff
@@ -24,6 +24,9 @@ const (
 	// a path to a credentials file, or directly as the variable's value in JSON
 	// format.
 	SopsGoogleCredentialsEnv = "GOOGLE_CREDENTIALS"
+	// SopsGoogleCredentialsOAuthTokenEnv is the environment variable used for the
+	// GCP Oauth2 Token.
```
felixfontein, line 28:
```suggestion
	// GCP OAuth 2.0 Token.
```

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 		}
+
 		return []byte(defaultCredentials), nil
 	}
-	return nil, nil
+	return nil, fmt.Errorf("could not find Google credential file")
+}
+
+// getGoogleOAuthTokenFromEnv returns the SopsGoogleCredentialsOauthTokenEnv variable,
+// as the oauth token.
```
felixfontein, line 295:
```suggestion
// as the OAauth 2.0 token.
```

felixfontein:
@sabre1041 @matheuscscp can you take another look at the latest version?

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
 	case key.credentialJSON != nil:
 		opts = append(opts, option.WithCredentialsJSON(key.credentialJSON))
 	default:
-		credentials, err := getGoogleCredentials()
-		if err != nil {
-			return nil, err
-		}
+		credentials, errCredentialsFile := getGoogleCredentials()
 		if credentials != nil {
 			opts = append(opts, option.WithCredentialsJSON(credentials))
```
matheuscscp, line 251: I think here I'd check both if creds is not nil and the error is nil, same below

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 		opts = append(opts, option.WithGRPCConn(key.grpcConn))
 	}
 
 	ctx := context.Background()
-	client, err := kms.NewKeyManagementClient(ctx, opts...)
-	if err != nil {
-		return nil, err
+	client, errCredentials := kms.NewKeyManagementClient(ctx, opts...)
+	if errCredentials != nil {
+		return nil, errCredentials
```
matheuscscp, line 272: I don't see why rename this error variable here

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
+			break
+		}
+
+		atCredentials, errCredentialsToken := getGoogleOAuthTokenFromEnv()
+		if atCredentials != nil {
+			opts = append(opts, option.WithTokenSource(atCredentials))
+		}
+
+		if errCredentialsFile != nil && errCredentialsToken != nil {
+			return nil, fmt.Errorf("credentials: failed to get credentials for gcp kms, add default credentials or oauth access token from env")
```
hiddeco, line 261:
```suggestion
			return nil, fmt.Errorf("credentials: none found: missing %q and/or %q environment variables", SopsGoogleCredentialsEnv, SopsGoogleCredentialsOAuthTokenEnv)
```

makes it actionable.

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 	}
-	return nil, nil
+	return nil, fmt.Errorf("could not find Google credential file")
+}
+
+// getGoogleOAuthTokenFromEnv returns the SopsGoogleCredentialsOauthTokenEnv variable,
+// as the OAauth 2.0 token.
+// It returns an error and a nil byte slice if the envrionment variable is not set.
+func getGoogleOAuthTokenFromEnv() (oauth2.TokenSource, error) {
+	if token, isSet := os.LookupEnv(SopsGoogleCredentialsOAuthTokenEnv); isSet {
```
hiddeco, line 298:
```suggestion
	if token, isSet := os.LookupEnv(SopsGoogleCredentialsOAuthTokenEnv); isSet && len(token) > 0 {
```

matheuscscp:
I'd also change the name of the PR to `Add support for gcp access token from environment variable`

gcpkms/keysource.go
hiddeco: Also: checking the errors is theoretically less fool proof than checking if `credentials` and `atCredentials` are both `nil`, because that's the condition to actually have them configured.

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
 	case key.credentialJSON != nil:
 		opts = append(opts, option.WithCredentialsJSON(key.credentialJSON))
 	default:
-		credentials, err := getGoogleCredentials()
-		if err != nil {
-			return nil, err
-		}
+		credentials, errCredentialsFile := getGoogleCredentials()
 		if credentials != nil {
```
felixfontein, line 250:
```suggestion
		if credentials != nil && errCredentialsFile == nil {
```

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 			opts = append(opts, option.WithCredentialsJSON(credentials))
+			break
+		}
+
+		atCredentials, errCredentialsToken := getGoogleOAuthTokenFromEnv()
+		if atCredentials != nil {
```
felixfontein, line 256:
```suggestion
		if atCredentials != nil && errCredentialsToken == nil {
```

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
+			opts = append(opts, option.WithTokenSource(atCredentials))
```
felixfontein, line 257:
```suggestion
			opts = append(opts, option.WithTokenSource(atCredentials))
			break
```

gcpkms/keysource.go
felixfontein: With the other changes suggested, I'd remove the `if` condition here completely.

gcpkms/keysource.go
felixfontein, line 272:
```suggestion
	client, err := kms.NewKeyManagementClient(ctx, opts...)
	if err != nil {
		return nil, err
```

gcpkms/keysource.go
```diff
@@ -243,37 +246,62 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
+		}
+
+		if errCredentialsFile != nil && errCredentialsToken != nil {
+			return nil, fmt.Errorf("credentials: failed to get credentials for gcp kms, add default credentials or oauth access token from env")
 		}
```
felixfontein, line 262:
```suggestion
		return nil, fmt.Errorf("credentials: none found: missing %q and/or %q environment variables", SopsGoogleCredentialsEnv, SopsGoogleCredentialsOAuthTokenEnv)
```

gcpkms/keysource.go
felixfontein: Not `nil`, or `len(...) > 0`?

gcpkms/keysource.go
felixfontein: PTAL at the set of my suggestions that are still open.

gcpkms/keysource.go
hiddeco: I think the `break` makes the code hard to follow and I would prefer simply keeping the `if` check below to return the error.

felixfontein: I find it personally very confusing that the first `if` has a `break` and the second does not.

Also the condition for the final `if` gets rather involved if it mirrors the conditions of the two `if`s, which are no longer just `x == nil`, but now `y != nil && x == nil`.

I find the code a lot easier to understand if the conditions are not repeated all over the place in different forms, and you have to spend a lot of time thinking on what the final `if` condition actually means in relation to the previous two `if` conditions.

felixfontein (approved):
LGTM as far as I can tell.

gcpkms/keysource.go
```diff
@@ -245,12 +248,21 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
...
 			opts = append(opts, option.WithCredentialsJSON(credentials))
+			break
 		}
+
+		if atCredentials := getGoogleOAuthTokenFromEnv(); atCredentials != nil {
+			opts = append(opts, option.WithTokenSource(atCredentials))
+			break
+		}
+
+		return nil, fmt.Errorf("credentials: none found: missing %q or %q environment variable", SopsGoogleCredentialsEnv, SopsGoogleCredentialsOAuthTokenEnv)
```
felixfontein, line 263: Hmm, the old behavior was to not fail if no credentials have been provided in the `default` branch. (Whether that makes sense - no idea. Maybe the GCP SDK does something else in that case to get hold of credentials?) If we want to restore this:
```suggestion
		}
```

gcpkms/keysource_test.go
```diff
@@ -157,6 +165,29 @@ func TestMasterKey_createCloudKMSService(t *testing.T) {
...
+
+func TestMasterKey_createCloudKMSService_withoutCredentials(t *testing.T) {
+	masterKey := MasterKey{
+		ResourceID: testResourceID,
+	}
+
+	_, err := masterKey.newKMSClient()
+
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "credentials: failed to get credentials")
```
felixfontein, line 188: In any case, this test needs adjustments.

gcpkms/keysource.go
matheuscscp: > Maybe the GCP SDK does something else in that case to get hold of credentials?

Yes, it does. All the google libraries default to [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials), which is essentially searching for credentials in the environment:

* The `GOOGLE_APPLICATION_CREDENTIALS` standard environment variable (btw none of the other env vars here are this one)
* A file created in a well-known location by the command `gcloud auth application-default login`
* The GCE/GKE metadata server

However, `getGoogleCredentials()` (which should probably be called `getGoogleCredentialsFromEnv()`), only returns an error if the non-standard `GOOGLE_CREDENTIALS` environment variable is set, and this variable is only set when the user is declaring a clear intention of using it, so erroring out the entire operation makes sense to me. If the code ends up using an unintended authentication method that is coincidentally available, that's pretty bad in my opinion. People need to organize their configuration well, this new version of the code should help them out on that. Disruptive? Yes, probably, but for a greater good in my opinion.

gcpkms/keysource.go
```diff
@@ -277,3 +287,16 @@ func getGoogleCredentials() ([]byte, error) {
 	}
 	return nil, nil
 }
+
+// getGoogleOAuthTokenFromEnv returns the SopsGoogleCredentialsOauthTokenEnv variable,
+// as the OAauth 2.0 token.
+// It returns an error and a nil byte slice if the envrionment variable is not set.
+func getGoogleOAuthTokenFromEnv() oauth2.TokenSource {
+	if token, isSet := os.LookupEnv(SopsGoogleCredentialsOAuthTokenEnv); isSet && len(token) > 0 {
```
matheuscscp, line 295:
```suggestion
	if token, ok := os.LookupEnv(SopsGoogleCredentialsOAuthTokenEnv); ok && len(token) > 0 {
```

gcpkms/keysource.go
felixfontein: > Disruptive? Yes, probably, but for a greater good in my opinion.

It is still a breaking change, I'd prefer not to introduce these in a minor release.

matheuscscp: Oh nevermind, for some reason I wrapped my head around the error returned from `getGoogleCredentials()`, and not this one on the bottom. Yes, this comments makes a lot of sense!

hiddeco (approved):

matheuscscp (approved):

felixfontein (approved):

felixfontein:
@marensofier @warwick-mitchell1 @matheuscscp and everyone else involved in the PRs, thanks a lot for your contributions, reviews, and comments! I'm glad we finally got this resolved and merged :)

PR #1595, opened 2024-09-04
tests: ignore env var SOPS_AGE_KEY_FILE

duthils:
Why:

* setting the variable SOPS_AGE_KEY_FILE will make tests fail

I also looked at other env variables, but only `SOPS_GPG_EXEC` would cause an issue. I guess keeping this variable could be useful to run tests in a non-standard environment, so I did not include it in the variables to unset.

Fixes https://github.com/getsops/sops/issues/1410


felixfontein (approved):

felixfontein:
@duthils thanks!

PR #1598, opened 2024-09-07
Add `--recursive` flag

qiu-x:
Adds a `--recursive` flag like described in https://github.com/getsops/sops/issues/1204.


qiu-x:
@schlichtanders please let me know if this works as expected for you

felixfontein:
Thanks for your contribution! I don't think this is a good idea though. We don't want to add new features to the `--encrypt` / `--decrypt` / ... flags, but instead concentrate on the `encrypt`, `decrypt`, `edit`, ... subcommands (see also #1333).

I'm also not sure whether it's a good idea to add this to the `encrypt` and `decrypt` subcommands, it might be better to have separate subcommands for this, since the is sufficiently different from regular encryption/decryption (you don't need a specific filename, but rather an optional path; you always encrypt/decrypt in-place; etc.).

qiu-x:
> Thanks for your contribution! I don't think this is a good idea though. We don't want to add new features to the --encrypt / --decrypt / ... flags, but instead concentrate on the encrypt, decrypt, edit, ... subcommands (see also https://github.com/getsops/sops/issues/1333).

I see, I have been wondering why there are two ways of executing actions, but #1333 explains everything.

> I'm also not sure whether it's a good idea to add this to the encrypt and decrypt subcommands, it might be better to have separate subcommands for this, since the is sufficiently different from regular encryption/decryption (you don't need a specific filename, but rather an optional path; you always encrypt/decrypt in-place; etc.).

Personally, I am more in favor of a `--recursive` flag, because that seems analogous to what other CLI tools do. Currently, this change also works modes like `--rotate` or `--unset`, so adding subcommands instead would lead to a lot of duplication from my perspective.

PR #1604, opened 2024-09-10
GitHub Actions linter for Rust formatting

duthils:
This PR takes the changes from https://github.com/getsops/sops/pull/1403 and applies them in a dedicated CI workflow.

Fixes https://github.com/getsops/sops/pull/1390

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -1,0 +1,27 @@
+name: Linters
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    # Only run when linted files change
+    paths:
+      - 'functional-tests/**/*.rs'
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    name: Lint Rust source files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Run Formatting Check
+        run: cargo fmt --check
+        working-directory: ./functional-tests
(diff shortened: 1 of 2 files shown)


.github/workflows/linters.yml
```diff
@@ -0,0 +1,27 @@
...
+  contents: read
+
+jobs:
+  lint:
+    name: Lint Rust source files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
```
felixfontein, line 24: I think we should also install rust here. I've copied this from `.github/workflows/cli.yml`:
```suggestion

      - name: Install rustup
        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y --default-toolchain 1.70.0

```

felixfontein (commented):
Thanks for your contribution!

duthils:
Rust is already available on [Github runners in version 1.80.1](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#rust-tools), but I guess we want to keep the same version as is used for running the tests.

Thanks for the suggestion!

felixfontein:
> Rust is already available on [Github runners in version 1.80.1](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#rust-tools), but I guess we want to keep the same version as is used for running the tests.

Yep; also having a controlled environment is generally a good idea :) Then we can decide which Rust version to use, instead of having to use the version provided by the GHA image.

felixfontein (approved):

felixfontein:
@duthils @shikharvashistha thanks for your contribution!

PR #1608, opened 2024-09-11
updatekeys subcommand: actually use option `--shamir-secret-sharing-threshold`

felixfontein:
The option was renamed while the `updatekeys` subcommand was developed, and the name was not updated in the `updatekeys` PR. See the following for more information: https://github.com/getsops/sops/pull/1509#discussion_r1623288498

PR #1609, opened 2024-09-11
updatekeys subcommand: show changes in shamir_threshold

felixfontein:
Inspired by #1509. Also fixes that changes in the Shamir threshold are detected (when the command line flag is used, see #1608).

This does not change the behavior which Shamir threshold is actually used (see the `TODO` comment), that raised the original change in #1509. I'm currently thinking that it should be the user's choice and that should be added as a feature, not a bugfix.

### DIFF
--- a/cmd/sops/subcommand/updatekeys/updatekeys.go
+++ b/cmd/sops/subcommand/updatekeys/updatekeys.go
@@ -66,11 +66,22 @@ func updateFile(opts Opts) error {
 			keysWillChange = true
 		}
 	}
-	if !keysWillChange {
+
+	// TODO: use conf.ShamirThreshold instead of tree.Metadata.ShamirThreshold in the next line?
+	//       Or make this configurable?
+	var shamirThreshold = tree.Metadata.ShamirThreshold
+	if opts.GroupQuorum != 0 {
+		shamirThreshold = opts.GroupQuorum
+	}
+	shamirThreshold = min(shamirThreshold, len(conf.KeyGroups))
+	var shamirThresholdWillChange = tree.Metadata.ShamirThreshold != shamirThreshold
+
+	if !keysWillChange && !shamirThresholdWillChange {
 		log.Printf("File %s already up to date", opts.InputPath)
 		return nil
 	}
 	fmt.Printf("The following changes will be made to the file's groups:\n")
+	common.PrettyPrintShamirDiff(tree.Metadata.ShamirThreshold, shamirThreshold)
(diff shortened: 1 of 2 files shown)


cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -66,11 +66,22 @@ func updateFile(opts Opts) error {
 			keysWillChange = true
 		}
 	}
-	if !keysWillChange {
+
+	// TODO: use conf.ShamirThreshold instead of tree.Metadata.ShamirThreshold in the next line?
+	//       Or make this configurable?
+	var shamirThreshold = tree.Metadata.ShamirThreshold
+	if opts.GroupQuorum != 0 {
```
jorn-ola-birkeland, line 73: The term `GroupQuorum` is only used here. The term `GroupThreshold` is used in options in the `groups add` and `groups delete `commands, while `ShamirThreshold` is used elsewhere. It is not terribly important, but renaming the field in the `Opts` struct would improve readability imho.

felixfontein: I agree, but that should happen in a feature (or even major) release, not in a bugfix release. While this is not part of the offically declared public interface, we still try to avoid breaking changes there.

jorn-ola-birkeland, line 73: A problem in [#1509](https://github.com/getsops/sops/pull/1509), was that the argument `shamir-secret-sharing-quorum` was not allowed for the command, and therefore the `Opts.GroupQuorum`was always 0 (zero). In this branch, the issue does not seem to have been resolved (and if fixed, it should probably be renamed, e.g to `shamir-secret-sharing-threshold`).

felixfontein: That has already been fixed in #1608. This branch is only for showing the changes, not for anything else.

jorn-ola-birkeland: > I agree, but that should happen in a feature (or even major) release, not in a bugfix release. While this is not part of the offically declared public interface, we still try to avoid breaking changes there.

Excellent point 👍

jorn-ola-birkeland: Ah, beautiful! 😃 Did not spot that one. I'll 🤐

hiddeco (approved):

felixfontein:
@jorn-ola-birkeland @hiddeco thanks for reviewing!

PR #1611, opened 2024-09-11
Use SOPS_EDITOR before EDITOR

### LINKED ISSUE #1605: Special `EDITOR` variable for sops
A special `EDITOR` variable for sops would be nice to have (e.g. `SOPS_EDITOR` following the `GIT_EDITOR`, `SUDO_EDITOR`, etc. convention). My use case is for better integration with GUI editors.

For example, I have my `EDITOR` set to `code` because I don't want to block my terminal when opening files in VS Code. But for `GIT_EDITOR`, I set it to `code -rw` because I want it to block so that git waits until I'm done writing my commit message.

felixfontein:
Fixes #1605.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -106,7 +106,8 @@ encryption/decryption transparently and open the cleartext file in an editor
     please wait while an encryption key is being generated and stored in a secure fashion
     file written to mynewtestfile.yaml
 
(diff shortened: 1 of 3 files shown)


onedr0p (approved):

felixfontein:

PR #1612, opened 2024-09-13
github actions: set rust version globally

duthils:
Following https://github.com/getsops/sops/pull/1604, we now have two different GH actions requiring a version of Rust.

Rustup allows setting the toolchain version in a file `rust-toolchain.toml`, allowing to set the Rust version once for all Rust commands.


felixfontein:
@duthils thanks for improving CI!

Issue #1572, opened 2024-08-02
CVE-2024-24791 in sops binary v3.9.0


oschni:
Our scanning jobs have identified a new CVE "[CVE-2024-24791](https://www.cve.org/CVERecord?id=CVE-2024-24791)" in the sops binary v3.9.0. This is an issue with the Go standard library net/http.

Is it possible to rebuilt sops binaries once the vulnerability has been fixed?

duthils:
This CVE is fixed in go 1.22.5, see the [release announcement](https://groups.google.com/g/golang-announce/c/gyb7aM1C9H4).

The go toolchain was updated to 1.22.5 in https://github.com/getsops/sops/pull/1589

felixfontein:
I know almost nothing about how goreleaser works, but the release workflow still runs with Go 1.21.x. I created #1615 to change that. (PR to update Go version in CI: #1531.)

duthils:
Oh, right... For the record, the CVE issue is also fixed in 1.21.12, see the [release announcement](https://groups.google.com/g/golang-announce/c/gyb7aM1C9H4).

reneleonhardt:
FYI the latest security fixes have not been backported anymore, so 1.21.13 shoudn't be used anymore:
https://go.dev/doc/devel/release#go1.22.7

felixfontein:
This was fixed in 3.9.1.

PR #1531, opened 2024-06-16
feat(build): Build last 2 Go versions

reneleonhardt:
### Features
* CI builds both latest Go versions (1.21 and 1.22)

PR #1615, opened 2024-09-14
Build release with Go 1.22.x

felixfontein:
The Go version for CI is upgraded in #1531, but for the release process we're still using 1.21.x.

Ref: #1572

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v4.0.1
         with:
-          go-version: 1.22.x
+          go-version-file: go.mod
           cache: false
 
       - name: Setup Syft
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v4.0.1
         with:
-          go-version: 1.21.x
+          go-version: 1.22.x
```
duthils, line 35: The `setup-go` action allows setting the Go version [directly from the `go.mod` file](https://github.com/actions/setup-go/tree/v4.0.1?tab=readme-ov-file#getting-go-version-from-the-gomod-file). It would avoid duplication and future desynchronization of the Go version:
```suggestion
          go-version-file: go.mod
```

felixfontein: That would be even better! @getsops/maintainers what do you think? Should we use `go.mod` as our source of truth?

onedr0p (approved):

PR #1617, opened 2024-09-16
unit tests: add coverage for unset

duthils:


sops_test.go
```diff
@@ -1108,6 +1108,322 @@ func TestSetArrayNonLeaf(t *testing.T) {
...
+					Value: "bar",
+				},
+			},
+		},
+	}
+	unset, err := branch.Unset([]interface{}{"foo", "unknown"})
+	assert.Equal(t, err.(*SopsKeyNotFound).Key, "unknown")
+	if unset != nil {
+		t.Errorf("Unset result was not nil upon %s", err)
+	}
```
felixfontein, line 1368: I think you can also simply do
```suggestion
	assert.Nil(t, unset, "Unset result was not nil upon %s", err)
```
(https://github.com/stretchr/testify/blob/b074924938f86d417f1c9a845c7e8b0784d7f937/assert/assertions.go#L704-L715)

Same below.

sops_test.go
```diff
@@ -1108,6 +1108,322 @@ func TestSetArrayNonLeaf(t *testing.T) {
...
+					TreeItem{
+						Key: "bar",
+						Value: "bar",
+					},
+				},
+			},
+		},
+	}
+	unset, err := branch.Unset([]interface{}{"foo", 0})
+	assert.Equal(t, nil, err)
```
felixfontein, line 1342:
```suggestion
	assert.Nil(t, err)
```

felixfontein (commented):
Thanks for improving the tests!

sops_test.go
duthils: Thanks for the tip!

sops_test.go
duthils: Thanks for the tip!

felixfontein (approved):

felixfontein:
@duthils thanks for your contribution!

PR #1621, opened 2024-09-17
Add 'check' task to CI that makes sure everything is green

felixfontein:
This allows to simply the branch protection steps, which in turn makes it easier to upgrade CI without having to adjust the branch protection conditions every time.

Ref: https://github.com/getsops/sops/pull/1531#pullrequestreview-2304547983
Ref: https://github.com/re-actors/alls-green?tab=readme-ov-file#alls-green

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -122,3 +122,22 @@ jobs:
       - name: Run tests
         run: cargo test
         working-directory: ./functional-tests
+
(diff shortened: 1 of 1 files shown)


hiddeco (approved):

felixfontein:
@hiddeco thanks for reviewing! Can you adjust the branch protections to only check for 'check'?

PR #1625, opened 2024-09-19
refactor: specify missing file in error message

t3hmrman:
This commit updates the error output by `sops` when failing to operate on a msising file to include the path to the file that was missing.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -728,7 +728,7 @@ func main() {
 					return toExitError(err)
 				}
 				if _, err := os.Stat(fileName); os.IsNotExist(err) {
-					return common.NewExitError("Error: cannot operate on non-existent file", codes.NoFileSpecified)
+					return common.NewExitError(fmt.Sprintf("Error: cannot operate on non-existent file %q", fileName), codes.NoFileSpecified)
 				}
 				fileNameOverride := c.String("filename-override")
 				if fileNameOverride == "" {
@@ -892,7 +892,7 @@ func main() {
 					return toExitError(err)
 				}
 				if _, err := os.Stat(fileName); os.IsNotExist(err) {
-					return common.NewExitError("Error: cannot operate on non-existent file", codes.NoFileSpecified)
+					return common.NewExitError(fmt.Sprintf("Error: cannot operate on non-existent file %q", fileName), codes.NoFileSpecified)
 				}
 				fileNameOverride := c.String("filename-override")
 				if fileNameOverride == "" {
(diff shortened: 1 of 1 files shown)


cmd/sops/main.go
```diff
@@ -1203,7 +1203,7 @@ func main() {
 					return toExitError(err)
 				}
 				if _, err := os.Stat(fileName); os.IsNotExist(err) {
-					return common.NewExitError("Error: cannot operate on non-existent file", codes.NoFileSpecified)
+					return common.NewExitError(fmt.Sprintf("Error: cannot operate on non-existent file [%s]", fileName), codes.NoFileSpecified)
```
felixfontein, line 1206: Why use `[]` to wrap the filename, instead of quotes?

felixfontein (commented):
Thanks for your contribution! I'm wondering why you are only picking this specific error, instead of the very similar errors in lines 731, 895, 1053, 1655, and 1558 as well?

t3hmrman:
Thanks for taking a look @felixfontein ! There was no particular reason I skipped those other lines -- I personally was looking to solve the failure case I ran into, I'm not sure if the other lines would be quite as useful to know the file (though in general, it's useful everywhere!)

I can go through and update the rest of the places as well.

cmd/sops/main.go
t3hmrman: That's just a personal preference, things in `[]`s are usually easy to double-click and highlight in a terminal -- but I could definitely use quotes, is that preferred?

felixfontein: Good question... I grepped a bit, and it looks like the common way is to use no delimiter, and separate it from the previous text by a colon, i.e. like this:
```suggestion
					return common.NewExitError(fmt.Sprintf("Error: cannot operate on non-existent file: %s", fileName), codes.NoFileSpecified)
```

t3hmrman: Ahhh, so that I think is a different pattern, I saw that and it seems to be more for when there's an internal error that should be surfaced. 

Since the file being non-existent *is* the "terminal" error, I put the file name directly in the message -- augmenting it rather than wrapping another error's message, essentially.

felixfontein: I think in that case using `%q` instead of manually quoting is preferable.

t3hmrman: Sure -- will update to use `%q` everywhere `[%s]` was

t3hmrman:
Hey @felixfontein thanks for the additional review, I've just rebased and *should* have everything addressed!

felixfontein (approved):

felixfontein:
@t3hmrman thanks for your contribution!

t3hmrman:
Thanks for the reviews -- glad to be able to contribute to `sops`!

PR #1632, opened 2024-09-25
Dependabot: fix config for Docker, add config for Rust/Cargo

felixfontein:
This was originally part of #1530 until it got removed from that PR.

PR #1637, opened 2024-09-26
Add Cargo.lock for better reproducability of functional tests

felixfontein:
Right now `cargo test` always uses the latest versions, or the versions from the last time you ran the tests (picked from your local `Cargo.lock` file). So different users might use different versions of dependencies without noticing. Adding `Cargo.lock` to the repository avoids this. Since we have Dependabot for Cargo/Rust (#1632), the lock file will get updated regularly in PRs, and in there we also notice in case the updated dependencies break something.


hiddeco (approved):

felixfontein:

PR #1641, opened 2024-10-02
feat: add age plugin support

brianmcgee:
Another attempt at #1465 without bringing in so much code from [age](https://github.com/FiloSottile/age).

Instead, I created https://github.com/FiloSottile/age/pull/591 upstream to expose `PluginTerminalUI`.

## TODO
~~- [ ] update `go.mod` once https://github.com/FiloSottile/age/pull/591 is merged~~

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -1,8 +1,10 @@
 package age
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
+	"filippo.io/age/plugin"
 	"fmt"
 	"io"
 	"os"
@@ -115,7 +117,10 @@ type ParsedIdentities []age.Identity
 // parsing (using age.ParseIdentities) and appending to the slice yourself, in
 // combination with e.g. a sync.Mutex.
 func (i *ParsedIdentities) Import(identity ...string) error {
-	identities, err := parseIdentities(identity...)
+	// one identity per line
+	r := strings.NewReader(strings.Join(identity, "\n"))
+
+	identities, err := parseIdentities(r)
 	if err != nil {
 		return fmt.Errorf("failed to parse and add to age identities: %w", err)
 	}
@@ -339,6 +344,12 @@ func (key *MasterKey) loadIdentities() (ParsedIdentities, error) {
 // key or a public ssh key.
 func parseRecipient(recipient string) (age.Recipient, error) {
 	switch {
+	case strings.HasPrefix(recipient, "age1") && strings.Count(recipient, "1") > 1:
+		parsedRecipient, err := plugin.NewRecipient(recipient, pluginTerminalUI)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse input as age key from age plugin: %w", err)
+		}
+		return parsedRecipient, nil
 	case strings.HasPrefix(recipient, "age1"):
 		parsedRecipient, err := age.ParseX25519Recipient(recipient)
 		if err != nil {
@@ -357,17 +368,39 @@ func parseRecipient(recipient string) (age.Recipient, error) {
 	return nil, fmt.Errorf("failed to parse input, unknown recipient type: %q", recipient)
 }
 
-// parseIdentities attempts to parse the string set of encoded age identities.
-// A single identity argument is allowed to be a multiline string containing
-// multiple identities. Empty lines and lines starting with "#" are ignored.
-func parseIdentities(identity ...string) (ParsedIdentities, error) {
-	var identities []age.Identity
-	for _, i := range identity {
-		parsed, err := age.ParseIdentities(strings.NewReader(i))
+// parseIdentities attempts to parse one or more age identities from the provided reader.
+// One identity per line.
+// Empty lines and lines starting with "#" are ignored.
+func parseIdentities(r io.Reader) (ParsedIdentities, error) {
+	var identities ParsedIdentities
+
+	scanner := bufio.NewScanner(r)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		parsed, err := parseIdentity(line)
 		if err != nil {
 			return nil, err
 		}
-		identities = append(identities, parsed...)
+
+		identities = append(identities, parsed)
 	}
+
 	return identities, nil
+}
+
+func parseIdentity(s string) (age.Identity, error) {
+	switch {
+	case strings.HasPrefix(s, "AGE-PLUGIN-"):
+		return plugin.NewIdentity(s, pluginTerminalUI)
+	case strings.HasPrefix(s, "AGE-SECRET-KEY-1"):
+		return age.ParseX25519Identity(s)
+	default:
+		return nil, fmt.Errorf("unknown identity type")
+	}
 }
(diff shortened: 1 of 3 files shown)


felixfontein:
Thank you very much for this! I've marked it as a draft (we can put it back to ready once the upstream change is merged).

go.mod
```diff
@@ -134,3 +134,5 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20240903143218-8af14fe29dc1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace filippo.io/age => github.com/brianmcgee/age v0.0.0-20241002093043-152b6edfe56a
```
loa, line 138: I guess this slipped in, thanks for pushing this forward @brianmcgee !

brianmcgee: That's what the todo is about above. Once upstream merges, I can remove that.

age/keysource.go
```diff
@@ -125,7 +128,8 @@ func (i ParsedIdentities) ApplyToMasterKey(key *MasterKey) {
 }
 
 // Encrypt takes a SOPS data key, encrypts it with the Recipient, and stores
-// the result in the EncryptedKey field.
+// the result in the EncryptedKey field.cccccbgflnjtdhfdb
```
matdibu, line 131: that looks like a yubikey code, is that intentional?

Ma27:
Using this, works fine for `sops -d`. However if I specify public keys for creation_rules in a `.sops.yaml`, `sops` will break when editing these files since `parseRecipients` doesn't know what to do with age-plugin-yubikey pubkeys.

I fixed this with the following patch on top:

```patch
From 64e77bd60f8ebc8a0a5c7f8602a6f5855c892fd3 Mon Sep 17 00:00:00 2001
From: Maximilian Bosch <maximilian@mbosch.me>
Date: Wed, 20 Nov 2024 22:44:49 +0100
Subject: [PATCH] age/keysource: parse recipients using plugin system

Otherwise I get

    failed to parse input as Bech32-encoded age public key: malformed recipient "age1yubikey1...": invalid type "age1yubikey"

for `sops secrets/.../secrets.sops.yaml` in a directory with a
`.sops.yaml` that has creation rules with age1yubikey1* keys in its
creation rules.
---
 age/keysource.go | 19 ++++++++++++++-----
 1 file changed, 14 insertions(+), 5 deletions(-)

diff --git a/age/keysource.go b/age/keysource.go
index f04e4aff8..a9051c926 100644
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -304,12 +304,21 @@ func (key *MasterKey) loadIdentities() (ParsedIdentities, error) {
 
 // parseRecipient attempts to parse a string containing an encoded age public
 // key.
-func parseRecipient(recipient string) (*age.X25519Recipient, error) {
-	parsedRecipient, err := age.ParseX25519Recipient(recipient)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse input as Bech32-encoded age public key: %w", err)
+func parseRecipient(recipient string) (age.Recipient, error) {
+	switch {
+	case strings.HasPrefix(recipient, "age1") && strings.Count(recipient, "1") > 1:
+		parsedRecipient, err := plugin.NewRecipient(recipient, tui.PluginTerminalUI)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse input as age key from age plugin: %w", err)
+		}
+		return parsedRecipient, nil
+	default:
+		parsedRecipient, err := age.ParseX25519Recipient(recipient)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse input as Bech32-encoded age public key: %w", err)
+		}
+		return parsedRecipient, nil
 	}
-	return parsedRecipient, nil
 }
 
 // parseIdentities attempts to parse the string set of encoded age identities.
-- 
2.47.0

```

Feel free to pick this to your branch @brianmcgee :)

brianmcgee:
@Ma27 applied the patch, thanks :+1:

visualphoenix:
+1 would be great to have this in sops

visualphoenix:
I dream of a day I could use https://github.com/Foxboron/ssh-tpm-agent with sops…

Foxboron:
@felixfontein An alternative is for sops to use the `[filippo/plugin](https://github.com/FiloSottile/age/compare/filippo/plugin)` branch as it contains the required changes for this to work.

I've moved this over in `age-plugin-tpm` and considering the low velocity of the project I suspect it should be fine.

brianmcgee:
Ok cool, I'll have a look at rebasing this one today.

brianmcgee:
@felixfontein I manually tested this with the `age-plugin-fido2-hmac` plugin.

Ramblurr:
I'd like to test this PR too! Is this feature added transparently to the user? Or are there flags I need to pass to make it work?

brianmcgee:
> I'd like to test this PR too! Is this feature added transparently to the user? Or are there flags I need to pass to make it work?

It should be relatively transparent, you just need to pass an appropriate age key per the plugin you're using. 

Here's an example of how I'm using it with [age-plugin-fido2-hmac](https://github.com/olastor/age-plugin-fido2-hmac):

```console
❯ SOPS_AGE_KEY=$(age-plugin-fido2-hmac -m) sops test.yml
```
My `.sops.yaml` file contains a key entry of the form `age1fido2-hmac1xxxxxxxxxxx`.

felixfontein:
I'll take a closer look at this during the next days. The issues reported by CodeQL can likely be ignored (I think they're the same as in #1692, I'll have to take a cloesr look).

age/tui.go
```diff
@@ -67,3 +71,149 @@ func readPassphrase(prompt string) ([]byte, error) {
...
+	// handling of CONOUT$. Only when running a Windows binary from WSL2, the
+	// cursor would not go back to the start of the line with a simple LF.
+	// Honestly, it's impressive CONIN$ and CONOUT$ work at all inside WSL2.)
+	fmt.Fprintf(out, "\r\n"+CPL+EL)
+}
+
+// WithTerminal runs f with the terminal input and output files, if available.
+// WithTerminal does not open a non-terminal stdin, so the caller does not need
+// to check stdinInUse.
+func WithTerminal(f func(in, out *os.File) error) error {
```
felixfontein, line 105: Can you replace the copy of this function's body in `readPassphrase()` with a call to `WithTerminal`?

age/tui.go
```diff
@@ -1,5 +1,6 @@
 // These functions have been copied from the age project
 // https://github.com/FiloSottile/age/blob/v1.0.0/cmd/age/encrypted_keys.go
+// https://github.com/FiloSottile/age/blob/266c0940916da6bfa310fdd23156fbd70f9d90a0/tui/tui.go
```
felixfontein, line 3: Commit `266c0940916da6bfa310fdd23156fbd70f9d90a0` is not part of FiloSottile/age. Can you use https://github.com/FiloSottile/age/blob/3d91014ea095e8d70f7c6c4833f89b53a96e0832/cmd/age/tui.go instead (and also replace the code with the code from that commit)? It is also preferable to use lower-case identifiers (as in the original tui.go) so that the symbols defined in here are private to SOPS' `age` package.

age/tui.go
brianmcgee: Done

age/tui.go
brianmcgee: Done

felixfontein (approved):

felixfontein:
@Kranzes @Mic92 @brianmcgee and everyone else involved - thank you very much for your contributions!

PR #1647, opened 2024-10-09
Correctly handle comments during JSON serialization

### LINKED ISSUE #1646: Trailing comment in YAML causes `Error marshaling to json: invalid character '}' looking for beginning of object key string` when output type is JSON
Sops v3.9.1

To reproduce:

```
$ echo -e 'foo: bar\n#Hello world' | sops --input-type yaml --output-type yaml -e /dev/stdin > test.yaml
$ sops -d test.yaml 
foo: bar
#Hello world
$ sops --output-type json --decrypt --input-type yaml test.yaml 
Error dumping file: Error marshaling to json: invalid character '}' looking for beginning of object key string
```

felixfontein:
Fixes #1646.

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -186,18 +186,20 @@ func (store Store) encodeValue(v interface{}) ([]byte, error) {
 
 func (store Store) encodeArray(array []interface{}) ([]byte, error) {
 	out := "["
-	for i, item := range array {
+	empty := true
+	for _, item := range array {
 		if _, ok := item.(sops.Comment); ok {
 			continue
+		}
+		if !empty {
+			out += ","
 		}
 		v, err := store.encodeValue(item)
 		if err != nil {
 			return nil, err
 		}
 		out += string(v)
-		if i != len(array)-1 {
-			out += ","
-		}
+		empty = false
 	}
 	out += "]"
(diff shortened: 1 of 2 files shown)


stores/json/store.go
```diff
@@ -186,29 +186,35 @@ func (store Store) encodeValue(v interface{}) ([]byte, error) {
 
 func (store Store) encodeArray(array []interface{}) ([]byte, error) {
 	out := "["
-	for i, item := range array {
+	empty := true
+	for _, item := range array {
 		if _, ok := item.(sops.Comment); ok {
 			continue
 		}
+		if !empty {
```
devstein, line 194: We no longer need to skip the last item in the array? `if i != len(array)-1 {`

devstein (approved):
LGTM - thanks for testing.

One question, but I'm assuming we have sufficient tests coverage to catch a regression for that case

stores/json/store.go
felixfontein: Yes, because the comma is inserted before an item is added. The `empty` flag keeps track whether an item has been added already (if `empty==true` none has been added, so no comma is needed).

felixfontein:
@devstein thanks for reviewing this one as well!

PR #1652, opened 2024-10-19
fix(azkv): handle whitespace in Azure Key Vault URLs

sEpt0r:
### Problem:

The Azure Key Vault key parser would fail when URLs contained leading or trailing whitespace, which commonly occurs when using YAML Block Scalar syntax.

```yaml
creation_rules:
  - azure_keyvault: >-
      https://test.vault.azure.net/keys/test-key/a2a690a4fcc04166b739da342a912c90,
      https://test2.vault.azure.net/keys/another-test-key/cf0021e8b743453bae758e7fbf71b60e
```

This resulted in the error:

```bash
could not parse " https://test2.vault.azure.net/keys/another-test-key/cf0021e8b743453bae758e7fbf71b60e" into a valid Azure Key Vault MasterKey
```

### Fix:

- Added `strings.TrimSpace()` to clean the URL before parsing in `NewMasterKeyFromURL()`
- Added test case to verify handling of URLs with leading/trailing spaces

### DIFF
--- a/azkv/keysource.go
+++ b/azkv/keysource.go
@@ -76,6 +76,7 @@ func NewMasterKey(vaultURL string, keyName string, keyVersion string) *MasterKey
 // NewMasterKeyFromURL takes an Azure Key Vault key URL, and returns a new
 // MasterKey. The URL format is {vaultUrl}/keys/{keyName}/{keyVersion}.
 func NewMasterKeyFromURL(url string) (*MasterKey, error) {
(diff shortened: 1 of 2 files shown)


felixfontein (approved):
Thanks for your contribution!

Sorry for missing it on Saturday, I only noticed it now that you created the issue.

felixfontein:
@sEpt0r thanks for your contribution!

PR #1675, opened 2024-11-15
Add a clue for editors under "sops edit"

rsrchboy:
Sometimes there are things we'd prefer our editors do differently when editing files with secrets in them: different configuration, disable certain plugins, etc.  This change sets a clue in the launched editor's environment that can be used to easily and trivially determine if the editor was launched by `sops edit`.

### DIFF
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -265,6 +265,7 @@ func runEditor(path string) error {
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	cmd.Env = append(cmd.Environ(), "SOPS_IS_EDITING=true")
 	return cmd.Run()
 }
 


rsrchboy:
Correct me if I'm wrong, but the `SOPS_EDITOR` variable introduced in #1611 is set by the user, not `sops` itself, right?  If that's the case, there's two problems for trying to use it in this PR's use case:

1. We cannot depend on `SOPS_EDITOR` always being set for editors invoked by `sops`, and
2. We cannot depend that `SOPS_EDITOR` being set means we're actually being run under `sops`, as it is the sort of thing likely to be set in one's '~/.bashrc` / `~/.zshrc` / etc.

`SOPS_EDITOR` in #1611 is not a solution to the problem being addressed by this PR: providing an unambiguous, reliable clue that we're in an editor launched by `sops edit`.  These two PRs solve _different_ problems.

felixfontein:
Well, other programs can also set `SOPS_IS_EDITING` to `true`, so I don't see how the solution from this PR is better than the solution from the other PR for this problem.

rsrchboy:
> Well, other programs can also set `SOPS_IS_EDITING` to `true`, so I don't see how the solution from this PR is better than the solution from the other PR for this problem.

Because the work in the other PR is not a solution to this problem.

Nothing we do here can control other programs setting env variables.  What we can control is ensuring a specific var will always be set in the editor's environment, for editors launched by `sops edit`.  AFAICT #1611 does not do that with any var.

felixfontein:
You use #1611 by pointing `SOPS_EDITOR` to a script that sets whatever env variable you need and then launches your editor. Then you have what this PR does in a way more general form.

Which editor is used and how that editor is configured is highly system dependent (and thus perfectly suited for system-wide or user-wide env variables). If you can control the use of `SOPS_IS_EDITING`, you can also control setting `SOPS_EDITOR`.

rsrchboy:
> You use #1611 by pointing `SOPS_EDITOR` to a script that sets whatever env variable you need and then launches your editor. Then you have what this PR does in a way more general form.

Ah, I see what you're saying now.  Using `SOPS_EDITOR` to work around this problem is certainly feasible, but doesn't feel quite right, in a way I'm going to have to think about in order to be able to articulate.  (Likely that with this PR I'd be depending on something I know `sops` would set and could just handle it entirely in my `.vimrc`.)

nicklasfrahm:
@rsrchboy I've tested this using VS code, because I am not sure which editor you use. Would this not work?

```shell
export EDITOR="code -n --disable-extensions --wait"
sops example.txt
```

felixfontein:
Closing since this is basically covered by `SOPS_EDITOR`. Thanks for your contribution.

PR #1679, opened 2024-11-19
Allow SOPS to use custom AWS KMS and STS Endpoint

anandavj:
There's case when I have custom endpoint for my AWS service, with this changes I could specify STS and KMS endpoint for my AWS service
Example usage:
```
creation_rules:
  - kms: "kms-arn"
    aws_profile: "default"
    aws_kms_endpoint: "https://kms.us-west-1.customawsendpoint.com"
    aws_sts_endpoint: "https://sts.us-west-1.customawsendpoint.com"
```


keyservice/keyservice.pb.go
felixfontein, line 1: This file is automatically generated, you should not manually modify it, but instead modify the protobuf definition and re-generate it. Right now protobuf regeneration doesn't work (see #1576) though, but I hope that will get fixed soon.

felixfontein (commented):
Thanks for your contribution! I can't really comment on the AWS specific things, but as a general note, it looks like you should use gofmt to reformat the changes you made.

anandavj:
> Thanks for your contribution! I can't really comment on the AWS specific things, but as a general note, it looks like you should use gofmt to reformat the changes you made.

Thanks for the note, I've ran `gofmt` and it should be good now

anandavj:
I apologize forgot to run the test after merging with main, I've fix the protobuf and finish test it in my local @felixfontein 
Screenshot 2024-12-14 212146

felixfontein:
You have quite a few commits in this PR that do not belong there. Could you please **rebase** your PR (instead of adding more merge commits) to get rid of the superfluous commits? Thanks.

PR #1683, opened 2024-11-24
fix(config): allow usage of full config struct

### LINKED ISSUE #1682: Expose full config via Go API
Hi,

I am working on a project where I would like to use SOPS as the secret engine. In this context, I want to be agnostic whether the user is using my CLI or `sops`  to interact with the secret, but that requires that I generate the `.sops.yaml` in their repository. 

Currently, there is no simple public `Load` function that returns a `Config` and I am wondering if that is by design. My use-case requires both the full config `creation_rules`  and `destination_rules`.

I am happy to implement this, but I wanted to know if I am not doing anything wrong here.

nicklasfrahm:
This allows the usage of the full config `struct` in another Go program.

Closes #1682.

I added comments to edits that were not directly related to this PR, but that I considered a reduction of technical debt.

All tests pass locally.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,18 +17,10 @@ import (
 	"github.com/getsops/sops/v3/gcpkms"
 	"github.com/getsops/sops/v3/hcvault"
 	"github.com/getsops/sops/v3/kms"
-	"github.com/getsops/sops/v3/logging"
 	"github.com/getsops/sops/v3/pgp"
 	"github.com/getsops/sops/v3/publish"
-	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 )
-
-var log *logrus.Logger
-
-func init() {
-	log = logging.NewLogger("CONFIG")
-}
 
 type fileSystem interface {
 	Stat(name string) (os.FileInfo, error)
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -17,19 +17,11 @@ import (
...
 	"github.com/getsops/sops/v3/publish"
-	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 )
 
-var log *logrus.Logger
-
-func init() {
-	log = logging.NewLogger("CONFIG")
-}
```
nicklasfrahm, line 31: I got a warning from `go-staticcheck` and I could not see this being used anywhere so I removed it to avoid the warning.

config/config.go
```diff
@@ -162,10 +155,10 @@ func NewStoresConfig() *StoresConfig {
 }
 
 // Load loads a sops config file into a temporary struct
-func (f *configFile) load(bytes []byte) error {
+func (f *ConfigFile) load(bytes []byte) error {
 	err := yaml.Unmarshal(bytes, f)
 	if err != nil {
-		return fmt.Errorf("Could not unmarshal config file: %s", err)
+		return fmt.Errorf("could not unmarshal config file: %s", err)
```
nicklasfrahm, line 161: This was also flagged by `go-staticcheck`.

felixfontein: This is a change that should go into a feature release and not into a bugfix release since it changes the observable behavior a bit and isn't a bugfix. Because of that I would keep it separate from the other changes.

PR #1684, opened 2024-11-24
Allow users to disable version check via environment variable

### LINKED ISSUE #1665: Any way to disable probing GitHub?
I'm using sops in a network environment where I will never ever in a million years be able to contact github.com directly. I'd like to be able to call `sops -v` without it complaining that it can't check what's the latest version.

Is there any environment variable perhaps that controls this behavior?

nicklasfrahm:
Closes #1665.


felixfontein:
I changed the title since this isn't a fix, but a feature.

version/version.go
```diff
@@ -27,7 +29,7 @@ func PrintVersion(c *cli.Context) {
 
 	out.WriteString(fmt.Sprintf("%s %s", c.App.Name, c.App.Version))
 
-	if c.Bool("disable-version-check") {
+	if c.Bool("disable-version-check") || os.Getenv("SOPS_DISABLE_VERSION_CHECK") != "" {
```
felixfontein, line 32: Why not add the environment variable to https://github.com/getsops/sops/blob/7cfbfd951918a3ff21dda2d5e4fe22c82e4a8dd7/cmd/sops/main.go#L1519-L1522 (with `EnvVar: "SOPS_DISABLE_VERSION_CHECK"`)?

The behavior will be slightly different then from your current implementation (it uses https://pkg.go.dev/strconv#ParseBool to parse the string if it's not empty), but it would avoid using `os.Getenv` explicitly.

felixfontein (commented):
Thanks for your contribution!

nicklasfrahm:
@felixfontein I have adjusted the code according to your suggestion and updated the comment.

felixfontein (approved):

felixfontein:
@nicklasfrahm thanks for your contribution!

PR #1687, opened 2024-11-25
fix(config): remove unused variables and align error casing

nicklasfrahm:
This removes the issues mentioned in #1683.

@felixfontein Can we label this for the next feature release due to the change to the error message? (I don't know how to do this)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,18 +17,10 @@ import (
 	"github.com/getsops/sops/v3/gcpkms"
 	"github.com/getsops/sops/v3/hcvault"
 	"github.com/getsops/sops/v3/kms"
-	"github.com/getsops/sops/v3/logging"
 	"github.com/getsops/sops/v3/pgp"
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -165,7 +157,7 @@ func NewStoresConfig() *StoresConfig {
 func (f *configFile) load(bytes []byte) error {
 	err := yaml.Unmarshal(bytes, f)
 	if err != nil {
-		return fmt.Errorf("Could not unmarshal config file: %s", err)
+		return fmt.Errorf("could not unmarshal config file: %s", err)
```
felixfontein, line 160: Would you mind moving this change to another PR so we can merge the rest *now* instead of somewhen in the future? :)

felixfontein (approved):

felixfontein:
@nicklasfrahm thanks for your contribution!

PR #1690, opened 2024-12-01
Allow to encrypt and decrypt from stdin

### LINKED ISSUE #739: Piping using `/dev/stdin` doesn't work on Windows

felixfontein:
Allows to encrypt/decrpyt from stdin by providing no input filename to the `encrpyt` and `decrpyt` commands. For `encrypt`, `--filename-override` must be provided in that case to be able to select the proper creation rule.

Fixes #739
Ref: #1104

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -382,7 +382,7 @@ Encrypting and decrypting from other programs
 When using ``sops`` in scripts or from other programs, there are often situations where you do not want to write
 encrypted or decrypted data to disk. The best way to avoid this is to pass data to SOPS via stdin, and to let
 SOPS write data to stdout. By default, the encrypt and decrypt operations write data to stdout already. To pass
-data via stdin, you need to not provide an input filename. For encrpytion, you also must provide the
+data via stdin, you need to not provide an input filename. For encryption, you also must provide the
 ``--filename-override`` option with the file's filename. The filename will be used to determine the input and output
 types, and to select the correct creation rule.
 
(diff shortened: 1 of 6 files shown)


README.rst
```diff
@@ -382,33 +382,33 @@ Encrypting and decrypting from other programs
...
 To decrypt data, you can simply do:
 
 .. code:: sh
 
-	$ cat encrypted-data | sops decrypt /dev/stdin > decrypted-data
+	$ cat encrypted-data | sops decrypt --filename-override filename.yaml > decrypted-data
 
 To control the input and output format, pass ``--input-type`` and ``--output-type`` as appropriate. By default,
-``sops`` determines the input and output format from the provided filename, which is ``/dev/stdin`` here, and
+``sops`` determines the input and output format from the provided filename, which is the empty string here, and
```
sabre1041, line 396: What is the empty string in this case?

sabre1041 (changes requested):
Functionality works as described. A few minor items in the README

README.rst
felixfontein: Hmm, that wasn't a good way to describe it :) Internally the filename will be the empty string, but that isn't very helpful to most users. I've rewritten the whole decryption part in c67caf020644718f67a4b8b8e6d7f1b75f083f90, I hope it's better now.

sabre1041: much clearer!

felixfontein:

PR #1694, opened 2024-12-03
[updatekeys] Fix input-type cli flag being ignored

kleinschrader:
At my workplace we use sops, but we use yaml files and end them in `.yaml.enc` instead of just `.yaml`.
We recently wanted to add a new key and used `sops updatekeys --input-type=yaml  <file>` however it would throw json decode errors.

I do not know go and have never worked with it so this is the best i could conjure up and i didnt mange to get tests to run so i can only say that --input-type now works with updatekeys.

Any feedback is welcome and i am happy to make changes.


felixfontein:
This actually used to work, it got accidentally disabled in e9e2346fdd919bdff97300a62093814c1249ccac, which is part of #1273, and got included in 3.9.0. So this is definitely a bugfix.

cmd/sops/subcommand/updatekeys/updatekeys.go
```diff
@@ -45,12 +46,23 @@ func updateFile(opts Opts) error {
 	if err != nil {
 		return err
 	}
-	store := common.DefaultStoreForPath(sc, opts.InputPath)
```
felixfontein, line 48: I think this should simply be changed to `store := common.DefaultStoreForPathOrFormat(sc, opts.InputPath, opts.InputType)`. That function takes care of everything.

PR #1701, opened 2024-12-13
Add a way to set sops config location via envvar

sledzikowy:
It would be usefull if you would like to use config param without need to edit all the scripts to add param for that


felixfontein:
Thanks for your contribution! Sounds reasonable to me. Will take a closer look at the implementation soon.

sledzikowy:
@felixfontein Bump :)

felixfontein (approved):
Thanks for the reminder, and sorry it took so long!

felixfontein:
@sledzikowy thanks for your contribution!

PR #1704, opened 2024-12-14
CI: add 'persist-credentials: false' to checkout actions

felixfontein:
I ran zizmor (https://github.com/woodruffw/zizmor) on our GH actions, the only thing it complained about was missing `persist-credentials: false` for the checkout actions. Since we don't use the credentials later in the workflows (to my knowledge), we should be able to safely add it.


onedr0p (approved):

felixfontein:

PR #1711, opened 2024-12-18
Fix typo in the README's demo gif.

### LINKED ISSUE #1126: Error in readme file
The animation message stating the instructions has a spelling error .

image

not he keys^C--> not the keys

denschub:
This is, of course, a very critical bug and I'm very disappointed in this project - this unacceptable issue should have been fixed years ago.

I am kidding. But it still tickled me. This PR also removes the dependency on imgur, and replaces the 2 MiB GIF with a 1.3 MiB GIF.

Fixes #1126


sabre1041 (approved):

felixfontein:
I don't really like the idea of adding a 1.3 MiB blob to the repository, but right now I'm not sure of a better solution. Maybe we should add another repository for such binary data, and put the GIF there? @getsops/maintainers what do you think?

denschub:
@felixfontein I mean, if you want, you can also upload it to imgur again, send me the URL, and I'll update the PR. I just didn't do it, because it feels .. very wrong for this to be done by someone not even affiliated with the project. :)

felixfontein:
imgur is definitely a worse place where the image could live than a repository in https://github.com/getsops/. But I don't think it should be https://github.com/getsops/sops/ either. Putting binary blobs into code repositories is generally not a good idea.

PR #1719, opened 2024-12-24
Tests: use alternative container images from public.ecr.aws

### LINKED ISSUE #1667: Stop using images from Docker Hub in CI
Docker Hub's rate limits make CI fail from time to time, see for example https://github.com/getsops/sops/actions/runs/11560893288/attempts/1.

felixfontein:
Fixes #1667.

We cannot use the identical versions that were used before for nsmithuk/local-kms, since public.ecr.aws has only version 3.11.7, but not version 3.11.1 we used before.

Also vault seems to be more up-to-date on public.ecr.aws than on Docker Hub. This might make a difference.

Let's see what CI says...

### DIFF
--- a/hcvault/keysource_test.go
+++ b/hcvault/keysource_test.go
@@ -41,7 +41,7 @@ func TestMain(m *testing.M) {
 	}
 
 	// Pull the image, create a container based on it, and run it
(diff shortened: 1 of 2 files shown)


felixfontein:
Looks like public.ecr.aws has rate limiting as well that makes this even worse than Docker Hub.

PR #1721, opened 2024-12-25
updatekeys subcommand: fix input-type CLI flag being ignored

felixfontein:
An adjusted version of @kleinschrader's #1694 that uses `common.DefaultStoreForPathOrFormat()` instead of basically re-implementing that helper function.

Closes #1694


onedr0p (approved):

felixfontein:
@kleinschrader thanks for reporting this and providing a fix!
@onedr0p thanks for reviewing this!

PR #1733, opened 2025-01-17
Fix ToString for AWS KMS to include role, context, and profile

### LINKED ISSUE #1580: Add second `kms` list entry when SOPS file created
Hi folks. My approach to allowing users to create secrets, and then for CI to decrypt and use those secrets, is to allow users to create encrypted files using the `sops` CLI.

Users are auth'd using AWS SSO with an AWS CLI profile called `development`.

My creation rules in `.sops.yaml` look like this:

```yaml
creation_rules:
    - kms: arn:aws:kms:us-east-1:123456789123:alias/sops-kms-key
      aws_profile: development
```

So when the encrypted file (say, `secrets.yaml`) is created, it looks like this:

```yaml
username: ENC[...]
password: ENC[...]
sops:
    kms:
        - arn: arn:aws:kms:us-east-1:123456789123:alias/sops-kms-key
          created_at: "1970-01-01T00:00:00Z"
          enc: ABCDEF
          aws_profile: development
```

This is great, but I also need my CI system to decrypt these files. The CI server has an AWS IAM role attached (let's call it `CIServerRole`) with permissions to access the KMS key. My solution at the moment is to manually add another `kms` list entry like this:

```diff
[...]
sops:
    kms:
        - arn: arn:aws:kms:us-east-1:123456789123:alias/sops-kms-key
          created_at: "1970-01-01T00:00:00Z"
          enc: ABCDEF
          aws_profile: development
+       # Added manually after file has been encrypted.
+       - arn: arn:aws:kms:us-east-1:123456789123:alias/sops-kms-key
+         created_at: "1970-01-01T00:00:00Z"
+         enc: ABCDEF
+         aws_profile: ""
+         role: arn:aws:iam::123456789123:role/CIServerRole
```

I then commit and push this, and the CI server can happily decrypt the file using that IAM role.

I'd rather avoid having to add this manually, due to the toil and possibility of human error. I'd like the `sops` CLI to automatically populate this second "use a role, not a profile" entry to the `kms` list. Is this possible using `.sops.yaml`?

Thanks, and hugely appreciate the awesome work on SOPS!

Aaron

felixfontein:
Follow-up for #1493, in particular https://github.com/getsops/sops/pull/1493#discussion_r1625318840. Turns out the other fields were important as well.

This not only applies to #1493, but also other operations where `kms.ToString` is used, for example the `rotate` subcommand, the `updatekeys` subcommand, and the `publish` subcommand (the two latter use `DiffKeyGroups`, which relies on `ToString` to identify a full key).

Fixes #1580.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -94,6 +94,9 @@ creation_rules:
     - kms:
       - arn: foo
         aws_profile: bar
+      - arn: foo
+        context:
+          baz: bam
       pgp:
       - bar
       gcp_kms:
@@ -421,6 +424,7 @@ func TestLoadConfigFile(t *testing.T) {
 }
 
 func TestLoadConfigFileWithGroups(t *testing.T) {
+	bam := "bam"
 	expected := configFile{
 		CreationRules: []creationRule{
 			{
@@ -432,7 +436,18 @@ func TestLoadConfigFileWithGroups(t *testing.T) {
 				PathRegex: "",
 				KeyGroups: []keyGroup{
 					{
-						KMS:     []kmsKey{{Arn: "foo", AwsProfile: "bar"}},
+						KMS: []kmsKey{
+							{
+								Arn:        "foo",
+								AwsProfile: "bar",
+							},
+							{
+								Arn: "foo",
+								Context: map[string]*string{
+									"baz": &bam,
+								},
+							},
+						},
 						PGP:     []string{"bar"},
 						GCPKMS:  []gcpKmsKey{{ResourceID: "foo"}},
 						AzureKV: []azureKVKey{{VaultURL: "https://foo.vault.azure.net", Key: "foo-key", Version: "fooversion"}},
@@ -464,7 +479,7 @@ func TestLoadConfigFileWithMerge(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, 2, len(conf.KeyGroups))
 	assert.Equal(t, 1, len(conf.KeyGroups[0]))
-	assert.Equal(t, 22, len(conf.KeyGroups[1]))
+	assert.Equal(t, 23, len(conf.KeyGroups[1]))
 }
 
 func TestLoadConfigFileWithNoMatchingRules(t *testing.T) {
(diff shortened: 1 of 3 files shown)


config/config_test.go
```diff
@@ -538,9 +553,10 @@ func TestKeyGroupsForFileWithGroups(t *testing.T) {
 	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithGroups, t), "/conf/path", "whatever", nil)
 	assert.Nil(t, err)
 	assert.Equal(t, "bar", conf.KeyGroups[0][0].ToString())
-	assert.Equal(t, "foo", conf.KeyGroups[0][1].ToString())
+	assert.Equal(t, "foo||bar", conf.KeyGroups[0][1].ToString())
+	assert.Equal(t, "foo|baz:bam", conf.KeyGroups[0][2].ToString())
```
jonasbadstuebner, line 557: I think there should be a test-case where both context and profile are set?

kms/keysource.go
```diff
@@ -278,7 +311,18 @@ func (key *MasterKey) NeedsRotation() bool {
...
-	return key.Arn
+	arnRole := key.Arn
+	if key.Role != "" {
+		arnRole = fmt.Sprintf("%s+%s", key.Arn, key.Role)
+	}
+	context := kmsContextToString(key.EncryptionContext)
+	if key.AwsProfile != "" {
+		return fmt.Sprintf("%s|%s|%s", arnRole, context, key.AwsProfile)
+	}
+	if len(key.EncryptionContext) > 0 {
```
jonasbadstuebner, line 322:
```suggestion
	if context != "" {
```
Just a thought, that the `kmsContextToString` could return something even if the length of the `EncryptionContext` is zero. It doesn’t right now, but maybe in the future?
Checking if the string is empty is also more consistent with the other if statements.

jonasbadstuebner (changes requested):
Hey Felix,
first of all, thank you for creating this fix. Sorry that I broke stuff for others with my change.

The PR looks good to me.
I still am no AWS guy, so my comments on it can not cover the behavior of the code after your changes. I cannot judge if it is working correctly afterwards, but I understood the https://github.com/getsops/sops/issues/1580#issuecomment-2599051188 to match what you implemented now.

:rocket:

config/config_test.go
felixfontein: I've added some more tests.

kms/keysource.go
felixfontein: I don't think `kmsContextToString()` should ever return anything than the empty string for an empty `EncryptionContext`.

benlangfeld:
This fixes my use case, thanks!

/cc @iblackman

hiddeco (approved):

felixfontein:
@jonasbadstuebner @hiddeco thanks a lot for reviewing this!

Issue #1435, opened 2024-02-10
INI and DotEnv stores are not roundtrip-safe, and 


felixfontein:
While looking at #784 I noticed that the situation for both the INI and DotEnv stores is a lot worse than I thought: since they both escape newlines as literal `\n`s, but do not escape backslashes themselves, they cannot distinguish newlines from `\n` in plaintext, and both end up as newlines after decrypting.

DotEnv has its own set of problems; there used to be a PR which improved the parser and emitter, see #622, but that had to be reverted since it introduced a breaking change - see #706. The discussions in these PRs, especially the latter, also show that changes to the store formats are dangerous and should be avoided if possible.

This brings up a question on how to handle this mess. The only *real* fix is to do a breaking change. There are three breaking behaviors:
1. When loading an encrypted file written by an older version of sops. This can be handled since sops's version is included in the metadata (as `sops.version`). I think this is the very basic thing that any fix **must** do. This will complicate our code, though.
2. When emitting plaintext DotEnv or INI files. This can happen while decrypting DotEnv and INI files, but also when decrypting other files (`--output-type` parameter set to `ini` or `dotenv`). This is definitely a breaking change.
3. When encrypting a DotEnv or INI file which uses `\n` in the input, or some other form of quoting that's suddenly supported.

Especially for 2. and 3. it is probably best if the behavior is configurable. The big question is how can it be configured. For encrypting, `.sops.yaml` could be used, or command line flags. For decryption, only command line flags and file metadata can be used.

What do folks think about this? I don't expect that we can find a good solution quickly, but we definitely have to start a more focussed discussion :)

CC @getsops/maintainers

felixfontein:
A specific issue for INI quoting: #1597.

felixfontein:
A relevant PR: #752.

felixfontein:
The problem with DotEnv and INI is that there is no standardized file format of that name, and everyone has slightly different assumptions on how the format is defined. (This is even more true for INI than for DotEnv; it's just that for DotEnv the format that SOPS implements is rather far away from what everyone else is doing. I'm not saying this judgingly, I'm just stating how it is. We have to deal with this somehow.)

[...]

PR #706, opened 2020-07-24
Revert "Add standard newline/quoting behavior to dotenv store"

autrilla:
Reverts mozilla/sops#622

It seems based on #702 and #705 this change was not backwards compatible. I suggest we revert, make a new release and remove 3.6.0 from our Releases page.

PR #1748, opened 2025-02-04
Add (simple) support for quoted multiline variables in .env files

### LINKED ISSUE #965: SOPS cannot handle multi-line .env files
Let's say I have a .env file as follows:

`myfile.env`
```
HELLO="fooo
bar"
WORLD="blahblah"
```

This is a perfectly valid way of defining a .env file. In bash I can run `source myfile.env`. Thereater `echo $HELLO` will result in "fooo bar" (not that this is interpreted as a string on a single line).

However, SOPS cannot handle this. `sops -e --pgp=xxxxxxx myfile.env` resulting in `Error unmarshalling file: invalid dotenv input line: bar"`

Am I missing something, or is the behaviour here different from what is in bash?

MidasLamb:
This re-uses almost all previously existing logic, it only changes what constitutes a "line" when parsing the file.

This is not a fully fledged parser (e.g. escaping quotes doesn't work) but it should handle some additional cases now.

Fixes #965

### DIFF
--- a/stores/dotenv/store.go
+++ b/stores/dotenv/store.go
@@ -3,6 +3,7 @@ package dotenv //import "github.com/getsops/sops/v3/stores/dotenv"
 import (
 	"bytes"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


felixfontein:
Thanks for your contribution!

Please note that this is a **breaking change** and cannot be merged. There has been an earlier attempt at this (#622) that had to be reverted because of this (#706).

Ref: #1435

MidasLamb:
Thanks for the quick response @felixfontein !

I quickly went through the issue, and I'd like to point out that this PR does not add a full dotenv parser.
On the encryption side, it just looks make sure that it considers an entire key-value pair, not just lines, but how to value is encrypted has not changed.
On the decryption side there is maybe a small inconsistency, that shouldn't be an actual issue, where it might turn a "\n" combination into and actual newline.
i.e.
```
FOO="bar\nbaz"
```
becomes
```
FOO="bar
baz"
```
after an encryption->decryption cycle. But semantically, these values are actually the same, so this shouldn't be a breaking change.

This PR is just meant as a small incremental step to allow for multiline strings, without changing anything else really. In my reasoning the changes introduced here just expand the amount of allowed values, without doing any actual internal changes. 

The main issue seemed to be with this change: https://github.com/getsops/sops/pull/622/files#diff-c9aa3c7e333fa3bc6fa314822b8da1e449389f6bb14b68608e47d048c4997a08R104, where it adds single quotes unconditionally to values.

felixfontein:
While it's definitely true that your implementation comes very close to not breaking backwards compatibility, there are some cases that are broken. One of them you yourself listed. Your claim on "semantically equivalence" is problematic, since there is no well-defined `.env` file format. There are some assumptions that are true for many of these `.env` formats, but SOPS' implementation always was somewhat different. And since there is no formal specification to follow, every change is breaking.

Next to the change you mentioned above, is that if someone already encrypted `a=foo\nbar"` (something not compatible with many `.env` file format definitions, but with SOPS' definition), then your change will result in this being decrypted to
```
a=foo
bar"
```
Right now I don't see a way of adjusting the `.env` format (and the `.ini` format) without some versioning/file format identifier. I'll add some more thoughts on that to #1435 later today.

PR #1754, opened 2025-02-15
set subcommand: add --idempotent flag that will not write the file if no change happened

### LINKED ISSUE #939: `--set` always sets `lastmodified`, even when no change made
# Context

I am trying to make automated changes to a large YAML file encrypted with SOPS stored in Git.  Often, these changes are essentially No-Ops, i.e. the data I am setting is the same as the data already present. In order to prevent the file changing on every write due to rotating data keys, I am using `--set` to update the values in place.

# Problem

Even when none of the raw keys has been changed, the `sops->lastmodified` field is updated to the time at which `sops --set` was called.  This means there is always a delta for this file, albeit much smaller than it would be if I were writing it fresh each time.  However, _any_ delta will cause Git to store a new copy of the whole file, still leading to a massively bloated repository.

# Proposed Solution

The behaviour of `--set` should be the same as that of interactive editing: If no material change has been made, do not re-encrypt or write the file.  This will implicitly prevent the updating of the `lastmodified` field.

felixfontein:
Fixes #939.

This uses an extra flag (analogous to the `unset` subcommand) to preserve compatibility, and to avoid all kind of edge cases (input format can be different from output format, for example, in which case not writing the result in general would be a bad idea).

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -1356,6 +1356,10 @@ func main() {
 					Usage:  "comma separated list of decryption key types",
 					EnvVar: "SOPS_DECRYPTION_ORDER",
 				},
+				cli.BoolFlag{
+					Name:  "idempotent",
+					Usage: "do nothing if the given index already has the given value",
+				},
 			}, keyserviceFlags...),
 			Action: func(c *cli.Context) error {
 				if c.Bool("verbose") {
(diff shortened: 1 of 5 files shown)


sabre1041 (approved):
LGTM

felixfontein:

PR #1755, opened 2025-02-15
[WIP] Keep empty lines

felixfontein:
One annoying thing with SOPS-encrypted files is that all whitespace is normalized, and most importantly empty lines are removed.

I've been thinking for some time whether we can do better, so I did an experiment:

* DotEnv: works fine.
* INI: no way to access empty lines during parsing or emitting. This is due to the INI library used (`gopkg.in/ini.v1`).
* JSON: no way to access empty lines during parsing or emitting. This is due to the JSON library used (Golang's JSON encoder/decoder).
* YAML: it's possible to emit newlines, but no way to get them during parsing. One can try to use the line numbers from the node objects, *but* one has no idea how many lines the node itself needed, so no chance....

I guess for JSON it's OK to ignore newlines, but for INI and YAML it would be really nice. *Especially* for YAML, which I consider the main format for SOPS. Maybe this will be better with https://github.com/goccy/go-yaml...

I'm creating a PR for this experiment so that others can take a look, and maybe it's possible to improve on this once/if we switch to https://github.com/goccy/go-yaml.

Ref: #396

### DIFF
--- a/sops.go
+++ b/sops.go
@@ -114,6 +114,11 @@ type Comment struct {
 	Value string
 }
 
+// EmptyLines represents empty lines in the sops tree
+type EmptyLines struct {
+	Count int  // Must be positive
+}
+
 // TreeItem is an item inside sops's tree
 type TreeItem struct {
 	Key   interface{}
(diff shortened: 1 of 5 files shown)


felixfontein:
Related: https://github.com/yaml/go-yaml/issues/34

Issue #944, opened 2021-09-23
unknown type: time.Time


trjate:
`sops -e -i secrets.yaml`

```---
kind: Secret
apiVersion: v1
metadata:
    name: linkerd-identity-issuer
    namespace: linkerd
    labels:
        linkerd.io/control-plane-component: identity
        linkerd.io/control-plane-ns: linkerd
    annotations:
        linkerd.io/created-by: linkerd/cli stable-2.10.2
        linkerd.io/identity-issuer-expiry: 2031-01-01T00:00:00Z
data:
    crt.pem: LS0tLS...
    key.pem: LS0tLS1CR...
---
```
=>
`Error encrypting tree: Error walking tree: Cannot walk value, unknown type: time.Time`

Remove the `linkerd.io/identity-issuer-expiry: 2031-01-01T00:00:00Z` annotation an no problem.

enote-kane:
Still an issue with version 3.8.0:
```bash
$ echo 'test: 2023-10-04T15:42:23+00:00' | sops -e --input-type yaml /dev/stdin
Error encrypting tree: Error walking tree: Cannot walk value, unknown type: time.Time
```

GiovanniLeo:
I have the same error but I think that via the flag  **--unencrypted-regex** the issue can be solved but the problem already persists. Has someone found a way to overcome this that the value (a date) is needed?
This is my regexp **unencrypted-regex: '^([a-zA-Z]+[Dd]ate[a-zA-Z]+)$'**
The value that should ignored is this 
```YAML
server:
    startDate: 2021-19-09
    expirationDate: 2023-19-09
```

felixfontein:
My guess is that the main reason this hasn't been implemented yet is that decryption will yield an equivalent date/time string, which can look very different from the input.

bvanderhorn:
as a workaround: ensuring single or double quotes around the datetime string fixes this

felixfontein:
#1759 implements support for timestamps.

PR #1756, opened 2025-02-15
YAML: fail loading when timestamp is found

felixfontein:
Right now SOPS does not support timestamps (see also #944), and trying to encrypt a YAML document containing one exits at encryption time (and not load time).

This has the consequence that if you `sops edit` a YAML file and happen to enter a timestamp, then save and exit to let SOPS encrypt it, then it will successfully load the edited file, thus exit the edit loop, but then fail to encrypt, and exit. Thus you won't get back to the editor with a chance to quote the timestamp.

Changing this to a load-time error ensures that you stay in the edit loop and can fix the timestamp in the editor.

(The other formats don't support timetstamps either, so nothing is lost.)

One potential drawback: if the timestamp is not encrypted (due to one of the many options to not encrypt certain values), loading and storing it did work before. (See https://github.com/getsops/sops/issues/944#issuecomment-1875275498.) This makes this potentially a **breaking change**. WDYT?

### DIFF
--- a/stores/yaml/store.go
+++ b/stores/yaml/store.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
 	"github.com/getsops/sops/v3"
 	"github.com/getsops/sops/v3/config"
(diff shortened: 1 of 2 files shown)


felixfontein:
An alternative approach would be to convert timestamps to strings before encryption. I don't like that solution either, though, since it is a destructive conversion of content.

Maybe the best solution is to implement encryption and decryption of timestamps?

felixfontein:
Alternative: #1759.

PR #1692, opened 2024-12-01
Add SSH support for age

haoqixu:
Continues #1134.

Closes #1134
Closes #1496
Fixes #692

PR #1775, opened 2025-02-27
vendored age code: consolidate passphrase reading functionality

felixfontein:
Follow-up to #1692 and #1641.

While digging through age's history to figure out why `readPassphrase` vendored in #1692 wasn't there anymore in the latest age version, I noticed that basically readPassphrase was replaced by readSecret in https://github.com/FiloSottile/age/commit/c0e80ef2c985442f957990098ed1cb4b4dd4aa7e. This PR basically does the same, so we can stick to the latest version of cmd/age/tui.go.

Sorry @brianmcgee for the extra work...

### DIFF
--- a/age/encrypted_keys.go
+++ b/age/encrypted_keys.go
@@ -145,7 +145,7 @@ func unwrapIdentities(key string, reader io.Reader) (ParsedIdentities, error) {
 			Passphrase: func() (string, error) {
 				conn, err := gpgagent.NewConn()
 				if err != nil {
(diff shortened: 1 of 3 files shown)


sabre1041 (approved):
LGTM

felixfontein:
@sabre1041 thanks for reviewing this one as well!

Enzime:
@felixfontein do you know when the next SOPS release that includes these changes will be?

felixfontein:
@Enzime not really. I think/hope it will be out in less than a month, though.

Issue #1606, opened 2024-09-10
Support specifying Azure auth type


nbjohnson:
Currently the azure authentication flow seems to just pass off the auth options to the azidentity NewDefaultAzureCredential function: https://github.com/getsops/sops/blob/main/azkv/keysource.go#L232 While this does simplify things on sops side, that auth function does not allow fine grained control over what auth type is used. It is structured as a fall through but no ability to override, in my case managed identity and cli auth are populated so since managed identity will always be used even if I would like to use cli auth.

Can sops add some logic to be able to specify/disable certain auth types? (In my case to be able to disable managed identity, but the ability to generally control all auth types is probably ideal) I understand the simplicity of handing the auth flow off to the azure default auth flow, but a lot of control over the auth flow is lost in the process.

djw-bsn:
I have the same issue. We use a VM jumphost inside our private network. This VM has a managed identity but we don't want to authenticate to the Key Vault with this ID. We want to use the users cli identity but there is no way to override the standard order of priority so that the CLI is used. Unless someone knows otherwise?

djw-bsn:
I did a bit more digging and I think this used to be supported via the AZURE_AUTH_METHOD env var but was removed here: #1067

hmrc87:
Same issue her on Jenkins while using a docker in docker build sops is using sometimes the managed identity of the vm and sometimes the information gained after calling az login in the current shell

djw-bsn:
Any update here? We cannot upgrade SOPS on these jumphosts as the authentication to the Azure Key vault breaks if the VM has a managed identity. There is no way for us to force it to use the logged in users CLI identity. This doesn't feel like an enhancement? To me the removal of support for the AZURE_AUTH_METHOD environment variable has broken the functionality in this scenario? I would be very happy if there is a workaround but I'm not aware of anything. Removing the managed identity is not possible.

This also impacts terragrunt which has integration with SOPS to decrypt secrets.

djw-bsn:
I've never worked with go before so forgive me if I'm over simplifying here. But would it be possible to reintroduce the old AZURE_AUTH_METHOD or a new FORCE_AZURE_CLI_AUTH env var which would, if set, simply then call `NewAzureCLICredential()` instead of `NewDefaultAzureCredential()` here: https://github.com/getsops/sops/blob/main/azkv/keysource.go#L233

See also: https://github.com/Azure/azure-sdk-for-go/issues/23113#issuecomment-2192438069

Issue #885, opened 2021-06-04
Caching decryption key from Azure Keyvault?


LP0101:
We have a repo with several hundred secrets, all encrypted through SOPS. Through our CI, all these secrets have to be decrypted before we can work with them. It takes >5 minutes to decrypt them, as each decryption requires querying Azure.

Is there any way to cache the decryption key after the first decryption, so all following secrets encrypted with that key won't have to query AKV?

lieberlois:
@LP0101 Did you find a solution for this? We have the same issue!

PR #1777, opened 2025-02-28
azkv: Allow specifying auth method and add cachable authentication methods 

daytime-saxophone:
Azure is just really awful and slow to work with, and using sops with a azkv key to edit secrets from e.g. your laptop where you are authenticating with your user and not through a SP/MSI is nightmarishly slow.

To improve the situation this PR implements:
- Allow explicitly specifying auth method through environement variable, this bypass the need to go through the default authentication chain.
- Adds two cachable auth methods `InteractiveBrowserCredential` and `DeviceCodeCredential` which adds a very significant speedup compared to `AzureCliCredential`. See below.

The time to unlock a secret using sops `v3.9.4` and authenticating as a user through `azure-cli`: 
```
time sops -d ./path/to/secret
4.77s user 1.17s system 82% cpu 7.217 total
```

The time to unlock a secret using my PR and authenticating as a user with `InteractiveBrowserCredential` after the initial authentication:
```
> time SOPS_AZURE_AUTH_METHOD="BROWSER" sops -d ./path/to/secret
0.08s user 0.05s system 25% cpu 0.511 total
```

Related issues: #885, #1606 

I'm happy to add documentation, tests or whatever is needed. I just don't want to go through the trouble before I know if the design is acceptable, and that there is a chance that this could be merged.

### DIFF
--- a/azkv/keysource.go
+++ b/azkv/keysource.go
@@ -9,13 +9,19 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
 
+	"encoding/json"
+	"os"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	azidentitycache "github.com/Azure/azure-sdk-for-go/sdk/azidentity/cache"
 	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys"
 	"github.com/sirupsen/logrus"
 
@@ -25,6 +31,8 @@ import (
 const (
 	// KeyTypeIdentifier is the string used to identify an Azure Key Vault MasterKey.
 	KeyTypeIdentifier = "azure_kv"
+
+	sopsAzureAuthMethod = "SOPS_AZURE_AUTH_METHOD"
 )
 
 var (
@@ -230,7 +238,132 @@ func (key *MasterKey) TypeToIdentifier() string {
 // azidentity.NewDefaultAzureCredential.
 func (key *MasterKey) getTokenCredential() (azcore.TokenCredential, error) {
 	if key.tokenCredential == nil {
-		return azidentity.NewDefaultAzureCredential(nil)
+
+		authMethod := strings.ToUpper(os.Getenv(sopsAzureAuthMethod))
+		switch authMethod {
+		case "BROWSER":
+
+			return cachedInteractiveBrowserCredentials()
+
+		case "DEVICE_CODE":
+			return cachedDeviceCodeCredentials()
+		case "AZURE_CLI":
+			return azidentity.NewAzureCLICredential(nil)
+		case "MSI":
+			return azidentity.NewManagedIdentityCredential(nil)
+		// If "DEFAULT" or not explicitly specified then use the default authentication chain.
+		case "", "DEFAULT":
+			return azidentity.NewDefaultAzureCredential(nil)
+		default:
+			return nil, fmt.Errorf("Value `%s` is unsupported for `%s`, to resolve this either leave it unset or use one of DEFAULT/MSI/BROWSER/DEVICE_CODE", authMethod, sopsAzureAuthMethod)
+		}
 	}
 	return key.tokenCredential, nil
+}
+
+func sopsCacheDir() (string, error) {
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		return "", err
+	}
+
+	cacheDir := filepath.Join(userCacheDir, "/sops")
+
+	if err = os.MkdirAll(cacheDir, 0o700); err != nil {
+		return "", err
+	}
+
+	return cacheDir, nil
+}
+
+type CachableTokenCredential interface {
+	Authenticate(ctx context.Context, opts *policy.TokenRequestOptions) (azidentity.AuthenticationRecord, error)
+	GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error)
+}
+
+func cacheStoreRecord(cachePath string, record azidentity.AuthenticationRecord) error {
+	b, err := json.Marshal(record)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(cachePath, b, 0600)
+}
+
+func cacheLoadRecord(cachePath string) (azidentity.AuthenticationRecord, error) {
+	var record azidentity.AuthenticationRecord
+
+	b, err := os.ReadFile(cachePath)
+	if err != nil {
+		return record, err
+	}
+
+	err = json.Unmarshal(b, &record)
+	if err != nil {
+		return record, err
+	}
+
+	return record, nil
+}
+
+func cacheTokenCredential(cachePath string, tokenCredentialFn func(cache azidentity.Cache, record azidentity.AuthenticationRecord) (CachableTokenCredential, error)) (azcore.TokenCredential, error) {
+	cache, err := azidentitycache.New(nil)
+	// Errors if persistent caching is not supported by the current runtime
+	if err != nil {
+		return nil, err
+	}
+
+	cachedRecord, cacheLoadErr := cacheLoadRecord(cachePath)
+
+	credential, err := tokenCredentialFn(cache, cachedRecord)
+	if err != nil {
+		return nil, err
+	}
+
+	// If loading the authenticationRecord from the cachePath failed for any reason (validation, file doesn't exist, not encoded using json, etc.)
+	if cacheLoadErr != nil {
+		record, err := credential.Authenticate(context.Background(), nil)
+		if err != nil {
+			return nil, err
+		}
+
+		if err = cacheStoreRecord(cachePath, record); err != nil {
+			return nil, err
+		}
+	}
+
+	return credential, nil
+}
+
+func cachedInteractiveBrowserCredentials() (azcore.TokenCredential, error) {
+	cacheDir, err := sopsCacheDir()
+	if err != nil {
+		return nil, err
+	}
+	return cacheTokenCredential(
+		filepath.Join(cacheDir, "azure_auth_record_browser.json"),
+		func(cache azidentity.Cache, record azidentity.AuthenticationRecord) (CachableTokenCredential, error) {
+			return azidentity.NewInteractiveBrowserCredential(&azidentity.InteractiveBrowserCredentialOptions{
+				AuthenticationRecord: record,
+				Cache:                cache,
+			})
+		},
+	)
+}
+
+func cachedDeviceCodeCredentials() (azcore.TokenCredential, error) {
+	cacheDir, err := sopsCacheDir()
+	if err != nil {
+		return nil, err
+	}
+
+	return cacheTokenCredential(
+		filepath.Join(cacheDir, "azure_auth_record_device_code.json"),
+		func(cache azidentity.Cache, record azidentity.AuthenticationRecord) (CachableTokenCredential, error) {
+			return azidentity.NewDeviceCodeCredential(&azidentity.DeviceCodeCredentialOptions{
+				AuthenticationRecord: record,
+				Cache:                cache,
+			})
+		},
+	)
 }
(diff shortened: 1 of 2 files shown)


felixfontein:
Thanks for your contribution! I've been looking at this and thinking about this since yesterday, and I think the general idea is sound. (I'm not an Azure user though... I only use AZP for another open source project, and if you're writing that Azure is awful and slow to work with, I can guess what you mean since AZP definitely also falls into that category :D)

I'm a bit hesitant on the caching parts, since that writes credentials to disk (I guess not in an encrypted form, and even if its encrypted, it likely won't be anything safe). This should definitely be explicitly mentioned in the documentation, and maybe it would be better to rename the corresponding values for `SOPS_AZURE_AUTH_METHOD` to include "caching on disk" (I first just wanted to suggest to include that it's caching, but then I thought that later one might want to cache it somewhere else, like in the OS'es keyring, or some other agent, so including the method where it's cached is probably a good idea). WDYT?

I also have some smaller comments, I'm adding them here (even though they're not relevant to the general design) before I forget about them:
- Please undo the changes in go.mod to `go` and `toolchain`, that's likely the reason that the builds fails in CI.
- I'd probably move most of the new functions to a separate file, say `azkv/credentials.go`.
- Is there a reason the environment variable values are upper-case? I think lower-case is more common. (They also used to be lower-case with the `AZURE_AUTH_METHOD` env variable that was removed in #1067.)
- The error message in the switch's default case should mention that it's an environment variable. Someone might have forgotten that the environment variable is set and be wondering where to find this `SOPS_AZURE_AUTH_METHOD` setting.
- For the cached credential filenames, I'd use dashes instead of underscores.
- Generally I'd move the strings used (env variable values, cache filenames) into constants.

CC @hiddeco, since you created #1067 which removed the `AZURE_AUTH_METHOD` environment variable.

felixfontein:
(And I think it would be better if Azure's SDK would take care of things like this in its default auth flow, or would offer another auth flow that allows the user to configure this via env variables. But if it doens't offer that, I guess we have to do it ourselves...)

daytime-saxophone:
The [AuthenticationRecord](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#AuthenticationRecord) that I am storing in the function [cacheStoreRecord](https://github.com/getsops/sops/pull/1777/files#diff-17057699dd441d9c5ab351e49afb76566f698c8e4e812e39a052ec8b9f90031dR284) is not actually a secret, it only contains information like identifiers, version, username, etc.

The actual storing of the token is handled by the azure-sdk, the tokens are encrypted at rest via an os specific mechanisms:

Persistent caches are encrypted at rest using a mechanism that depends on the operating system [see docs](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/TOKEN_CACHING.MD#persistent-token-caching)
- Linux: kernel key retention service (keyctl)
- Macos: Keychain (requires cgo and native build tools)
- Windows: Data Protection API (DPAPI)

Their respective implementations can be found here:
* Linux: https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/cache/v0.3.2/sdk/azidentity/cache/linux.go
* Windows: https://github.com/AzureAD/microsoft-authentication-extensions-for-go/blob/main/cache/accessor/windows.go
* Macos: https://github.com/AzureAD/microsoft-authentication-extensions-for-go/blob/main/cache/accessor/darwin.go


> This should definitely be explicitly mentioned in the documentation, and maybe it would be better to rename the corresponding values for SOPS_AZURE_AUTH_METHOD to include "caching on disk" (I first just wanted to suggest to include that it's caching, but then I thought that later one might want to cache it somewhere else, like in the OS'es keyring, or some other agent, so including the method where it's cached is probably a good idea). WDYT?

Since the token is encrypted using a method specfic to the OS it makes it difficult to think of what to add as a contextual information in `SOPS_AZURE_AUTH_METHOD`, the best I can come up with is: `browser_cache_in_keyring` and `device_code_cache_in_keyring`. Does that seem fine?


I'll submit an update in the next few days with the fixes to the comments, adding some docs and tests (if I can figure something out to get around the interactive parts).

I just have one question: should I keep the environment variable `SOPS_AZURE_AUTH_METHOD` or should I change it back to `AZURE_AUTH_METHOD`?  The context for me changing it is that it was not clear to me when I was using `AZURE_AUTH_METHOD` that it was a sops specific variable until digging through the code.  That's why I figured it might be good to make that clear by prefixing it, kind of like `SOPS_AGE_KEY` is prefixed with sops. I'm happy to defer to you judgement on this though.

felixfontein:
> The [AuthenticationRecord](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#AuthenticationRecord) that I am storing in the function [cacheStoreRecord](https://github.com/getsops/sops/pull/1777/files#diff-17057699dd441d9c5ab351e49afb76566f698c8e4e812e39a052ec8b9f90031dR284) is not actually a secret, it only contains information like identifiers, version, username, etc.
> 
> The actual storing of the token is handled by the azure-sdk, the tokens are encrypted at rest via an os specific mechanisms:
> 
> Persistent caches are encrypted at rest using a mechanism that depends on the operating system [see docs](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/TOKEN_CACHING.MD#persistent-token-caching)

Thanks for clarifying that! In that case, I don't mind the caching on disk.

> > This should definitely be explicitly mentioned in the documentation, and maybe it would be better to rename the corresponding values for SOPS_AZURE_AUTH_METHOD to include "caching on disk" (I first just wanted to suggest to include that it's caching, but then I thought that later one might want to cache it somewhere else, like in the OS'es keyring, or some other agent, so including the method where it's cached is probably a good idea). WDYT?
> 
> Since the token is encrypted using a method specfic to the OS it makes it difficult to think of what to add as a contextual information in `SOPS_AZURE_AUTH_METHOD`, the best I can come up with is: `browser_cache_in_keyring` and `device_code_cache_in_keyring`. Does that seem fine?

Since no sensitive information is cached, I think it's OK to not use a prefix, or simply use `cached-` or something like that (without more details).

> I just have one question: should I keep the environment variable `SOPS_AZURE_AUTH_METHOD` or should I change it back to `AZURE_AUTH_METHOD`? The context for me changing it is that it was not clear to me when I was using `AZURE_AUTH_METHOD` that it was a sops specific variable until digging through the code. That's why I figured it might be good to make that clear by prefixing it, kind of like `SOPS_AGE_KEY` is prefixed with sops. I'm happy to defer to you judgement on this though.

I would keep the prefix. I think using env variables without a `SOPS_` prefix is only a good idea if they use same convention as the tool they hint at (which is usually best handled by a library provided by that tool). Also the new environment variable uses different values and semantics than the previous one, so its previous existence is no reason to re-use that name.

daytime-saxophone:
I've updated the minor comments and added docs for the feature. However I am not sure what to do about the `go` and `toolchain` version in `go.mod` file. This auto upgrading behaviour of the go toolchain is a bit new to me.


If I run `go mod tidy` with the `toolchain` and `go` version that is specified in the `go.mod` file it gives me this error:
```
> GOTOOLCHAIN=go1.23.6 go mod tidy -go=1.22
go: cloud.google.com/go/auth@v0.15.0 requires go@1.23.0, but 1.22 is requested
```

If I run `go mod tidy` with the `toolchain` as the latest `go1.22` patch release it gives me this error:
```
> GOTOOLCHAIN=go1.22.12 go mod tidy 
go: golang.org/x/crypto@v0.35.0 requires go >= 1.23.0 (running go 1.22.12; GOTOOLCHAIN=go1.22.12)
```

If I run `go mod tidy` with the `toolchain` that is specified in the `go.mod` but don't specify `-go=<version>` it succeeds but it modifies the `go` version in the `go.mod` file. 
```
GOTOOLCHAIN=go1.23.6 go mod tidy
```

felixfontein:
Please note that you must sign off every single commit, otherwise we cannot merge this (see the failing DCO test).

daytime-saxophone:
> Please note that you must sign off every single commit, otherwise we cannot merge this (see the failing DCO test).
I've rebased all the changes to a single commit, will sign the DCO.

Since the `main` branch has updated the go version in the `go.mod` file that also fixed the issues I was having where go wanted to upgrade to `1.23`. So hopefully the tests will pass now. 

Since we've been using the forked version for a little while now I've realized that on the initial authentication the browser and device-code login auth methods would pollute the `stdout`, I've provided a fix to this issue in this PR.

azkv/keysource.go
```diff
@@ -230,7 +242,142 @@ func (key *MasterKey) TypeToIdentifier() string {
...
+	}
+
+	cacheDir := filepath.Join(userCacheDir, "/sops")
+
+	if err = os.MkdirAll(cacheDir, 0o700); err != nil {
+		return "", err
+	}
+
+	return cacheDir, nil
+}
```
hiddeco, line 279: I think it would be better to create a `cache` package, with something like:

```go
var (
	cacheDir     string
	cacheDirErr  error
	cacheDirOnce sync.Once
)

// Directory returns the path to the SOPS cache directory, creating it if it
// does not exist. The directory with the name "sops" is created in the user's
// cache directory (see [os.UserCacheDir]). Once created, it will not be
// recreated, even if the function is called multiple times. If an error occurs
// while creating the directory, it will be returned on later calls.
func Directory() (string, error) {
	cacheDirOnce.Do(func() {
		base, err := os.UserCacheDir()
		if err != nil {
			cacheDirErr = err
			return
		}

		cacheDir = filepath.Join(base, "sops")
		cacheDirErr = os.MkdirAll(cacheDir, 0o700)
	})
	if cacheDirErr != nil {
		return "", cacheDirErr
	}
	return cacheDir, nil
}

// Put writes the provided data to a file in the SOPS cache directory.
// At present, no validation is performed on the file name, and it is the
// caller's responsibility to ensure that the file name is safe (i.e., does not
// contain path traversal characters or other unsafe elements).
func Put(fileName string, data []byte) error {
	dir, err := Directory()
	if err != nil {
		return err
	}

	filePath := filepath.Join(dir, fileName)
	if err = os.WriteFile(filePath, data, 0o600); err != nil {
		return fmt.Errorf("failed to write to cache file %s: %w", filePath, err)
	}
	return nil
}

// Get reads the contents of a file from the SOPS cache directory.
// If the file does not exist or cannot be read, an error will be returned.
func Get(fileName string) ([]byte, error) {
	dir, err := Directory()
	if err != nil {
		return nil, err
	}

	filePath := filepath.Join(dir, fileName)
	data, err := os.ReadFile(filePath)
	if err != nil {
		return nil, fmt.Errorf("failed to read cache file %s: %w", filePath, err)
	}
	return data, nil
}

// Exists checks if a file exists in the SOPS cache directory.
func Exists(fileName string) (bool, error) {
	dir, err := Directory()
	if err != nil {
		return false, err
	}

	filePath := filepath.Join(dir, fileName)
	if _, err = os.Stat(filePath); err != nil {
		if os.IsNotExist(err) {
			return false, nil
		}
		return false, err
	}
	return true, nil
}
```

azkv/keysource.go
```diff
@@ -9,14 +9,21 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
 
+	"encoding/json"
+	"os"
```
hiddeco, line 18: These imports aren't properly ordered.

hiddeco (commented):
To actually get this into a shape where it is mergable, changes have to be made to build the macOS binaries with CGO enabled. This will likely raise some issues around cross-compilation.

If we go the CGO route, we should probably plan out how this affects our CI, release process, and broader contributor experience.

daytime-saxophone:
> To actually get this into a shape where it is mergable, changes have to be made to build the macOS binaries with CGO enabled. This will likely raise some issues around cross-compilation.
> If we go the CGO route, we should probably plan out how this affects our CI, release process, and broader contributor experience.

Ah, it was CGO that was the issue!  I have been maintaining a fork with this change for the past 9 months now which we use in my team, several people in my team are mac users so I've been very confused why it refused to build here in the CI. But it's because we build it with nix which defaults to CGO. 

I've spent a fair bit of time today trying to get this cross compiling and the only method so far that actually built the thing was using zig. Would using the c compiler and linker from zig to build the cross compiled version be an acceptable solution?

Otherwise using `macos-latest` seems to work out of the box for building the darwin/arm64 and darwin/amd64 targets. The tests  fail but you are not actually running the tests for darwin right now either.

wdyt? @hiddeco

PR #1781, opened 2025-03-02
Fix caching of Metadata.DataKey

felixfontein:
At the end of the function, the key was stored in `m.DataKey`. Since `m` was a copy of the called object, this did not cache anything.

(Also found with `make staticcheck`; I created a separate PR since this is clearly a bugfix.)


felixfontein:

PR #1794, opened 2025-03-10
Add support for `oauth2.TokenSource` in GCP KMS

### LINKED ISSUE #1792: GCP KMS: Support the `option.WithTokenSource()` for the GCP KMS client, similar to `kms.CredentialsProvider` and `azkv.TokenCredential`
First of all, thank you so very much for SOPS! In my opinion it powers a very critical component of GitOps. I'm a Flux core maintainer so you can guess how much I love GitOps and consequently how much I also love SOPS :)

Now, on to the feature request :)

As a user I'd like the ability to use multi-tenant workload identity for decrypting SOPS-encrypted secrets in CNCF Flux through the `Kustomization` API using GCP KMS. How I'm planning to do it on the Flux side is detailed in [this section](https://github.com/fluxcd/flux2/blob/rfc-multi-tenant-workload-identity/rfcs/0010-multi-tenant-workload-identity/README.md#kustomization-api) of the RFC for Multi-Tenant Workload Identity (RFC [link](https://github.com/fluxcd/flux2/pull/5209)).

In order to enable this use case it is necessary to implement an interface for GCP KMS similar to the interfaces for AWS and Azure `kms.CredentialsProvider` and `azkv.TokenCredential`. The equivalent interface for GCP would be `oauth2.TokenSource`. There are a number of ways how one could create an object that implements this interface, so this would definitely be very helpful for users of the GCP SOPS SDK, and hence would enable more advanced methods of GCP KMS authentication to be used for SOPS.

I'm more than willing to contribute this feature, my plan is to open a PR next weekend, please consider reviewing it! 🙏 🙏 🙏 

Edit: PR opened sooner

matheuscscp:
Closes #1792

### DIFF
--- a/gcpkms/keysource.go
+++ b/gcpkms/keysource.go
@@ -12,6 +12,7 @@ import (
 	kms "cloud.google.com/go/kms/apiv1"
 	"cloud.google.com/go/kms/apiv1/kmspb"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/oauth2"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
 
@@ -50,6 +51,11 @@ type MasterKey struct {
 	// for NeedsRotation.
 	CreationDate time.Time
 
+	// tokenSource contains the oauth2.TokenSource used by the GCP client.
+	// It can be injected by a (local) keyservice.KeyServiceServer using
+	// TokenSource.ApplyToMasterKey.
+	// If nil, the remaining authentication methods are attempted.
+	tokenSource oauth2.TokenSource
 	// credentialJSON is the Service Account credentials JSON used for
 	// authenticating towards the GCP KMS service.
 	credentialJSON []byte
@@ -80,6 +86,22 @@ func MasterKeysFromResourceIDString(resourceID string) []*MasterKey {
 		keys = append(keys, NewMasterKeyFromResourceID(s))
 	}
 	return keys
+}
+
+// TokenSource is an oauth2.TokenSource used for authenticating towards the
+// GCP KMS service.
+type TokenSource struct {
+	source oauth2.TokenSource
+}
+
+// NewTokenSource creates a new TokenSource from the provided oauth2.TokenSource.
+func NewTokenSource(source oauth2.TokenSource) TokenSource {
+	return TokenSource{source: source}
+}
+
+// ApplyToMasterKey configures the TokenSource on the provided key.
+func (t TokenSource) ApplyToMasterKey(key *MasterKey) {
+	key.tokenSource = t.source
 }
 
 // CredentialJSON is the Service Account credentials JSON used for authenticating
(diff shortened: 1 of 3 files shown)


sabre1041:
@matheuscscp would you be able to address the merge conflict?

matheuscscp:
@sabre1041 Done!

sabre1041:
Hey @matheuscscp thanks for clearing up the merge conflict. Question as #1578 looks very similar to yours. Thoughts?

matheuscscp:
Hey @sabre1041, I think it's orthogonal. I really don't want to retrieve this token from an environment variable, my use case is multi-tenant. The same Go process will call the SOPS SDK passing different tokens at different moments (CNCF Flux, see the issue I opened).

I'd really like to have something in the exact same style we already have for AWS and Azure, which is what this PR introduces:

https://github.com/getsops/sops/blob/5355c249aa8ea413bea6b2ea20a029cb21048e99/kms/keysource.go#L72-L76

https://github.com/getsops/sops/blob/5355c249aa8ea413bea6b2ea20a029cb21048e99/azkv/keysource.go#L58-L62

The fact that this other PR also uses `option.WithTokenSource()` has actually little to do with this one, you could retrieve a token from a thousand different places and still feed it via `option.WithTokenSource()`. This one is for passing a token in memory, calling SOPS as a Go library.

sabre1041:
> Hey @sabre1041, I think it's orthogonal. I really don't want to retrieve this token from an environment variable, my use case is multi-tenant. The same Go process will call the SOPS SDK passing different tokens at different moments (CNCF Flux, see the issue I opened).
> 
> I'd really like to have for GCP something in the exact same style we already have for AWS and Azure, which is what this PR introduces:
> 
> https://github.com/getsops/sops/blob/5355c249aa8ea413bea6b2ea20a029cb21048e99/kms/keysource.go#L72-L76
> 
> https://github.com/getsops/sops/blob/5355c249aa8ea413bea6b2ea20a029cb21048e99/azkv/keysource.go#L58-L62
> 
> The fact that this other PR also uses `option.WithTokenSource()` has actually little to do with this one, you could retrieve a token from a thousand different places and still feed it via `option.WithTokenSource()`. This one is for passing a token in memory, calling SOPS as a Go library.

That makes sense. it is indeed related but has distinct use cases/differences

gcpkms/keysource.go
```diff
@@ -216,6 +238,8 @@ func (key *MasterKey) newKMSClient() (*kms.KeyManagementClient, error) {
 
 	var opts []option.ClientOption
 	switch {
+	case key.tokenSource != nil:
+		opts = append(opts, option.WithTokenSource(key.tokenSource))
```
felixfontein, line 242: Can you update the comment for this function to also mention `tokenSource` (and not just `credentialJSON` and `grpcConn` etc.)?

felixfontein (commented):
Just one small comment, besides that it looks good from my point of view.

gcpkms/keysource.go
matheuscscp: Fixed, thanks!

felixfontein:
Looks good from my point of view!

hiddeco (approved):

sabre1041 (approved):
LGTM

felixfontein:
@matheuscscp thanks for your contribution!
@sabre1041 @hiddeco thanks for reviewing!

PR #1798, opened 2025-03-13
Warn user about .yml config file extension

onnimonni:
Hey,

I spent way too much time to debug my environment and it turned out that the reason was that I had added my config into `.sops.yml` instead of `.sops.yaml`.

```sh
# This is not working
$ sops secrets/server.yaml
config file not found, or has no creation rules, and no keys provided through command line options

# fix the problem
$ mv .sops.yml .sops.yaml                                                                                                                                                 
renamed '.sops.yml' -> '.sops.yaml'

# It works now
$ sops secrets/server.yaml
```

This is my first contribution into sops and I'm happy to hear feedback and improve this. Thanks in advance 🙇

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -39,23 +39,30 @@ var fs fileSystem = osFS{stat: os.Stat}
 const (
 	maxDepth       = 100
 	configFileName = ".sops.yaml"
-	configFileNameAlternative = ".sops.yml"
 )
 
 // FindConfigFile looks for a sops config file in the current working directory and on parent directories, up to the limit defined by the maxDepth constant.
 func FindConfigFile(start string) (string, error) {
 	filepath := path.Dir(start)
 	for i := 0; i < maxDepth; i++ {
-		for _, configFileName := range []string{configFileName, configFileNameAlternative} {
-			_, err := fs.Stat(path.Join(filepath, configFileName))
-			if err != nil {
-				filepath = path.Join(filepath, "..")
-			} else {
-				return path.Join(filepath, configFileName), nil
-			}
+		_, err := fs.Stat(path.Join(filepath, configFileName))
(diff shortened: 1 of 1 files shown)


felixfontein:
Thanks for your contribution!

Personally I'm against this change. Having different config file names is usually more a source of confusion.

There's already a note in the documentation, though it's a bit hidden since there's no central configuration documentation so far: https://github.com/getsops/sops?tab=readme-ov-file#213using-sopsyaml-conf-to-select-kms-pgp-and-age-for-new-files

onnimonni:
> Personally I'm against this change. Having different config file names is usually more a source of confusion.

I think this is a good point and similarly as we have linting rules in the code it's fine if sops wants to keep this convention.

How about logging a warning instead if `.sops.yml` variant is found? Or throwing an error?

The warning would make it more obvious for the user that their config was found but `.yml` is not supported.

AFAIK most of the other software world allows both variants and I assume I will not be the last one to stumble for this.

felixfontein:
> How about logging a warning instead if `.sops.yml` variant is found? Or throwing an error?

I think it would be best to report a warning if SOPS needs a config file and doesn't find one, but does find a `.sops.yml` while searching for `.sops.yaml`. I think that helps already a lot and improves the failure message you'll get.

onnimonni:
Sorry that it took so long and I'm definitely not golang expert but I think this should do it 👍

felixfontein:
I created a new PR (#1820) which comes closer to what I'd prefer the warning to behave like, and which also compiles :) Thanks for your contribution though, I think this warning is really helpful!

onnimonni:
Thanks @felixfontein 🙇! I'm very happy with the new error messages 👍

PR #1804, opened 2025-03-19
Run go mod tidy

onnimonni:
Hey,

Because the `go.mod` file is not up to date different upstream packaging software need to rely on weird hacks to build from main branch.

[For example if you look homebrew they by default always need to run](https://github.com/homebrew/homebrew-core/blob/e98c759b61cab72ac15caef771e10a7d44d69de8/Formula/s/sops.rb#L22C5-L22C31):
```sh
$ go mod tidy
```

Because of this I'm also not able to build the current master with nix as seen [in this pretty long thread in nix forums](https://discourse.nixos.org/t/how-to-override-and-build-certain-package-from-master-main-branch/61877/24).

Would it be possible to update this?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/getsops/sops/v3
 
-go 1.22
+go 1.23.0
+
 toolchain go1.24.1
 
 require (


onnimonni:
Without this update I get following error when trying to build current main branch with nix:
```
$ nix-build --expr 'with (import <nixpkgs> { });
                                   (sops.override {
                                     buildGo122Module = args: buildGo122Module ( args // { vendorHash = "sha256-anKhfq3bIV27r/AvvhSSniKUf72gklv8waqRz0lKypQ="; });
                                   }).overrideAttrs {
                                     src = fetchFromGitHub {
                                       owner = "getsops";
                                       repo = "sops";
                                       rev = "2eb776b01df5df04eee626da3e99e9717fffd9e0";
                                       hash = "sha256-VB4/DyQoQnV/AAXteJPsD2vbtAilZcJPTCXk2nvUZU8=";
                                     };
                                   }'

...

> go: cloud.google.com/go in vendor/modules.txt requires go >= 1.23.0 (running go 1.22.12; GOTOOLCHAIN=local)
```

felixfontein:
CC @getsops/maintainers since others know more about Go packaging than me...

onnimonni:
Sorry I'm not familiar but I would want to learn. What do you mean when you say?
> (once vendoring is done).

I noticed that even [the latest sops release in nixpkgs needs to patch the go.mod file to get it compiling](https://github.com/NixOS/nixpkgs/blob/4833a5fb0d03c5206f843e81d643d3b9920c0354/pkgs/by-name/so/sops/package.nix#L23-L26):
```
postPatch = ''
    substituteInPlace go.mod \
      --replace-fail "go 1.22" "go 1.22.7"
  '';
```

And they are using go 1.22 in the builds.

felixfontein:
To build it locally, I usually run `make vendor` (which runs `go mod tidy` and `go mod vendor`), then undo the changes to `go.mod` to get rid of the bumped Go version, and then `make install` (which runs `go install github.com/getsops/sops/v3/cmd/sops`). (When using the Makefile, `GOPROXY=https://proxy.golang.org` is set for all three calls, if that is of any relevance.)

felixfontein:
Closing since no longer needed due to #1819. Thanks for your contribution!

PR #1811, opened 2025-03-24
Added SOPS_AGE_KEY_CMD option to age, fixes #1323

### LINKED ISSUE #1323: Generalize SOPS_AGE_KEY to arbitrary executable
I currently see a few workarounds of the form `SOPS_AGE_KEY=$(…) sops …` floating around.

I would like to store my age key in my password manager, not have to specify a workaround each invocation as above, not wrap my sops binary in an outer shell script to set `SOPS_AGE_KEY` (difficult when sops is provided by various nix flakes via direnv), and not just set `SOPS_AGE_KEY` as environment for my entire user session on login.

It seems to me that these and similar usecases would benefit from a new environment variable `SOPS_AGE_KEY_EXEC` (bikeshedding welcome) that would contain a command that would be passed to `exec.Command` and run without arguments.  
The stdout the command produces would then be used as an additional sops identity.

This would allow me to specify `SOPS_AGE_KEY_EXEC` globally for my entire user session and have my, securely stored, age key picked up by sops, even in my various `direnv`en.

danilobuerger:
Fixes #1323.


age/keysource.go
```diff
@@ -310,6 +315,18 @@ func (key *MasterKey) loadIdentities() (ParsedIdentities, error) {
 		readers[SopsAgeKeyFileEnv] = f
 	}
 
+	if ageKeyCmd, ok := os.LookupEnv(SopsAgeKeyCmdEnv); ok {
+		args, err := shellquote.Split(ageKeyCmd)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse command %s: %w", ageKeyCmd, err)
```
felixfontein, line 321: How about mentioning the environment variable's name in the error message?
```suggestion
			return nil, fmt.Errorf("failed to parse command %s from %s: %w", ageKeyCmd, SopsAgeKeyCmdEnv, err)
```

age/keysource.go
```diff
@@ -310,6 +315,18 @@ func (key *MasterKey) loadIdentities() (ParsedIdentities, error) {
...
+		}
+		out, err := exec.Command(args[0], args[1:]...).Output()
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute command %s: %w", ageKeyCmd, err)
```
felixfontein, line 325: Same here:
```suggestion
			return nil, fmt.Errorf("failed to execute command %s from %s: %w", ageKeyCmd, SopsAgeKeyCmdEnv, err)
```

felixfontein (commented):
Thanks for your contribution! The code generally looks fine, though I'm not sure whether adding yet another external dependency (`github.com/kballard/go-shellquote`) is a good idea.

danilobuerger:
Hi @felixfontein 👋 thanks for the quick review. I have made the requested changes.

haslersn:
> uses the `github.com/google/shlex` package for that, which has been unmaintained for six years

A software that's not broken doesn't need updates.

felixfontein:
@danilobuerger thanks for your contribution!

PR #1813, opened 2025-03-25
Improve Shamir Secret Sharing code

felixfontein:
This applies patches to the vendored Shamir Secret Sharing code that have been applied to the source (HashiCorp Vault's code before the Vault license change). It also cleans it up a bit and adds a lot of comments to make the "magic" happening in some functions more transparent and maintainable.

I'm making this public now even though the next release is still a few days out since this problem of the Vault code has been long known (see https://github.com/advisories/GHSA-vq4h-9ghm-qmrr), so there's not much sense in keeing it hidden.


hiddeco (approved):

felixfontein:

PR #1816, opened 2025-03-28
Add option to explicitly check for the latest version; deprecate current default of always doing that unless disabled

felixfontein:
Ref: #1665

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -1583,6 +1583,10 @@ func main() {
 			Usage:  "do not check whether the current version is latest during --version",
 			EnvVar: "SOPS_DISABLE_VERSION_CHECK",
 		},
+		cli.BoolFlag{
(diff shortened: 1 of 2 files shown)


cmd/sops/main.go
```diff
@@ -1583,6 +1583,10 @@ func main() {
 			Usage:  "do not check whether the current version is latest during --version",
 			EnvVar: "SOPS_DISABLE_VERSION_CHECK",
 		},
+		cli.BoolFlag{
+			Name:  "enable-version-check",
```
hiddeco, line 1587: I wonder if something like `check-for-updates` (or `has-updates`) would be a nicer way to phrase this.

felixfontein: I changed it to `--check-for-updates`.

hiddeco (approved):

PR #1820, opened 2025-03-30
Warn about `.sops.yml` files found while searching for `.sops.yaml`

felixfontein:
Some users are confused since they try to create a config file named `.sops.yml` without realizing that it must be called `.sops.yaml`. `sops` does not warn about it, so they spend some time trying to figure out what went wrong.

This prints a message such as
```
[CMD]	 WARN[0000] Ignoring "../.sops.yml" when searching for config file. The config file must be called ".sops.yaml". Found and using "../../.sops.yaml" further up the directory tree instead. 
```
when a real `.sops.yaml` was found somewhere further up than `.sops.yml`, and 
```
[CMD]	 WARN[0000] Ignoring "../.sops.yml" when searching for config file. The config file must be called ".sops.yaml". 
```
when no `.sops.yaml` was found but a `.sops.yml` file.

Closes #1798, which also tried to implement something like this.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -44,7 +44,13 @@ import (
 	"github.com/getsops/sops/v3/version"
 )
 
-var log *logrus.Logger
+var (
+	log *logrus.Logger
+
+	// Whether the config file warning was already shown to the user.
+	// Used and set by findConfigFile().
+	showedConfigFileWarning bool
+)
 
 func init() {
 	log = logging.NewLogger("CMD")
@@ -364,7 +370,7 @@ func main() {
 				if c.GlobalString("config") != "" {
 					configPath = c.GlobalString("config")
 				} else {
-					configPath, err = config.FindConfigFile(".")
+					configPath, err = findConfigFile()
 					if err != nil {
 						return common.NewExitError(err, codes.ErrorGeneric)
 					}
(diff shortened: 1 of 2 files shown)


matheuscscp (approved):
Nice! LGTM

hiddeco (approved):
In a later iteration, we should try to get rid of the global variable.

For now, as this code should not be consumed by anyone else other than the person running the binary, it is okay by margin as it should not result in e.g. concurrency issues.

felixfontein:
@matheuscscp @hiddeco thanks for reviewing and for the improvements!

lkraav:
I ran into this and I'm not sure it makes sense to force specifically `sops.yaml`.

Literally everything and everybody (in public software domain) supports dual extensions.

Let's say a repo, like ours, has chosen to standardize on one, in this case `.yml`.

All repo tooling has been configured to lint and process this single extension to force internal alignment.

Now SOPS tries to require supporting dual extensions at the last mile, for no visibly good reason. What's the win?

This PR already makes a solid move introducing `alternateConfigName` discovery.

Why not just wire fully into using it and leave end-users choose whatever convention fits them best?

Until then, now a bunch of teams have to type `sops --config .sops.yml` or spam manifests with wrapper commands? Again.. what's the win in all this?

Regards L

felixfontein:
> Literally everything and everybody (in public software domain) supports dual extensions.

I am aware of other software which only accepts one of `yaml` and `yml` for its config files. One such example is ansible, where the configuration file for collections must be called `galaxy.yml`.

> Now SOPS tries to require supporting dual extensions at the last mile, for no visibly good reason. What's the win?

From my experience, this usually causes more confusion. This can also cause subtle security problems if different software implements searching for the config file differently.

> All repo tooling has been configured to lint and process this single extension to force internal alignment.

Using multiple linters for the same rule sounds excessive. And usually linters allow to configure exceptions.

PR #1829, opened 2025-03-31
Remove reserved keyword check from YAML store's `LoadPlainFile()`

### LINKED ISSUE #1828: sops -e -i on an encrypted file returns an error
With sops v3.10.0 running the following command on an already encrypted file returns the following error:

Command:
```
sops -e -i path/to/file
```

Output:

```
Error unmarshalling file: YAML doc used reserved word 'sops'
```

This was not the behavior in previous versions

felixfontein:
This check hid the better (and more general) check in `cmd/sops/encrypt.go` (added in a81f93919c5744fafa180516cea868aa905ebfb0).

The new check was added in adad27e2d4c10d1fb2688bbd3156ef85e1db0735 as part of #1203. I'm not sure why the reserved keyword check was added there, maybe it fixes another thing that the above check does not catch? Do you remember @holiman?

Fixes #1828.

### DIFF
--- a/stores/yaml/store.go
+++ b/stores/yaml/store.go
@@ -329,12 +329,6 @@ func (store *Store) LoadPlainFile(in []byte) (sops.TreeBranches, error) {
 		if err != nil {
 			return nil, fmt.Errorf("Error unmarshaling input YAML: %s", err)
 		}
-		// Prevent use of reserved keywords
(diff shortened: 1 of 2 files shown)


hiddeco (approved):

felixfontein:

PR #1838, opened 2025-04-13
Allow injecting custom HTTP client for AWS, Azure, GCP and HashiCorp Vault

### LINKED ISSUE #1837: Feature request: Support configuring an HTTP/S proxy for the AWS, Azure, GCP and HashiCorp master keys
Hi 👋 

In CNCF Flux we support configuring an HTTP/S proxy for interacting with cloud providers and external systems across most of our APIs, one of the few APIs that do not support proxy yet is the Kustomization API where we allow users to decrypt SOPS-encrypted secrets with KMS services from the three major cloud providers.

In order to support proxy for the Kustomization API we would like the AWS, Azure and GCP SOPS master keys to support the configuration of an HTTP/S proxy 🙏

matheuscscp:
Closes: #1837

### DIFF
--- a/hcvault/keysource.go
+++ b/hcvault/keysource.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"path"
@@ -71,6 +72,8 @@ type MasterKey struct {
 	// Token.ApplyToMasterKey. If empty, the default client configuration
 	// is used, before falling back to the token stored in defaultTokenFile.
 	token string
+	// httpClient is used to override the default HTTP client used by the Vault client.
+	httpClient *http.Client
 }
 
 // NewMasterKeysFromURIs creates a list of MasterKeys from a list of Vault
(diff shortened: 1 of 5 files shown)


hcvault/keysource.go
```diff
@@ -128,12 +131,23 @@ func NewMasterKey(address, enginePath, keyName string) *MasterKey {
 	return key
 }
 
+// HTTPClient is a wrapper around http.Client used for configuring the
+// Vault client.
+type HTTPClient struct {
```
hiddeco, line 136: This appears to be missing a constructor, rendering it unusable for external consumers.

kms/keysource.go
```diff
@@ -233,6 +236,17 @@ func (c CredentialsProvider) ApplyToMasterKey(key *MasterKey) {
 	key.credentialsProvider = c.provider
 }
 
+// HTTPClient is a wrapper around http.Client used for configuring the
+// AWS KMS client.
+type HTTPClient struct {
```
hiddeco, line 241: This appears to be missing a constructor, rendering it unusable for external consumers.

matheuscscp: lol good catch sorry :facepalm:

hcvault/keysource.go
matheuscscp: lol good catch sorry :facepalm:

hiddeco (approved):

PR #1860, opened 2025-05-23
Update authors in main.go

jvehent:
I think it's time :)

Thank you so much to the maintainers who have kept this project going for so many years. You're doing fantastic work!


hiddeco (commented):
Hehe, that's long due indeed. Thank you (for everything)!

PR #1865, opened 2025-06-01
Add OCI KMS as provider

hrvolapeter:
Add functionality for Oracle Cloud (OCI) and it's KMS.

Implements authentication using DEFAULT profile which is read from ~/.oci/config file


felixfontein:
Related: #1226

PR #1876, opened 2025-06-22
Add "--value-file" option to "sops set [...]"

bjornfor:
This allows running "sops set [...]" without leaking secrets in process
listings.

Fixes https://github.com/getsops/sops/issues/729

I'm not a go programmer, please review carefully :-)

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -1385,6 +1385,10 @@ func main() {
 					Name:  "output-type",
 					Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the input file's extension to determine the output format",
 				},
+				cli.BoolFlag{
+					Name:  "value-file",
+					Usage: "treat 'value' as a file to read the actual value from (avoids leaking secrets in process listings)",
+				},
 				cli.IntFlag{
 					Name:  "shamir-secret-sharing-threshold",
 					Usage: "the number of master keys required to retrieve the data key with shamir",
@@ -1430,7 +1434,18 @@ func main() {
 					return common.NewExitError("Invalid set index format", codes.ErrorInvalidSetFormat)
 				}
 
-				value, err := jsonValueToTreeInsertableValue(c.Args()[2])
+				var data string
+				if c.Bool("value-file") {
+					filename := c.Args()[2]
+					content, err := os.ReadFile(filename)
+					if err != nil {
+						return toExitError(err)
+					}
+					data = string(content)
+				} else {
+					data = c.Args()[2]
+				}
+				value, err := jsonValueToTreeInsertableValue(data)
 				if err != nil {
 					return toExitError(err)
 				}
--- a/functional-tests/src/lib.rs
+++ b/functional-tests/src/lib.rs
@@ -491,6 +491,54 @@ bar: baz",
     }
 
     #[test]
+    fn set_json_file_insert_with_value_file() {
+        let file_path = prepare_temp_file(
+            "test_set_json_file_insert_with_value_file.json",
+            r#"{"a": 2, "b": "ba"}"#.as_bytes(),
+        );
+        let value_file = prepare_temp_file("insert_value_file.json", r#"{"cc": "ccc"}"#.as_bytes());
+        assert!(
+            Command::new(SOPS_BINARY_PATH)
+                .arg("encrypt")
+                .arg("-i")
+                .arg(file_path.clone())
+                .output()
+                .expect("Error running sops")
+                .status
+                .success(),
+            "sops didn't exit successfully"
+        );
+        let output = Command::new(SOPS_BINARY_PATH)
+            .arg("set")
+            .arg("--value-file")
+            .arg(file_path.clone())
+            .arg(r#"["c"]"#)
+            .arg(value_file.clone())
+            .output()
+            .expect("Error running sops");
+        println!(
+            "stdout: {}, stderr: {}",
+            String::from_utf8_lossy(&output.stdout),
+            String::from_utf8_lossy(&output.stderr)
+        );
+        assert!(output.status.success(), "sops didn't exit successfully");
+        let mut s = String::new();
+        File::open(file_path)
+            .unwrap()
+            .read_to_string(&mut s)
+            .unwrap();
+        let data: Value = serde_json::from_str(&s).expect("Error parsing sops's JSON output");
+        if let Value::Mapping(data) = data {
+            let a = data.get(&Value::String("c".to_owned())).unwrap();
+            if let &Value::Mapping(ref a) = a {
+                assert_encrypted!(&a, Value::String("cc".to_owned()));
+                return;
+            }
+        }
+        panic!("Output JSON does not have the expected structure");
+    }
+
+    #[test]
     fn set_yaml_file_update() {
         let file_path = prepare_temp_file(
             "test_set_update.yaml",


bjornfor:
About the CI failure; `make test` succeeds on my machine.

I amended the commit to include sign-off. When is CI going to be run again? (It's been 8+ hours.)

felixfontein:
> About the CI failure; `make test` succeeds on my machine.

That runs the unit tests; these also pass in CI. What fails are the functional tests (`make functional-tests`).

> When is CI going to be run again? (It's been 8+ hours.)

When someone with appropriate rights presses the buttons. Please remember that we are volunteers who do this in their free time.

bjornfor:
> > About the CI failure; `make test` succeeds on my machine.
> 
> That runs the unit tests; these also pass in CI. What fails are the functional tests (`make functional-tests`).

Ok, thanks. I can reproduce the `make functional-tests` errors locally, so I can start looking at them now.

> > When is CI going to be run again? (It's been 8+ hours.)
> 
> When someone with appropriate rights presses the buttons. Please remember that we are volunteers who do this in their free time.

Oh, sorry, I thought it was automatic, and given the short time spent on "make test" locally I was surprised about 8+ hour latency on github :see_no_evil:

bjornfor:
Functional tests are passing locally now (except for `publish_json_file_vault` and `publish_json_file_vault_version_1` which fail in main branch for me).

I'll try to create a functional test for the new "--value-file" option tomorrow.

bjornfor:
Hm. The first full `make functional-tests` fails on my new test. Then, running `(cd functional-tests && cargo test set_json_file_insert)` allows my new test, and the test I based it upon, to succeed. And after that `make functional-tests` works again.

Where's the state that makes the test non-deterministic? How to get rid of it?

bjornfor:
> Where's the state that makes the test non-deterministic? How to get rid of it?

It seems test order is random, and some orders work, some does not:

* `(cd functional-tests && cargo test set_json_file_insert)` seems to always work.
* `(cd functional-tests && cargo test)` sometimes fails.

I based the new test, `set_json_file_insert_with_value_file`, on the existing `set_json_file_insert` and I'm surprised to see this difference.

Actually, it might be more nuanced than test _order_: I ran ` for i in $(seq 1000); do echo "i=$i"; (cd functional-tests && cargo test set_json_file_insert); echo $? >> results.txt; done` and got 8/1000 failures.

bjornfor:
Here's the output of a failing/racy run:

```
i=930                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.02s                                                                                                                                                                                                                                                                                                                                                                                                                      
     Running unittests src/lib.rs (target/debug/deps/functional_tests-f17a3fe490e958b4)                                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
running 2 tests                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
test tests::set_json_file_insert ... FAILED                                                                                                                                                                                                                                                                                                                                                                                                                                                   
test tests::set_json_file_insert_with_value_file ... FAILED                                                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
failures:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
---- tests::set_json_file_insert stdout ----                                                                                                                                                                                                                                                                                                                                                                                                                                                  
stdout: , stderr: Error unmarshalling input json: invalid character '}' after top-level value                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
thread 'tests::set_json_file_insert' panicked at src/lib.rs:476:9:                                                                                                                                                                                                                                                                                                                                                                                                                            
sops didn't exit successfully                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace                                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
---- tests::set_json_file_insert_with_value_file stdout ----                                                                                                                                                                                                                                                                                                                                                                                                                                  
stdout: , stderr: Error unmarshalling input json: invalid character '}' after top-level value                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
thread 'tests::set_json_file_insert_with_value_file' panicked at src/lib.rs:522:9:                                                                                                                                                                                                                                                                                                                                                                                                            
sops didn't exit successfully                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
failures:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
    tests::set_json_file_insert                                                                                                                                                                                                                                                                                                                                                                                                                                                               
    tests::set_json_file_insert_with_value_file                                                                                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 36 filtered out; finished in 0.03s                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
error: test failed, to rerun pass `--lib`                                                                                                                                                                                                                                                                                                                                                                                                                                                     
```

felixfontein:
Using different filenames from the test you modified can help.

bjornfor:
> Using different filenames from the test you modified can help.

Aha, `cargo test` runs tests [in parallel with threads](https://doc.rust-lang.org/book/ch11-02-running-tests.html)! Thanks! It all makes sense now.

The latest push fixes the tests, by using a unique file name.

cmd/sops/main.go
```diff
@@ -1385,6 +1385,10 @@ func main() {
 					Name:  "output-type",
 					Usage: "currently json, yaml, dotenv and binary are supported. If not set, sops will use the input file's extension to determine the output format",
 				},
+				cli.BoolFlag{
+					Name:  "value-file",
+					Usage: "treat 'value' as a file to read the actual value from (avoids leaking secrets in process listings)",
+				},
```
felixfontein, line 1391: How about adding a (mutually exclusive) further option that reads from stdin instead of a file? That would provide a platform independent way of reading from stdin (`/dev/stdin` doesn't work on Windows).

felixfontein: I personally think it's better to make things explicit, and let the user explicitly ask for reading from stdin, instead of implicitly just doing that and confusing users which forget the value argument and then wonder why the program is "hanging".

bjornfor: I added "treat - as stdin", and rebased.

bjornfor:
Linter issue fixed.

cmd/sops/main.go
felixfontein: Sorry for the late reply; I think having a dedicated option is better than using `-`. (We had a similar discussion in https://github.com/getsops/sops/issues/739#issuecomment-2343790721; after reading https://github.com/FiloSottile/age/issues/494 I prefer a separate option. Magic can always lead to trouble, including security problems, and treating `-` different from other files would be some magic.)

felixfontein:
@bjornfor thanks a lot for your contribution!

felixfontein:
I created #1894 for `--value-stdin`.

bjornfor:
@felixfontein: Thanks!

PR #1880, opened 2025-06-24
Resolves #1864. Adds Native List as an option for configuring keys. 

### LINKED ISSUE #845: Feature request: support arrays for `age` and `pgp` fields in configuration

### LINKED ISSUE #1864: Allow interpret keys as native list, not just CSV string
Currently keys are treated as CSV strings, ex 
``` yaml
creation_rules:
    - pgp: >-
        85D77543B3D624B63CEA9E6DBC17301B491B3F21,
        FBC7B9E2A4F9289AC0C1D4843D16CEE4A27381B4
```
I've been bitten by two problems recurringly with this format.

1. Comments aren't allowed in yaml block scalar form. It's hard to add metadata (ex, this is the semantic identity of this key), without comments.
2. There's an annoying error when there's a trailing comma on the last element, "" isn't a valid key. This could ofc be itself fixed... but it's symptomatic of not using a native list format, but deferring to a secondary parser and syntax. This is especially the case when the keys might be auto generated, moved around, etc...

Ex
``` yaml
creation_rules:
    - pgp:
        - 85D77543B3D624B63CEA9E6DBC17301B491B3F21 # name@email.com
        - FBC7B9E2A4F9289AC0C1D4843D16CEE4A27381B4 # server_XZY
```
Note: this shouldn't be a breaking change, so long as its a loosening of the restrictions; ie `typeof(pgp) == string | list<string>`

lucqui:
Resolves #1864 and resolves #845. Adds Native List as an option for configuring keys in addition to the trailing comma option already given.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -190,36 +190,36 @@ type creationRule struct {
 }
 
 // Helper methods to safely extract keys as []string
-func (c *creationRule) GetKMSKeys() []string {
+func (c *creationRule) GetKMSKeys() ([]string, error) {
 	return parseKeyField(c.KMS)
 }
 
-func (c *creationRule) GetAgeKeys() []string {
+func (c *creationRule) GetAgeKeys() ([]string, error) {
 	return parseKeyField(c.Age)
 }
 
-func (c *creationRule) GetPGPKeys() []string {
+func (c *creationRule) GetPGPKeys() ([]string, error) {
 	return parseKeyField(c.PGP)
 }
 
-func (c *creationRule) GetGCPKMSKeys() []string {
+func (c *creationRule) GetGCPKMSKeys() ([]string, error) {
 	return parseKeyField(c.GCPKMS)
 }
 
-func (c *creationRule) GetAzureKeyVaultKeys() []string {
+func (c *creationRule) GetAzureKeyVaultKeys() ([]string, error) {
 	return parseKeyField(c.AzureKeyVault)
 }
 
-func (c *creationRule) GetVaultURIs() []string {
+func (c *creationRule) GetVaultURIs() ([]string, error) {
 	return parseKeyField(c.VaultURI)
 }
 
 // Utility function to handle both string and []string
-func parseKeyField(field interface{}) []string {
+func parseKeyField(field interface{}) ([]string, error) {
 	switch v := field.(type) {
 	case string:
 		if v == "" {
-			return []string{}
+			return []string{}, nil
 		}
 		// Existing CSV parsing logic
 		keys := strings.Split(v, ",")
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -170,23 +170,78 @@ type destinationRule struct {
...
+func (c *creationRule) GetAzureKeyVaultKeys() []string {
+	return parseKeyField(c.AzureKeyVault)
+}
+
+func (c *creationRule) GetVaultURIs() []string {
+	return parseKeyField(c.VaultURI)
+}
+
+// Utility function to handle both string and []string
+func parseKeyField(field interface{}) []string {
```
felixfontein, line 218: I think it would be best if this function returns an error if the input isn't either a string, a list of strings, or `nil`. That way users get more helpful feedback in case of errors, instead of having to guess they have some invalid entry in the config file based on strange follow-up errors.

felixfontein (commented):
Thank you very much for picking this topic up! I think this is already looking great (and it worked well during testing)!

config/config.go
lucqui: Oh yes that makes a lot of sense! I'll make that change tonight when I get to my laptop

config/config_test.go
```diff
@@ -718,3 +718,40 @@ func TestLoadConfigFileWithVaultDestinationRules(t *testing.T) {
...
+    hc_vault_transit_uri:
+      - "https://vault.example.com:8200/v1/transit/keys/key1"
+`)
+	conf, err := parseCreationRuleForFile(parseConfigFile(sampleConfigWithNativeKeyLists, t), "/conf/path", "native_list_test", nil)
+	assert.Nil(t, err)
+	if conf == nil {
+		t.Fatal("Expected configuration but got nil")
+	}
+
+	assert.True(t, len(conf.KeyGroups) > 0)
```
felixfontein, line 744: Shouldn't this be
```suggestion
	assert.True(t, len(conf.KeyGroups) == 1)
```
?

lucqui:
@felixfontein Awesome thanks for the review! I'll get these changes in tomorrow!

felixfontein:
Ping @lucqui

lucqui:
> Ping @lucqui

Got distracted with some work things. Let me wrap this up so we can merge it

felixfontein:
Something seems to be wrong with the commit history. There are a lot of commits in it that do not belong there. Can you rebase to remove them?

felixfontein:
The code changes look good to me; I'll do some manual testing later today (didn't have time for that now).

Edit: manual testing also worked smooth! Looks great :)

felixfontein:
I rebased the branch to get rid of the stray commits. While at that I also removed the `@` from the commit message, since that results in me getting a notification (and thus an email) every time the commit is pushed somewhere ;)

felixfontein (approved):

felixfontein:
@lucqui thanks a lot for implementing this!

PR #1885, opened 2025-07-01
Correct Windows path to store keys.txt

EshemMimi:
The change displays the right path when README.rst is rendered


felixfontein (approved):
Thanks for fixing this!

PR #1894, opened 2025-07-23
Add `--value-stdin` option to `sops set`

### LINKED ISSUE #729: Setting values from stdin directly
We use sops extensively with terraform and ansible.

Often we have a scenario where we want to generate some secret with a tool and add it to a sops encrypted file under a key immediately.

For example: generating ssh keys, generate rsa keys.

The current workflow looks like:

```console
$ openssl genrsa 4096  # key is printed to screen
# copy key to clipboard
$ sops path/to/file.sops.yml
# edit file and paste the key
```

The desired workflow would look like:

```console
$ openssl genrsa 4096 | sops --set '["app1"]["key"]' path/to/file.sops.yml
# generate another one and store under a different key!
$ openssl genrsa 4096 | sops --set '["app2"]["key"]' path/to/file.sops.yml
```

Less user friendly but equally useful would be:
```console
$ openssl genrsa 4096 | sops --set '["app1"]["key"]' --value-file /dev/stdin path/to/file.sops.yml
# generate another one and store under a different key!
$ openssl genrsa 4096 | sops --set '["app2"]["key"]' --value-file /dev/stdin path/to/file.sops.yml
```

felixfontein:
Follow-up to #1876.

Fixes #729.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	encodingjson "encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


felixfontein:
@sabre1041 thanks again!

cmd/sops/main.go
```diff
@@ -1387,7 +1388,11 @@ func main() {
 				},
 				cli.BoolFlag{
 					Name:  "value-file",
-					Usage: "treat 'value' as a file to read the actual value from (avoids leaking secrets in process listings)",
+					Usage: "treat 'value' as a file to read the actual value from (avoids leaking secrets in process listings). Mutually exclusive with --value-stdin",
+				},
+				cli.BoolFlag{
+					Name:  "value-stdin",
+					Usage: "treat 'value' as a file to read the actual value from (avoids leaking secrets in process listings). Mutually exclusive with --value-file",
```
bjornfor, line 1395: This help text is wrong: `--value-stdin` doesn't take a value, only `--value-file` does that.

felixfontein: Neither option takes a value. They both modify what the `value` parameter of `sops set` means.

PR #1895, opened 2025-07-23
Fix docs: mention all stores instead of just YAML, JSON, and BINARY

### LINKED ISSUE #1842: Misleading README.rst regarding .env file partial encryption support
Hi!

So the README.rst states:

```rst
Encrypting only parts of a file
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Note: this only works on YAML and JSON files, not on BINARY files.

By default, SOPS encrypts all the values of a YAML or JSON file and leaves the
keys in cleartext. In some instances, you may want to exclude some values from
being encrypted. This can be accomplished by adding the suffix **_unencrypted**
to any key of a file. When set, all values underneath the key that set the
**_unencrypted** suffix will be left in cleartext.
```

However when testing, we found that ENV files are supported as well.

Is this simply a case that the someone forgot to update the docs or?

felixfontein:
There are a few places where only YAML and JSON were mentioned (next to BINARY).

Fixes #1842.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1477,7 +1477,7 @@ original file after encrypting or decrypting it.
 Encrypting binary files
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-SOPS primary use case is encrypting YAML and JSON configuration files, but it
+SOPS primary use case is encrypting YAML, JSON, ENV, and INI configuration files, but it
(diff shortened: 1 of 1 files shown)


felixfontein:

PR #1898, opened 2025-07-23
Collect age identity loading errors and only report if decryption failed

### LINKED ISSUE #1843: Bug: upgrade to v3.10.* causes "could not create encrypted SSH identity: RSA key size is too small" error
I've upgraded sops binary from v3.8.1 to v3.10.2. Without changing any encrypted file, my decrypt attempts started failing due to this one error:
```bash
$ SOPS_AGE_KEY_FILE="$HOME/.config/sops/age/my-prod-keys.txt" sops -d my-prod-shared-secrets.yaml

Group 0: FAILED
  age1gzjg67ckfw0gqa8y60dq9uxcmr2hgjwlzpu55wyu9qjuq4h54yyqw3jf7t: FAILED
    - | failed to load age identities: failed to get SSH identity:
      | could not create encrypted SSH identity: RSA key size is too
      | small

Recovery failed because no master key was able to decrypt the file. In
order for SOPS to recover the file, at least one key has to be successful,
but none were.
```

After some investigating, I realized that the error was due to my very old SSH-RSA key. But files encrypted using `age`-keys, not SSH.

It seems that this behavior is caused by [this PR](https://github.com/getsops/sops/pull/1692), which try to parse SSH key even it is **NOT NEEDED** for encryption/decryption.

As a workaround I suggest something like:
```bash
SOPS_AGE_SSH_PRIVATE_KEY_FILE="$HOME/.ssh/my_ed25519" SOPS_AGE_KEY_FILE="$HOME/.config/sops/age/my-prod-keys.txt" sops -d my-prod-shared-secrets.yaml
```

Or just create new `~/.ssh/id_ed25519` SSH key (`ssh-keygen -t ed25519 -C "your_email@example.com"`) and you can omit `SOPS_AGE_SSH_PRIVATE_KEY_FILE` in command above.

felixfontein:
Fixes #1843.

Maybe also *all* SSH identities should be loaded, and not the first one found. This needs some more adjustments to the tests, though, since suddenly the Age unit tests depend on all keys of interest in `~/.ssh`. Right now these can be ignored by setting `SOPS_AGE_SSH_PRIVATE_KEY_FILE` accordingly.

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -94,6 +94,18 @@ func MasterKeysFromRecipients(commaSeparatedRecipients string) ([]*MasterKey, er
 	return keys, nil
 }
 
+// errSet is a collection of captured errors.
+type errSet []error
+
+// Error joins the errors into a "; " separated string.
+func (e errSet) Error() string {
+	str := make([]string, len(e))
+	for i, err := range e {
+		str[i] = err.Error()
+	}
+	return strings.Join(str, "; ")
+}
+
 // MasterKeyFromRecipient takes a Bech32-encoded age public key, parses it, and
 // returns a new MasterKey.
 func MasterKeyFromRecipient(recipient string) (*MasterKey, error) {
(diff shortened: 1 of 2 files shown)


felixfontein:
Forgot to remark: this is a bit similar to how the PGP keystore operates. It tries to decode with GnuPG and the included OpenPGP client, and only fails if both fail (and then reports all errors).

sabre1041 (approved):
LGTM

felixfontein:
@sabre1041 thanks for revieiwng!

PR #1394, opened 2023-12-28
INI, DotEnv stores: shamir_threshold is an integer

felixfontein:
Fixes #857 similarly to #1338.

PR #1899, opened 2025-07-25
Fix Shamir threshold encoding for INI and ENV files

### LINKED ISSUE #837: Shamir is currently broken for INI file types
The resulting error is:
```
json: cannot unmarshal string into Go struct field Metadata.shamir_threshold of type int
```

With the shamir threshold being:
```
shamir_threshold  = 2.000000
```

felixfontein:
Looks like #1394 was inadequate. This fixes it correctly.

Fixes #837.

### DIFF
--- a/stores/flatten.go
+++ b/stores/flatten.go
@@ -248,7 +248,14 @@ func DecodeNonStrings(m map[string]interface{}) error {
 		case string:
 			vInt, err := strconv.Atoi(val)
 			if err != nil {
-				return fmt.Errorf("shamir_threshold is not an integer: %s", err.Error())
+				// Older versions of SOPS stored shamir_threshold as a floating point representation
+				// of the actual integer. Try to parse a floating point number and see whether it
+				// can be converted without loss to an integer.
+				vFloat, floatErr := strconv.ParseFloat(val, 64)
+				vInt = int(vFloat)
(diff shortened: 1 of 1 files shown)


stores/flatten.go
```diff
@@ -274,5 +274,11 @@ func EncodeNonStrings(m map[string]interface{}) {
 		if vInt, ok := v.(int); ok {
 			m["shamir_threshold"] = fmt.Sprintf("%d", vInt)
 		}
+		// FlattenMetadata serializes the input as JSON and then deserializes it.
+		// The JSON unserializer treats every number as a float, so the above 'if'
+		// never applies in that situation.
+		if vFloat, ok := v.(float64); ok {
+			m["shamir_threshold"] = fmt.Sprintf("%g", vFloat)
```
hiddeco, line 281: It would probably be safer to do something like: 

```suggestion
			m["shamir_threshold"] = fmt.Sprintf("%.0f", vFloat)
```

The problem is that otherwise `DecodeNonStrings` wouldn't be able to parse e.g. `1234567.0`, as it would become `1.234567e+06`.

felixfontein: shamir_threshold should never be that large, but this is still better since it avoids making the result look like a floating point number!

hiddeco (approved):

felixfontein:

PR #1903, opened 2025-08-01
Ensure temporary file for editing is only read-writable by owner

### LINKED ISSUE #1902: Permission issues with temporary files
The temporary file generated in edit mode has permissions `rw-r--r--`, but perhaps it should be `rw-------`?

felixfontein:
Fixes #1902.


felixfontein:
@sabre1041 thanks for reviewing this one as well!

PR #1919, opened 2025-08-21
feat(azkv): Skipping key-version will get latest key

### LINKED ISSUE #1924: Allow --azure-kv to encrypt using latest version of key without providing the version
Allow passing an Azure Key Vault key URL without a version, e.g. https://my-az-kv.vault.azure.net/keys/key-name

Today SOPS forces specifying /{version}, which creates churn on key rotation.

Azure return latest version if only key-name is requested.

daogilvie:
Firstly: thanks for maintaining such a great tool! It's made my life a lot easier.

Hopefully this PR is helpful in some way; I checked open/closed issues and PRs for something similar being rejected, and couldn't see it.

Anyway, what's the change? I've edited the AZKV implementation to support not including a version in the URL. I.e this makes `https://${VAULT_URL}/keys/${KEY_NAME}/` a valid AZKV URL.

This is a supported behaviour for Azure Key Vault, and is very helpful when it comes to doing rotation of those keys. Updated versions will automatically get fetched by sops for new files with no changes to the config in the repo. This brings AZKV in line with sops' AWS/GCP integrations, which do not require you to specify a key version.

EDIT adding clarification after @felixfontein's comment:
If you _don't_ specify a version in the URL, then sops will use the azkey client (already bundled, of course), to retrieve the specific version of the key, and puts that into the struct instead. I.e even with no version supplied, the [NewMasterKeyFromURL function](https://github.com/getsops/sops/pull/1919/files#diff-17057699dd441d9c5ab351e49afb76566f698c8e4e812e39a052ec8b9f90031dR80) always returns a version, so any files edited/encrypted will still have the specific version stored in the metadata.

>[!NOTE]
>I am not sure how to test this in the automated test suites, as I can't see any comparable test in [azkv/keysource_test.go](https://github.com/getsops/sops/blob/main/azkv/keysource_test.go) that would mock or otherwise inject the az client.
>I am more than happy to be told this is not something sops should even support, or that I need to redo some or all of the work.
>I don't really use golang much day-to-day, so I may be completely missing out on some sensible approach or pattern.

Fixes #1924

## Manual Tests
When no version set in URL, SOPS...
1. **SHOULD** fetch latest version when encrypting a file, and save the version in the file ✅ (seems to parse the URL multiple times though?)
2. **SHOULD NOT** fetch the latest version when editing the file ✅ 
3. **SHOULD NOT** fetch the latest version when decrypting a file ✅ 
4. **SHOULD** fetch the latest and update the file when using `updatekeys` ✅ 

When a version is provided in the URL, SOPS...
1. **SHOULD NOT** fetch it via the client ✅

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -378,6 +378,11 @@ a key. This has the following form::
 
     https://${VAULT_URL}/keys/${KEY_NAME}/${KEY_VERSION}
 
+You can omit the version, and have just a trailing slash, and this will use
+whatever the latest version of the key is::
+
+    https://${VAULT_URL}/keys/${KEY_NAME}/
+
 To create a Key Vault and assign your service principal permissions on it
 from the commandline:
 
@@ -400,6 +405,10 @@ from the commandline:
 Now you can encrypt a file using::
 
     $ sops encrypt --azure-kv https://sops.vault.azure.net/keys/sops-key/some-string test.yaml > test.enc.yaml
+
+or, without the version::
+
+    $ sops encrypt --azure-kv https://sops.vault.azure.net/keys/sops-key/ test.yaml > test.enc.yaml
 
 And decrypt it using::
 
--- a/azkv/keysource.go
+++ b/azkv/keysource.go
@@ -79,7 +79,7 @@ func NewMasterKey(vaultURL string, keyName string, keyVersion string) *MasterKey
 // MasterKey. The URL format is {vaultUrl}/keys/{keyName}/{keyVersion}.
 func NewMasterKeyFromURL(url string) (*MasterKey, error) {
 	url = strings.TrimSpace(url)
-	re := regexp.MustCompile("^(https://[^/]+)/keys/([^/]+)(/[^/]+)?$")
+	re := regexp.MustCompile("^(https://[^/]+)/keys/([^/]+)(/[^/]*)?$")
 	parts := re.FindStringSubmatch(url)
 	if len(parts) < 3 {
 		return nil, fmt.Errorf("could not parse %q into a valid Azure Key Vault MasterKey %v", url, parts)
--- a/azkv/keysource.go
+++ b/azkv/keysource.go
@@ -79,12 +79,21 @@ func NewMasterKey(vaultURL string, keyName string, keyVersion string) *MasterKey
 // MasterKey. The URL format is {vaultUrl}/keys/{keyName}/{keyVersion}.
 func NewMasterKeyFromURL(url string) (*MasterKey, error) {
 	url = strings.TrimSpace(url)
-	re := regexp.MustCompile("^(https://[^/]+)/keys/([^/]+)/([^/]+)$")
+	re := regexp.MustCompile("^(https://[^/]+)/keys/([^/]+)(/[^/]+)?$")
 	parts := re.FindStringSubmatch(url)
 	if len(parts) < 3 {
-		return nil, fmt.Errorf("could not parse %q into a valid Azure Key Vault MasterKey", url)
+		return nil, fmt.Errorf("could not parse %q into a valid Azure Key Vault MasterKey %v", url, parts)
 	}
-	return NewMasterKey(parts[1], parts[2], parts[3]), nil
+	// Blank key versions are supported in Azure Key Vault, as they default to the latest
+	// version of the key. We need to put the actual version in the sops metadata block though
+	var key *MasterKey
+	if len(parts[3]) > 1 {
+		key = NewMasterKey(parts[1], parts[2], parts[3][1:])
+	} else {
+		key = NewMasterKey(parts[1], parts[2], "")
+	}
+	err := key.ensureKeyHasVersion(context.Background())
+	return key, err
 }
 
 // MasterKeysFromURLs takes a comma separated list of Azure Key Vault URLs,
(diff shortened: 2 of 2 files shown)


felixfontein:
Not sure whether I understand this correctly, but isn't always using the latest version not a bad idea? If you encrypt a file with a specific key, and then later try to decrypt after the key has been rotated, with your change SOPS would try to use the latest key to decrypt, wouldn't that not work, because it needs the previous one?

So I guess it would be OK if `.sops.yaml` can skip the version, but once a file is encrypted, the key **including the version** used for encryption must be stored in the encrypted file?

daogilvie-9fin:
> Not sure whether I understand this correctly, but isn't always using the latest version not a bad idea? If you encrypt a file with a specific key, and then later try to decrypt after the key has been rotated, with your change SOPS would try to use the latest key to decrypt, wouldn't that not work, because it needs the previous one?
> 
> So I guess it would be OK if `.sops.yaml` can skip the version, but once a file is encrypted, the key **including the version** used for encryption must be stored in the encrypted file?

Thanks for having a look @felixfontein! Sorry — I should have been clearer.

Most of the lines of code added here are actually to correctly identify the version of the key and store that in the metadata in the yaml of the file. `sops edit my.sops.yaml` for example, will have the version in the metadata correctly, so subsequent `sops decrypt` or `sops edit` calls work.
I tested this by editing a file into existence, altering `.sops.yaml` to refer to a different key, and checking the decrypt/edit command still works.

It is entirely possible that my limited understanding of sops means I've broken other use-cases though.

felixfontein:
Ah, that's good to hear! I think that makes sense.

One thing I'm worrying about is that retrieving the version might be done too often, i.e. not only when it's actually needed. Did you try adding debug statements which log retrieval and check what happens during various kind of operations (encrypt, decrypt, edit, update keys, ...)?

Also did you test whether `updatekeys` works (I would expect it to update the key's version)?

README.rst
```diff
@@ -401,6 +406,10 @@ Now you can encrypt a file using::
 
     $ sops encrypt --azure-kv https://sops.vault.azure.net/keys/sops-key/some-string test.yaml > test.enc.yaml
 
+or, without the version:
```
felixfontein, line 409: The double colons are a shortcut that the next following lines are a code-block (and not a block quote):
```suggestion
or, without the version::
```

felixfontein (commented):
Generally, you need to sign-off your commits (to accept the [DCO](https://github.com/getsops/sops/blob/main/DCO)), otherwise we cannot merge this. (On the command line, you can do `git commit --sign-off --amend` to modify the commit, and then force-push.)

daogilvie-9fin:
> Ah, that's good to hear! I think that makes sense.
> 
> One thing I'm worrying about is that retrieving the version might be done too often, i.e. not only when it's actually needed. Did you try adding debug statements which log retrieval and check what happens during various kind of operations (encrypt, decrypt, edit, update keys, ...)?

I did not, no! That is a very fair question. "When it's actually needed" here being when we are encrypting/decrypting and the version cannot be sourced from anywhere else? I.e it's not in the url, it's not in the metadata somewhere?
 
> Also did you test whether `updatekeys` works (I would expect it to update the key's version)?

That is also something I did not test, no — a big oversight on my part. I'll do that

I did this PR as a side-effect of a rush to get something done at work; I will aim to do this testing this week. After that I will be changing jobs and won't have the pleasure of dealing with Azure any more, thus losing access to my testing environment. I'll redo the commits with sign-off then. Thank you!

felixfontein:
> > Ah, that's good to hear! I think that makes sense.
> > One thing I'm worrying about is that retrieving the version might be done too often, i.e. not only when it's actually needed. Did you try adding debug statements which log retrieval and check what happens during various kind of operations (encrypt, decrypt, edit, update keys, ...)?
> 
> I did not, no! That is a very fair question. "When it's actually needed" here being when we are encrypting/decrypting and the version cannot be sourced from anywhere else? I.e it's not in the url, it's not in the metadata somewhere?

It should only query the version when encrypting a new file (not when editing), or when running updatekey (and then it should replace older versions by the latest version). If it queries the latest version more than once in these cases, that would be OK I guess (at least for now), but if it also queries the version in other situations (like when decrypting) this would be bad. (The encrypted file should always contain AZKV URLs with specific versions, namely the version which can be used to decrypt the key.)

> > Also did you test whether `updatekeys` works (I would expect it to update the key's version)?
> 
> That is also something I did not test, no — a big oversight on my part. I'll do that

Thanks! As I wrote above, the expected behavior would be that it replaces the old version with the latest version of the AZKV URL.

(I think this should 'just work' with your PR, but without testing it... who knows :) )

> I did this PR as a side-effect of a rush to get something done at work; I will aim to do this testing this week.

That would be great!

> After that I will be changing jobs and won't have the pleasure of dealing with Azure any more, thus losing access to my testing environment.

Hehe, sounds good :) I'm lucky that I don't have to use Azure except for AZP in another open source project (and honestly I find the UI/UX there so bad that I'm really wondering why anyone pays money to use that).

> I'll redo the commits with sign-off then. Thank you!

Thank *you* for your contribution!

daogilvie-9fin:
Ok, I've manually tested the behaviours @felixfontein asked for, and it all looks good.
It does seem that the .sops.yaml URL is parsed more than once before the file is actually encrypted, and so the version is fetched each time that happens, but once it's saved in the file then sops only ever fetches the version again when doing `updatekeys`, so that seems to be good. Commit is squashed, rebased and signed off; hopefully ready for code review!
@felixfontein — I altered the behaviour of the URL parsing function slightly from when you last looked, to make it so users can omit the trailing slash in version-less URLs.

r4vi:
@felixfontein if you need someone to keep testing stuff on Azure with regards to this PR - just @ me. I work[ed] with @daogilvie and I still have the misfortune of having access to Azure

felixfontein (approved):

felixfontein:
I'm going to merge this and then rebase #1939; @r4vi would be great if you could test whether that PR/branch reduces the number of `ensureKeyHasVersion()` calls per key to one.

felixfontein:
@daogilvie thanks for your contribution!
@r4vi thanks for helping!

maonat:
@felixfontein when is the next release planned to include this feature?

PR #1928, opened 2025-08-30
Docs: remove paragraph on GPG/PGP keyservers

### LINKED ISSUE #1858: remove keyserver references from docs?
The docs suggest that one can use `keys.openpgp.org` as a keysource https://getsops.io/docs/#specify-a-different-gpg-key-server

However, this functionality seems to be removed in https://github.com/getsops/sops/commit/c6236ada5645b6c08e0d87dbb946fcc6d1cfc1a1

> - Support for fetching keys from servers has been removed. This can be
>  added back if we need to keep it around for a little longer.

could be related to:
https://github.com/getsops/sops/issues/727
https://github.com/getsops/sops/issues/1681

felixfontein:
Fixes #1858.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -846,14 +846,6 @@ Example: place the following in your ``~/.bashrc``
     SOPS_GPG_EXEC = 'your_gpg_client_wrapper'
 
 
-Specify a different GPG key server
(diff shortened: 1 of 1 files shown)


felixfontein:

PR #1929, opened 2025-08-30
Fix converting integers to strings; also improve time.Time formatting

### LINKED ISSUE #1915: INI store: conversion to string destroys integers
Instead of `123`, you get `%!s(int=123)`.

felixfontein:
Before:
```ini
[section]
int       = %!s(int=123)
float     = 1.230000
bool      = true
date      = 2025-01-01 00:00:00 +0000 UTC
timestamp = 2025-01-01 01:02:03 +0000 UTC
string    = this is a string
```

After:
```ini
[section]
int       = 123
float     = 1.230000
bool      = true
date      = 2025-01-01T00:00:00Z
timestamp = 2025-01-01T01:02:03Z
string    = this is a string
```

Fixes #1915.

### DIFF
--- a/stores/ini/store.go
+++ b/stores/ini/store.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


felixfontein:
@getsops/maintainers this change is a breaking change, since it changes the output. For integers, I think this is definitely OK. Timestamps haven't been supported for that long, and I think the new output is a lot better, and I think there it's definitely also OK. For floats, I think the old output was bad since it potentially destroyed the value (like `1.2345E-10` getting turned into `0`). What do you think, is this OK for a new feature release?

felixfontein:
@sabre1041 thanks a lot for reviewing this!

PR #1930, opened 2025-08-30
Allow to configure --enable-local-keyservice and --keyservice through env variables

### LINKED ISSUE #1918: [Feature Request] Global remote keyservice configuration
While most SOPS decryption operations require no configuration (beyond what is contained in the file being decrypted), decryption operations that use a remote keyservice are notable exception. This isn't an insurmountable problem when executing the `sops` command directly, but more and more software is starting to use SOPS internally, typically without being able to specify extra flags for the underlying `sops` command. While some software (e.g. Ansible's `community.sops` collection) does provide a way to configure a remote keyservice, there are other examples (e.g. Terragrunt's `sops_decrypt_file` function) that do not. Moreover, having to specify the `--keyservice` flag (often along with the `--enable-local-keyservice=false` flag) on every single `sops` invocation starts to get annoying (and defining an alias or `just` recipe feels like an anti-pattern, and often won't solve the internal-use issues described above).

Ideally, there'd be a way to configure a remote keyservice on a more global basis. It think even something as simple as `SOPS_KEYSERVICE` (etc.) environment variables would go a long way to making the remote keyservice functionality more consistently useful.

Thanks!

felixfontein:
Fixes #1918.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -78,12 +78,14 @@ func main() {
 
 	keyserviceFlags := []cli.Flag{
 		cli.BoolTFlag{
-			Name:  "enable-local-keyservice",
-			Usage: "use local key service",
+			Name:   "enable-local-keyservice",
+			Usage:  "use local key service",
+			EnvVar: "SOPS_ENABLE_LOCAL_KEYSERVICE",
 		},
 		cli.StringSliceFlag{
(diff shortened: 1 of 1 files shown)


sabre1041 (approved):
LGTM

felixfontein:

PR #1936, opened 2025-09-07
fix: correct destination validation logic to detect all conflicts

bruce-szalwinski-he:
The original validation logic used AND (&&) operators which only detected conflicts when ALL THREE destination types were specified. This meant invalid two-way conflicts like s3_bucket + gcs_bucket would silently pass validation.

Fixed by implementing proper counting-based validation that rejects any configuration with more than one destination type specified.

Added comprehensive test coverage for all conflict scenarios:
- S3 + GCS conflicts
- S3 + Vault conflicts
- GCS + Vault conflicts
- All three destinations conflicts
- Positive tests for valid single destinations

Fixes a critical configuration validation bug that could lead to unexpected publish behavior with misconfigured .sops.yaml files.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -508,7 +508,18 @@ func parseDestinationRuleForFile(conf *configFile, filePath string, kmsEncryptio
 	}
 
 	var dest publish.Destination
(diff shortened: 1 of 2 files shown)


felixfontein:
@bruce-szalwinski-he thanks for your contribution!

PR #1944, opened 2025-09-11
Fix mention of macOS XDG_CONFIG_HOME fallback

### LINKED ISSUE #1943: No fallback to old config dir in macOS
This has stopped working pretty recently, as instructed by docs my sops key is on `Library/Application Support...` on macOS, since I believe XDG_CONFIG_HOME was not documented as working for Mac at the time, nevertheless there's no fallback searching for this path anymore it appears:
```
 - | failed to load age identities: failed to open file: open
      | /Users/xored/.config/sops/age/keys.txt: no such file or
      | directory
 ```

It only attempts to read from `XDG_CONFIG_HOME/...`.
Either that or sops is not detecting I'm on macOS (very plausible since it's macOS 26 RC day, but it has worked throughout all betas).

sops version: 3.10.2

felixfontein:
Fixes #1943.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -231,7 +231,7 @@ configuration directory.
 - **macOS**
 
   - Looks for ``keys.txt`` in ``$XDG_CONFIG_HOME/sops/age/keys.txt``;
-  - Falls back to ``$HOME/Library/Application Support/sops/age/keys.txt``.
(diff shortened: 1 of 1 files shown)


felixfontein:

PR #1945, opened 2025-09-11
CI: Build with Go 1.24 and 1.25, release with 1.25

felixfontein:
Ref: https://github.com/getsops/sops/pull/1941#discussion_r2338482230

@getsops/maintainers or should we keep Go 1.23 in the matrix?


hiddeco (approved):
We can drop 1.23 given https://endoflife.date/go.

felixfontein:

Issue #1767, opened 2025-02-23
provide JSON Schema to allow validating config file


KiaraGrouwstra:
**Is your feature request related to a problem? Please describe.**
We now have a [sample config](https://github.com/getsops/sops/blob/main/.sops.yaml), which helps users get going to tweak their own setup. However, we currently do not yet offer static checks to help validate these.

**Describe the solution you'd like**
It would be nice to offer a [JSON Schema](https://json-schema.org/) file for config validation, including by LSP language servers such as [`yaml-language-server`](https://github.com/redhat-developer/yaml-language-server).

**Describe alternatives you've considered**
not make mistakes

**Additional context**
n/a

felixfontein:
I think there was an issue or discussion before where validation of `.sops.yaml` files came up, but I can't find it right now. It would be great to have validation, either as a `sops` subcommand or as a JSON schema (or both), but it would be best if that could be automatically extracted from the existing data structures.

It looks like there is a library which creates JSON schemas from Go data structures: https://github.com/invopop/jsonschema So this should be doable. (Actually both: creating a standalone JSON schema, and creating a config validation subcommand that uses the schema to do the validation.)

felixfontein:
Not directly related, but yaml.v4 will have a `yaml.WithKnownFields()` option that will return an error if an unknown field is found while unmarshaling. This could be used to create a "SOPS config validator" subcommand.

PR #1946, opened 2025-09-12
Start documenting the configuration file format

felixfontein:
I ran out of steam and skipped the destination rules for now. (I also never used them, so I'm not the best person to document them :) )

Related: #1767

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -429,7 +429,7 @@ The simplest way to decrypt data from stdin is as follows:
 
 .. code:: sh
 
-	$ cat encrypted-data | sops decrypt > decrypted-data
+    $ cat encrypted-data | sops decrypt > decrypted-data
 
 By default, ``sops`` determines the input and output format from the provided filename. Since in this case,
 no filename is provided, ``sops`` will use the binary store which expects JSON input and outputs binary data
@@ -440,8 +440,8 @@ the input and output formats by passing ``--input-type`` and ``--output-type`` a
 
 .. code:: sh
 
-	$ cat encrypted-data | sops decrypt --filename-override filename.yaml > decrypted-data
-	$ cat encrypted-data | sops decrypt --input-type yaml --output-type yaml > decrypted-data
+    $ cat encrypted-data | sops decrypt --filename-override filename.yaml > decrypted-data
+    $ cat encrypted-data | sops decrypt --input-type yaml --output-type yaml > decrypted-data
 
 In both cases, ``sops`` will assume that the data you provide is in YAML format, and will encode the decrypted
 data in YAML as well. The second form allows to use different formats for input and output.
@@ -452,7 +452,7 @@ SOPS which filename to use to match creation rules:
 
 .. code:: sh
 
-	$ echo 'foo: bar' | sops encrypt --filename-override path/filename.sops.yaml > encrypted-data
+    $ echo 'foo: bar' | sops encrypt --filename-override path/filename.sops.yaml > encrypted-data
 
 SOPS will find a matching creation rule for ``path/filename.sops.yaml`` in ``.sops.yaml`` and use that one to
 encrypt the data from stdin. This filename will also be used to determine the input and output store. As always,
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -1681,6 +1681,308 @@ Note: these six options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--en
...
+Settings
+********
+
+The following keys configure encryption settings:
+
+* ``shamir_threshold`` (integer, default ``0``): Must be ``0`` (disabled) or an integer greater or equal to 2.
+  Determines the number of key groups from whose one key must be present each to decrypt the file's key.
+
+* ``mac_only_encrypted`` (boolean, default ``false``): If set to ``true``, only encrypted strings will count towards the file's MAC.
+  If set to ``false``, also unencrypted values will be part of the MAC computation.
```
sabre1041, line 1911:
```suggestion
  If set to ``false``, unencrypted values will also be part of the MAC computation.
```

README.rst
```diff
@@ -1681,6 +1681,308 @@ Note: these six options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--en
...
+
+The following keys configure which values in a file are encrypted.
```
sabre1041, line 1913:
```suggestion
The following keys configure the specific values in a file that should be encrypted.
```

README.rst
```diff
@@ -1681,6 +1681,308 @@ Note: these six options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--en
...
+Note that at most one of these keys can be used.
```
sabre1041, line 1914:
```suggestion
Note that at most, one of these keys can be used.
```

README.rst
```diff
@@ -1681,6 +1681,308 @@ Note: these six options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--en
...
+* ``encrypted_comment_regex`` (string): A value or comment is encrypted if it has a preceding comment (or a trailing comment on the same line) matching this regular expression.
+  All other comments and values are not encrypted.
+  The matching comment itself is not encrypted.
+
+Destination rule object
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Not yet documented.
+
+.. TODO!
```
sabre1041, line 1944: Lets remove the TODO line as we already have a callout for the section of the document that is lacking content

sabre1041 (changes requested):
@felixfontein This looks REALLY good. Thank you for taking the time to write it up.

Added a number of suggestions

README.rst
```diff
@@ -1681,6 +1681,308 @@ Note: these six options ``--unencrypted-suffix``, ``--encrypted-suffix``, ``--en
...
+  .. code:: yaml
+
+    - arn: arn:aws:kms:us-west-2:927034868273:key/fe86dd69-4132-404c-ab86-4269956b4500
+      role: arn:aws:iam::927034868273:role/sops-dev-xyz
+      context:
+        Environment: production
+        Role: web-server
+      aws_profile: foo
+
+* ``gcp_kms`` (list of objects): list of GCP KMS ResourceIDs.
```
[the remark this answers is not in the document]
felixfontein: I got the impression from https://github.com/getsops/sops/blob/44e07cd9cc190a4d67f07cecdeaea28e8f97f342/README.rst#24encrypting-using-gcp-kms that `ResourceID` is a fixed term. On the other hand, Google talks about "resource IDs" for example in https://cloud.google.com/kms/docs/getting-resource-ids. So I guess it should be:
```suggestion
* ``gcp_kms`` (list of objects): list of GCP KMS resource IDs.
```
(I didn't add the apostrophe since we don't have a style guide that mandates it (to my knowledge), and https://english.stackexchange.com/questions/212040/is-the-plural-form-of-id-spelled-ids-or-id :) )

README.rst
sabre1041: It can stay. RST is indeed "fun"

felixfontein:
@sabre1041 thanks a lot for reviewing this!

PR #1933, opened 2025-09-02
Allow non-complex non-string values in dotenv and exec-env

billy4479:
Closes  #1914 (supersedes).
Fixes #879.
Fixes #1167 (as no longer applies).

Rebased on top of #1929.

PR #1959, opened 2025-09-27
Complex values in dotenv, and exec-env: do not print sensitive value in error message

### LINKED ISSUE #1198: Suggestion to not emit decrypted file values into logs on 'complex value in dotenv file' failure
If you try to run `exec-env` with a file with a complex value YAML entry, it fails and outputs the failing complex value block into the log. I wanted to raise this as a potential security risk, as it can lead to secrets being written into logs if this command fails in CI, for example.

I think this is the line where the error is generated:

https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/stores/dotenv/store.go#L116

I would be very happy to submit a PR to change this, but wanted to raise it first to get your thoughts.

felixfontein:
Fixes #1198.
Follow-up to #1933.

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -242,7 +242,7 @@ func main() {
 				var env []string
 				for _, item := range tree.Branches[0] {
 					if stores.IsComplexValue(item.Value) {
-						return cli.NewExitError(fmt.Errorf("cannot use complex value in environment; key is %s", item.Key), codes.ErrorGeneric)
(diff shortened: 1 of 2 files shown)


felixfontein:
I first wanted to add a commit that also checks for complex values in the INI store, but then I realized that the INI store already allowed to convert complex values to strings for a long time, and thus we shouldn't simply break it now as part of a bugfix.

(That was also the original motivation to move `IsComplexValue()` to `stores`.)

felixfontein:
@sabre1041 @hiddeco thanks for reviewing this!

PR #1965, opened 2025-10-03
fix: Add decryption-order flag to exec-env, exec-file, and publish commands

joel-u410:
These commands were referencing the `decryption-order` option but not actually declaring it as a flag. This commit adds the flag to these commands, allowing users to specify the order in which decryption methods are attempted, just like with "sops -d".

### DIFF
--- a/cmd/sops/main.go
+++ b/cmd/sops/main.go
@@ -196,6 +196,11 @@ func main() {
 					Name:  "same-process",
 					Usage: "run command in the current process instead of in a child process",
 				},
+				cli.StringFlag{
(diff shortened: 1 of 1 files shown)


felixfontein:
Thanks for your contribution! Can you please sign-off the commit (for the DCO) and force-push? Thanks.

joel-u410:
> Thanks for your contribution! Can you please sign-off the commit (for the DCO) and force-push? Thanks.

Done!

felixfontein:
@joel-u410 thanks for fixing this!

PR #1975, opened 2025-10-14
wrong word usage decryption > encryption

DennisGaida:
Stumbled upon this after mulling over `sops encrypt file > file.enc` vs `sops encrypt file --output file.enc` and both options don't warn me when overwriting an existing file.


felixfontein:
@DennisGaida thanks for fixing this!

PR #1980, opened 2025-10-21
Add support for skipping Azure KeyVault Uri Verification

PleaseStopAsking:
This PR introduces support for skipping the Azure Key Vault Uri Verification that was introduced in `azure-sdk-for-go/sdk/security/keyvault/azkeys@v1.0.0` which sops integrated with `v3.8.0`. This change introduced issues for those who use Key Vaults that are located behind proxies that leverage CNAMES.

Original MS Blog: https://devblogs.microsoft.com/azure-sdk/guidance-for-applications-using-the-key-vault-libraries/

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,19 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/go
+{
+	"name": "Go",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/go:1-1.24-bookworm",
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	"features": {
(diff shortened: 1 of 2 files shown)


cmd/sops/main.go
```diff
@@ -2262,7 +2266,9 @@ func toExitError(err error) error {
 
 func keyservices(c *cli.Context) (svcs []keyservice.KeyServiceClient) {
 	if c.Bool("enable-local-keyservice") {
-		svcs = append(svcs, keyservice.NewLocalClient())
+		// propagate azure-kv-skip-uri-validation flag to local keyservice server instance
+		local := keyservice.NewCustomLocalClient(keyservice.Server{Prompt: false, SkipAzureKvUriValidation: c.Bool("azure-kv-skip-uri-validation")})
```
felixfontein: You probably have to use both `c.Bool()` and `c.GlobalBool`, since `c.Bool` will likely not provide the right value in subcommands.

PleaseStopAsking:
@felixfontein I am not sure what I did but I appear to have broken my local branch in some way. I am going to close this PR and open a new one with my changes based off of the main branch. I believe the cause is that I used the v3.11 tag as the base for my PR.

PR #1981, opened 2025-10-22
Add support for skipping Azure KeyVault Uri Verification

PleaseStopAsking:
This PR (attempt 2) introduces support for skipping the Azure Key Vault Uri Verification that was introduced in azure-sdk-for-go/sdk/security/keyvault/azkeys@v1.0.0 which sops integrated with v3.8.0. This change introduced issues for those who use Key Vaults that are located behind proxies that leverage CNAMES.

Original MS Blog: https://devblogs.microsoft.com/azure-sdk/guidance-for-applications-using-the-key-vault-libraries/

Previous PR: #1980


PleaseStopAsking:
@felixfontein Is there anything I can do to get this moved along as its a major roadblock for me at the moment. Thanks!

PleaseStopAsking:
bump for review or clarification if this is not going to be considered.

PleaseStopAsking:
@hiddeco @felixfontein Would love any sort of feedback on this functionality making it into a release because as it stands, azure key vaults behind a CNAME based proxy are completely unusable.

felixfontein:
I have zero experience with AZKV and can't say much about this, but the way this PR implements a solution looks very wrong to me. If instead passing things through the keyservice it would use environment variables as the other cloud-based keyservices I think it would be a lot cleaner. Others from @getsops/maintainers might be able to give better comments.

PleaseStopAsking:
@felixfontein I have refactored this to align with the other cloud backends as suggested instead of passing it through CLI.

PR #2009, opened 2025-12-07
Fix SOPS_AGE_SSH_PRIVATE_KEY_FILE for age recipients

### LINKED ISSUE #1999: SOPS_AGE_SSH_PRIVATE_KEY_FILE fails without an error
I'm using sops 3.11.0 and trying to decrypt a file. This seems to be failing with basically output about why it fails. I'm using the following command: `sops -d secret_file`. This is the output:
```

Group 0: FAILED
  [AGE KEY]: FAILED
      | age: no identity matched any of the recipients. Did not find
      | keys in locations 'SOPS_AGE_SSH_PRIVATE_KEY_FILE',
      | '/home/<user>/.ssh/id_rsa', 'SOPS_AGE_KEY',
      | 'SOPS_AGE_KEY_FILE', and 'SOPS_AGE_KEY_CMD'.

[ANOTHER KEY]: FAILED
...

Recovery failed because no master key was able to decrypt the file. In
order for SOPS to recover the file, at least one key has to be successful,
but none were.
```

I have an ssh key file which should be able to decrypt this file, and I'm using the following command to try and get sops to recognize that:
`env SOPS_AGE_SSH_PRIVATE_KEY_FILE=/path/to/ssh_key sops -d secret_file`. This produces the following output:
```

Group 0: FAILED
  [AGE KEY]: FAILED
      | age: no identity matched any of the recipients. Did not find
      | keys in locations '/home/<user>/.ssh/id_rsa',
      | 'SOPS_AGE_KEY', 'SOPS_AGE_KEY_FILE', and 'SOPS_AGE_KEY_CMD'.

  [ANOTHER KEY]: FAILED
  ...

Recovery failed because no master key was able to decrypt the file. In
order for SOPS to recover the file, at least one key has to be successful,
but none were.
```

So it... checks the path? But seems to quietly fail???? The only weirdness here that I can think of is that the ssh key file is in a linked directory. It's annoying that I'm just getting - no additional info about WHY the key I provided fails.

When I run the command with `--verbose` I just get the following additional output:
```
[AGE]	 INFO[0000] Decryption failed
[PGP]	 INFO[0003] Decryption failed                             fingerprint=[ANOTHER KEY]
```

Anyways - not really sure what to do with this or how to even start debugging this since there is no real output.

gokselk:
Fixes #1999

## Problem

Users who convert their SSH public keys to age recipients using `ssh-to-age` cannot decrypt with `SOPS_AGE_SSH_PRIVATE_KEY_FILE`. This is a common workflow for reusing existing SSH keys with SOPS.

Example workflow that fails:
```bash
# Convert SSH public key to age recipient
ssh-to-age -i ~/.ssh/id_ed25519.pub > age-recipient.txt

# Encrypt to that recipient
sops -e --age $(cat age-recipient.txt) secrets.yaml > secrets.enc.yaml

# Try to decrypt with the SSH key - FAILS before this fix
SOPS_AGE_SSH_PRIVATE_KEY_FILE=~/.ssh/id_ed25519 sops -d secrets.enc.yaml
```

## Root Cause

SOPS only creates an SSH identity from the key file. SSH identities can decrypt data encrypted to SSH recipients, but NOT data encrypted to age X25519 recipients (even when derived from the same key).

## Solution

For ed25519 SSH keys, now create both:
1. **SSH identity** - decrypts data encrypted to SSH recipients (existing behavior)
2. **Age X25519 identity** - decrypts data encrypted to age recipients derived from the same key (new)

The conversion uses the `github.com/Mic92/ssh-to-age` library which implements the ed25519 -> curve25519 -> bech32-encoded age identity conversion.

For encrypted (passphrase-protected) SSH keys, a `lazyEd25519AgeIdentity` wrapper is used that:
- Defers passphrase prompting until decryption is actually attempted
- Pre-filters by recipient to avoid unnecessary passphrase prompts when the key doesn't match
- Caches the decrypted identity for subsequent use

Non-ed25519 keys (RSA, ECDSA) are unchanged - they only get an SSH identity. age uses X25519 which is based on Curve25519, the same curve as ed25519, making conversion possible only for ed25519 keys.

## Changes

| File | Change |
|------|--------|
| `age/ssh_parse.go` | Add `lazyEd25519AgeIdentity` for lazy passphrase prompting, `recipientMatcher` interface, use `ssh-to-age` library for key conversion |
| `age/keysource.go` | Filter identities by recipient before decryption, handle slice of identities instead of single identity |
| `age/keysource_test.go` | Add regression test for issue #1999 |
| `go.mod` | Add `github.com/Mic92/ssh-to-age` dependency |

## Test Plan

- [x] Unit test added that encrypts to an age recipient derived from an SSH key, then decrypts using SOPS with that SSH key
- [x] Existing tests updated to expect 2 identities from ed25519 keys
- [x] Manual verification with encrypted SSH key (passphrase prompting works correctly)

### DIFF
--- a/age/ssh_parse.go
+++ b/age/ssh_parse.go
@@ -164,7 +164,7 @@ func parseSSHIdentitiesFromPrivateKeyFile(keyPath string) ([]age.Identity, error
 			pubKeyBytes := ssh.MarshalAuthorizedKey(pubKey)
 			ageRecip, err := agesshconv.SSHPublicKeyToAge(pubKeyBytes)
 			if err != nil {
-				log.WithField("path", keyPath).Debug("Failed to derive age recipient from SSH public key, skipping age identity: " + err.Error())
+				log.WithField("path", keyPath).Debugf("Failed to derive age recipient from SSH public key, skipping age identity: %v", err)
 			} else if ageRecip != nil {
 				identities = append(identities, &lazyEd25519AgeIdentity{
 					contents:      contents,
@@ -186,11 +186,11 @@ func parseSSHIdentitiesFromPrivateKeyFile(keyPath string) ([]age.Identity, error
 	// data encrypted to age recipients derived from this SSH key (via ssh-to-age).
 	ageIdentityStr, _, err := agesshconv.SSHPrivateKeyToAge(contents, nil)
 	if err != nil {
-		log.WithField("path", keyPath).Debug("Failed to convert SSH key to age identity, skipping: " + err.Error())
+		log.WithField("path", keyPath).Debugf("Failed to convert SSH key to age identity, skipping: %v", err)
 	} else if ageIdentityStr != nil {
 		ageIdentity, err := age.ParseX25519Identity(*ageIdentityStr)
 		if err != nil {
-			log.WithField("path", keyPath).Debug("Failed to parse age identity from converted SSH key: " + err.Error())
+			log.WithField("path", keyPath).Debugf("Failed to parse age identity from converted SSH key: %v", err)
 		} else {
 			identities = append(identities, ageIdentity)
 		}


felixfontein:
Thanks for your contribution. I don't see that this PR has a chance of getting merged in its current form though. It contains a lot of apparently vendored code from an unknown source (I don't see any links to the source) that we would now have to maintain.

gokselk:
@felixfontein Thanks for taking a look! Yeah fair point, I should've added proper attribution. The bech32 code is from ssh-to-age which itself copies from age's internal package.

The tricky part is age keeps their bech32 internal, so we can't just import it directly. A few options:

1. Keep vendored code, just add source links to age's internal/bech32
2. Use github.com/Mic92/ssh-to-age/bech32 - same code but exported, drop-in replacement
3. Use github.com/btcsuite/btcd/btcutil/bech32 - bit weird adding a bitcoin lib for this though

Let me know which direction you'd prefer and I'll update the PR.

felixfontein:
I'd probably import the code from ssh-to-age:

> Use github.com/Mic92/ssh-to-age/bech32 - same code but exported, drop-in replacement

gokselk:
> I'd probably import the code from ssh-to-age:
> 
> > Use github.com/Mic92/ssh-to-age/bech32 - same code but exported, drop-in replacement

Done, switched to github.com/Mic92/ssh-to-age/bech32 and removed the vendored code.

gokselk:
Updated the PR description. The implementation has changed quite a bit. Also, the ssh-to-age tags don't follow Go module version numbering conventions, so I've opened an issue to address this: https://github.com/Mic92/ssh-to-age/issues/190

felixfontein:
CC @Mic92 - do you think using ssh-to-age in here like this makes sense?

felixfontein:
Ping @Mic92

felixfontein:
Please note that there's now a conflict since #2070 got merged. Also I'm still waiting for feedback from other maintainers (ping @getsops/maintainers) on whether adding the new dependency is OK.

Mic92:
> Ping @Mic92

Does age/sops with age not support ssh key directly (https://pkg.go.dev/filippo.io/age/agessh)? Afaik this was mostly needed when this wasn't the case. But I haven't read through the whole thread yet.

Mic92:
```
# Convert SSH public key to age recipient
ssh-to-age -i ~/.ssh/id_ed25519.pub > age-recipient.txt

# Encrypt to that recipient
sops -e --age $(cat age-recipient.txt) secrets.yaml > secrets.enc.yaml

# Try to decrypt with the SSH key - FAILS before this fix
SOPS_AGE_SSH_PRIVATE_KEY_FILE=~/.ssh/id_ed25519 sops -d secrets.enc.yaml
```

<- This seems like a weird use case. You can use both consistently use the ssh key format to encrypt/decrypt or you convert your ssh key to age, both public/private key and than use the age format. But what this PR is proposing is to mix these two.

felixfontein:
@Mic92 thanks a lot for weighting in! I spent some time today thinking about this, and I'm currently tending to agree. I guess this feature would mainly be useful for users who manually converted SSH keys to Age identities with your tool in the past, and used the converted keys to encrypt SOPS files, and now want to decode without having to keep the converted keys. While I can understand that this is useful, I don't like the increase of complexity this apparent convenience feature brings to the codebase.

@getsops/maintainers more opinions?

PR #2015, opened 2025-12-13
Add JSON Schema validation for .sops.yaml configuration files

miiyakumo:
Adds JSON Schema validation for `.sops.yaml` configuration files, enabling IDE auto-completion, real-time validation, and programmatic config checking.

---

## What's Changed

- **JSON Schema Definition** (`schema/sops.json`): A comprehensive schema defining all available configuration options, types, and constraints.
  
- **Schema Test Suite** (`config/config_schema_test.go`): Includes 8 test functions and 13 test cases (7 valid configurations and 6 invalid configurations) to ensure schema correctness.
  
- **Documentation Update**: A usage guide has been added to `README.rst`.
  
- **Build Target**: Introduced a new `make test-schema` target for conveniently running schema validation tests.
  
---

## Implementation Notes

I opted for a **hand-written** JSON Schema (`schema/sops.json`) to avoid introducing new external dependencies at this time.

While generating the schema from Go structs is technically feasible, it would require extensive custom logic and more complex struct tags to properly address the following challenges in the current configuration structure:

1. **Complex Type Handling**: Fields with flexible types, such as `Age interface{}` (which accepts a string or an array of strings), necessitate custom `oneOf` logic in the schema.
  
2. **Validation Constraints**: Business-specific rules, such as range constraints (`shamir_threshold >= 1`) and defined enum values, are not automatically derived from standard Go struct tags.
  
3. **Mutually Exclusive Fields**: Custom validation is required for options that are incompatible with each other (mutually exclusive fields).
  

---

Related: #1767

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -849,6 +849,22 @@ Creating a new file with the right keys is now as simple as
 Note that the configuration file is ignored when KMS or PGP parameters are
 passed on the SOPS command line or in environment variables.
 
+Validating .sops.yaml with JSON Schema
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SOPS provides a JSON Schema for validating ``.sops.yaml`` configuration files.
+The schema is located at ``schema/sops.json`` in the repository and can be used
+with editors and validation tools to catch configuration errors early.
+
+**Using with YAML Language Server:**
+
+You can add a schema reference directly in your ``.sops.yaml`` file:
+
+.. code:: yaml
+
+    # yaml-language-server: $schema=https://raw.githubusercontent.com/getsops/sops/main/schema/sops.json
+
+
 Specify a different GPG executable
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


felixfontein:
@miiyakumo thanks for your contribution!

As indicated in https://github.com/getsops/sops/issues/1767#issuecomment-2676849303 I think it's better to have a single source of truth. Right now (with this PR) we have three: the Go structures, the documentation (https://github.com/getsops/sops?tab=readme-ov-file#5config-file-format, that's not even complete yet), and the schema (added in this PR).

Maybe we should have a different source of truth, and use that one to generate all three (Go structures, documentation, and JSON schema)? WDYT @getsops/maintainers?

b0o:
I've tested the JSON schema manually with a comment like this at the top of `.sops.yaml`:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/miiyakumo/sops/fb89aeaae6d1bc8cecab27d636f1650029702b39/schema/sops.json
```

The schema is a bit too strict in that it prevents the user from using this trick described in the sops-nix docs to avoid repeating a key: https://github.com/Mic92/sops-nix/blob/c7067be8db2c09ab1884de67ef6c4f693973f4a2/README.md?plain=1#L216-L232

The LSP reports "property keys is not allowed".

PR #2019, opened 2025-12-15
Fix recursive publish

sabre1041:
Currently, the `publish` subcommand with the `--recursive` option is not configured properly and will cause errors. This enhancement resolves the issue


felixfontein (approved):
I'm not able to test this, but this looks correct to me.

It looks like this line was accidentally removed in c822b552903011488b521fa67f717f1bb07a97ae, so this has been broken since 3.9.0.

felixfontein:
Thanks for fixing this @sabre1041!

Issue #1651, opened 2024-10-17
Openstack Barbican KMS support


poikilotherm:
Hello there!

Would there be interest in upstreaming 3d67bf1f4b14ec8f19c38c5481a32a016ab44357 (coming from https://github.com/getsops/sops/compare/main...clelange:sops:master)?

As an Openstack user, I would greatly benefit from using SOPS with Barbican in FluxCD.

Cheers!

rochaporto:
From my side we've moved away from this solution long ago. There are still people using sops but with a vault backend. But most are using vault directly, with secret retrieval at deployment time with plugins for templating.

rochaporto:
> > From my side we've moved away from this solution long ago. There are still people using sops but with a vault backend. But most are using vault directly, with secret retrieval at deployment time with plugins for templating.
> 
> Mind going into some more detail on your solution?
> 
> We're building a multi-cloud (different providers due to regulations) k8s cluster setup, and want a solution that is as cloud provider agnostic as possible.

Sure. We're relying on a central ArgoCD deployment with a plugin hooking to the argocd-vault-plugin (https://argocd-vault-plugin.readthedocs.io/en/stable/howitworks/), the config is something like:
```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: argocd
  name: vault-cmp
data:
  avp-helm.yaml: |
    apiVersion: argoproj.io/v1alpha1
    kind: ConfigManagementPlugin
    metadata:
      namespace: argocd
      name: argocd-vault-plugin-helm
    spec:
      init:
        command:
        - "sh"
        - "-c"
        - "helm dependency update"
      allowConcurrency: true
      discover:
        find:
          command:
            - "sh"
            - "-c"
            - "find . -name 'Chart.yaml' && find . -name 'values.yaml'"
      generate:
        command:
          - "sh"
          - "-c"
          - "helm template $ARGOCD_APP_NAME --include-crds -n $ARGOCD_APP_NAMESPACE ${ARGOCD_ENV_HELM_ARGS} . | argocd-vault-plugin generate - "
      lockRepo: false
```

The placeholders in the app values.yaml then look like:
```
<path:kv/data/path/to/some/item#itemname>
``

We have a similar plugin for kustomize, just running `kustomize build` instead of `helm template`.

The main difference to SOPS is that the secrets are not committed so not part of the versioned app state, and added when doing reconciliation/sync.

Issue #2016, opened 2025-12-13
Proposal: out-of-tree encryption backends


ndrpnt:
This proposal explores ways to enable out-of-tree encryption backends in SOPS. The goal is to make it easy to add new encryption backends, including private ones.

Other components that could benefit from out-of-tree extensibility, such as stores and publishing backends, are out of scope.

Related PRs:

* #683
* #1097
* #1718
* #1226
* #1906
* #2001
* #1984

As stated [multiple](https://github.com/getsops/sops/issues/1352#issuecomment-1816818787) [times](https://github.com/getsops/sops/pull/1097#issuecomment-2607904325), maintainers are open to a plugin system for encryption backends. This avoids adding dependencies and responsibilities to core SOPS. In the future, moving even core providers out-of-tree would make evolving (and forking) them a lighter process.

I've explored two approaches, and I'm looking for maintainer feedback before investing more time.

## Option 1: build on the existing key service infrastructure

> This approach was discussed in #481, but I don't think it went anywhere.

I Hacked together [`sops-generic-keyservice`](https://github.com/ndrpnt/sops-generic-keyservice), a proof-of-concept key service server compatible with [Scaleway Key Manager](https://www.scaleway.com/en/key-manager) and [OVH Key Management Service](https://www.ovhcloud.com/en/identity-security-operations/key-management-service). For a lack of better alternative, it currently piggybacks on AWS' `--kms` flag to pass arbitrary configuration to the KMS.

I believe this approach can get us far without requiring major/breaking changes in SOPS. Some backward-compatible improvements are needed to make "Generic KMS" a first class backend:

1. Add support for `generic_key` in the key service API:
    ```protobuf
    message Key {
      oneof key_type {
        ...
        GenericKey generic_key = 7; // or google.protobuf.Struct ?
      }
    }

    message GenericKey {
      // Allow key service servers to filter only the requests they can handle.
      string kms_id = 1;

      map<string, google.protobuf.Value> configuration = 2;
    }
    ```
1. Add the corresponding flags, e.g `--generic-kms '{"kms_id": "mykms", "configuration": {"foo": "bar"}}'`. Passing a JSON object is inelegant but straightforward. One can also look at how [Terraform](https://developer.hashicorp.com/terraform/language/values/variables#command-line-variables) handles complex dynamic values through flags and environment variables.
1. Add the corresponding configuration options
    ```yaml
    creation_rules:
      - generic_kms:
          - kms_id: mykms
            configuration:
              foo: bar
[...]

PR #2030, opened 2025-12-30
feat: Add OpenStack Barbican support as a Key Source

devx:
This PR integrates OpenStack Barbican as a supported backend for key management. It allows the `keyservice` to authenticate with and retrieve secrets directly from a Barbican instance, expanding our integration capabilities beyond the current stores.

Key Changes
  * Barbican Implementation: Added `BarbicanKeySource` with full authentication (Keystone) and client logic.
  * Integration: Wired Barbican support into the core `keyservice` and store interfaces.
  * Configuration: Updated the configuration system to accept Barbican endpoints, project IDs, and credentials.
  * Safety: Implemented fallback logic and error handling to ensure backward compatibility with existing key sources.

Testing I have added a comprehensive test suite to ensure reliability:
  * [x] Unit Tests: Coverage for the new client and auth logic.
  * [x] Integration Tests: Verified against a live/mocked Barbican endpoint.
  * [x] Property-Based Tests: Added fuzzing/property tests to handle edge cases in key retrieval and parsing.

This also resolves issue #1651

### DIFF
--- a/barbican/client.go
+++ b/barbican/client.go
@@ -686,6 +686,12 @@ func (c *BarbicanClient) doRequestWithRetry(ctx context.Context, method, path st
 		log.WithError(err).WithField("attempt", attempt+1).Debug("Retryable error occurred")
 	}
 
+	// If the context was cancelled or timed out, surface that directly
+	// so callers can distinguish cancellation from server-side failures.
+	if ctx.Err() != nil {
+		return NewTimeoutError("Request cancelled by context", ctx.Err())
+	}
+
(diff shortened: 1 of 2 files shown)


felixfontein:
Thanks for your work! Unfortunately I think this is quite out of scope for the SOPS codebase, since it contains a complete Barbican client implementation. This is nothing we want to maintain here.

I think the best way forward for this is #2016 (which unfortunately isn't anywhere near realization).

devx:
Thank you for the feedback. I'll look at #2016 and see how I can help with that issue.

PR #2052, opened 2026-02-03
test: Unset user GNUPGHOME env var for gnuPGHome

### LINKED ISSUE #2051: Failing test for `gnuPGHome` with user-set GNUPGHOME
_Note: despite similar testing issues this is not the same test failure as #1297 as far as I can tell._

I cloned the repository and ran the tests with `make test`, to this result:

```
--- FAIL: Test_gnuPGHome (0.00s)
    keysource_test.go:687: 
                Error Trace:    /Users/uncenter/Dev/Forks/sops/pgp/keysource_test.go:687
                Error:          Not equal: 
                                expected: "/Users/uncenter/.gnupg"
                                actual  : "/Users/uncenter/.config/gnupg"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -/Users/uncenter/.gnupg
                                +/Users/uncenter/.config/gnupg
                Test:           Test_gnuPGHome
```

It looks like this test assumes I don't have the `GNUPGHOME` environment variable set; it asserts that the result of evaluating `gnuPGHome` is always the default path. The test fails because sops _correctly_ checks the aforementioned environment variable in `gnuPGHome`.

https://github.com/getsops/sops/blob/bf3b7a27d82b36d9e86cd2debebcbedefc10c593/pgp/keysource_test.go#L687

Adding `unset GNUPGHOME;` to the Makefile (for `make test` on line 76) doesn't fix this because other tests do depend on having access/reference to the folder that is `$GNUPGHOME`. Can/should we unset the GNUPGHOME variable just for this test? In a later assert of the same test the environment variable functionality is checked.

https://github.com/getsops/sops/blob/bf3b7a27d82b36d9e86cd2debebcbedefc10c593/pgp/keysource_test.go#L692-L694

uncenter:
Closes #2051 by unsetting the user-set `GNUPGHOME` var. Cannot be unset in Makefile without breaking later usage of actual gpg config.

### DIFF
--- a/pgp/keysource_test.go
+++ b/pgp/keysource_test.go
@@ -682,6 +682,7 @@ func Test_gpgBinary(t *testing.T) {
 }
 
 func Test_gnuPGHome(t *testing.T) {
+	t.Setenv("GNUPGHOME", "")
 	usr, err := user.Current()
 	if err == nil {
 		assert.Equal(t, filepath.Join(usr.HomeDir, ".gnupg"), gnuPGHome(""))


sabre1041 (approved):
LGTM

felixfontein:
@uncenter thanks for your contribution!
@sabre1041 thanks for testing and reviewing!

PR #2071, opened 2026-02-15
Add Go 1.26 to CI; use for release

felixfontein:
Golang 1.26.0 has been released a few days ago.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [linux, darwin, windows]
         arch: [amd64, arm64]
-        go-version: ['1.24', '1.25']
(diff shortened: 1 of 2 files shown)


sabre1041:
Should we also update the `go.mod` file?

felixfontein:
That would require (resp. be equivalent to) bumping the minimum supported Go version, right? (And would require to remove Go versions from CI?)

sabre1041:
> That would require (resp. be equivalent to) bumping the minimum supported Go version, right? (And would require to remove Go versions from CI?)

ah this PR is just to expand testing coverage and not to change the core version from a project perspective.

felixfontein:
Yep, though it switches the Go version that is used to compile the release binary with. But if someone downstream prefers to build it with an older version that's also fine.

felixfontein:

PR #2072, opened 2026-02-15
Remove dead code

felixfontein:
While looking at go-yaml/v4, I noticed there is some dead code in `LoadEncryptedFile()` which loads data that is never used.


sabre1041 (approved):
LGTM

felixfontein:
@sabre1041 thanks again for reviewing!

PR #2081, opened 2026-02-22
CI: Rearrange steps; disable setup-go's caching

felixfontein:
Right now the main test emits a lot of warnings https://github.com/getsops/sops/actions/runs/22272667513:

> Restore cache failed: Dependencies file is not found in /home/runner/work/sops/sops. Supported file pattern: go.sum

This comes from the setup-go action because we run it *before* doing the checkout. But then we're already doing the caching ourselves. This PR rearranges the steps to the recommended order by setup-go (first checkout, then setup-go) and disables setup-go's caching.

(Maybe we should also drop our caching and rely on setup-go's caching?)


felixfontein:

PR #2098, opened 2026-03-06
Check for metadata key(s) before re-encrypting file

### LINKED ISSUE #2093: no error when adding `sops` key to YAML file with `sops edit`
I've used `sops edit file.yaml` to create a YAML file like:

```yaml
# file.yaml
sops: "something"
```

After saving the change, `sops` is unable to decrypt the file: it complains of two `sops` keys in the YAML.

Please add a check for reserved key names to SOPS, and prevent saving a file with an illegal key name (`sops`).

felixfontein:
Fixes #2093.


felixfontein:

PR #2100, opened 2026-03-08
fix: handle mac only encrypted flag in global

CzBiX:
It always returning nil for the flag currently.


felixfontein:
`sops --mac-only-encrypted --encrypt` seems to work fine (with current `main`), but `sops --mac-only-encrypted encrypt` does not.

felixfontein (approved):
I had time to test the PR and it does what promised. Thanks a lot @CzBiX!

PR #2104, opened 2026-03-13
sops edit: delete temporary file on termination

felixfontein:
If SOPS cannot read the file after the user edited it, the user can chose between editing again and cancelling (with Ctrl+C). Also the user can kill SOPS while the editor is open. In both cases, the temporary directory with the temporary file was kept.

This PR tries to clean up the directory in these cases. If SOPS is force killed (`kill -9`) there's not much we can do, but in other cases (regular kill or Ctrl+C) the cleanup works - at least on Linux and macOS.

### DIFF
--- a/cmd/sops/edit.go
+++ b/cmd/sops/edit.go
@@ -3,12 +3,15 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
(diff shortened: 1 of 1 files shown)


felixfontein:

PR #1401, opened 2023-12-30
Serialize using mapstructure conversion instead of JSON marshalling.

slewsys:
These changes remove the JSON encoding/decoding steps that are performed during the serialization of *ini* and *dotenv* files. This roundtrip loses type information during the transformation which causes values to be incorrectly converted to the JSON marshaller defaults (`int` becomes `float64`, `bool` becomes `string`, etc, etc). In place of this JSON encoding, the **mapstructure** library allows for a direct conversion between the `Metadata struct` and `map[string]interface{}` needed to leverage the `stores.Flatten` and `stores.Unflatten` functions.

In addition, this adds mapstructure tags to the metadata structures to allow backwards compatibility with the JSON encoding.

Resolves #879 & resolves #857

This is a re-submission of @acastle's #1009 - with **mapstructure** updated to v1.5.0 and unused imports removed - applied to getsops/sops/main HEAD (previously #1338).

PR #1338, opened 2023-11-03
Serialize using mapstructure conversion instead of JSON marshalling.

slewsys:
These changes remove the JSON encoding/decoding steps that are performed during the serialization of ini and dotenv files. This roundtrip loses type information during the transformation which causes values to be incorrectly converted to the JSON marshaller defaults (int becomes float64, bool becomes string, etc, etc). In place of this JSON encoding, the mapstructure library allows for a direct conversion between the Metadata struct and
map[string]interface{} needed to leverage the stores.Flatten and stores.Unflatten functions.

In addition this adds mapstructure tags to the metadata structures to allow backwards compatibility with the JSON encoding.

Resolves #879 & resolves #857

This is a re-submission of @acastle's #1009 with mapstructure updated to v1.5.0 and unused imports removed applied to getsops/sops/main HEAD.

PR #1046, opened 2022-04-25
Serialize using mapstructure conversion instead of JSON marshalling.

slewsys:
Merge of Alex Castle's (acastle) PR #1009  against **HEAD** of branch _master_.
 
To make concrete the errors that this PR fixes, in a Linux terminal,  install `age`:

```
go install filippo.io/age/cmd/age@latest
go install filippo.io/age/cmd/age-keygen@latest
```

Generate an age key:

```
install -d -m 0700 $HOME/.config/sops/age
export SOPS_AGE_KEY_FILE=$HOME/.config/sops/age/keys.txt
age-keygen -o  "$SOPS_AGE_KEY_FILE"
export SOPS_AGE_RECIPIENTS=$(age-keygen -y "$SOPS_AGE_KEY_FILE")
```

Generate an age-encrypted secret:

```
echo 'secret: abc' >o.yml
sops -e o.yml >o.enc.yml
```

Use the secret in a script:

```
sops exec-env o.enc.yml 'echo $secret'
# => abc
```

So far, so good. Now, try a numeric secret:

```
echo 'secret: 123' >o.yml
sops -e o.yml >o.enc.yml
sops -d o.enc.yml
# => secret: 123
```

Decryption still works, but not in a script context:

```
sops exec-env o.enc.yml 'echo $secret'
```

> panic: interface conversion: interface {} is int, not string
> 
> goroutine 1 [running]:
> go.mozilla.org/sops/v3/stores/dotenv.(*Store).EmitPlainFile(0xc0004c4000, {0xc0001234d0, 0x1, 0x1})
> 	${GOPATH}/pkg/mod/go.mozilla.org/sops/v3@v3.7.2/stores/dotenv/store.go:122 +0x305
> main.decrypt({{0x102bc30, 0xc000422db0}, {0x7ffb607bca10, 0x165b608}, {0x1038c98, 0x165b608}, {0x7ffff43fc8aa, 0x9}, 0x0, {0x0, ...}, ...})
> 	${GOPATH}/pkg/mod/go.mozilla.org/sops/v3@v3.7.2/cmd/sops/decrypt.go:47 +0x23f
> main.main.func1(0xc00018cb00)
> 	${GOPATH}/pkg/mod/go.mozilla.org/sops/v3@v3.7.2/cmd/sops/main.go:156 +0x258
> gopkg.in/urfave/cli%2ev1.HandleAction({0xd0d640, 0xea99a8}, 0x8)
[...]

PR #2120, opened 2026-03-23
Use mapstructure to (de-)serialize internal metadata; move metadata processing from specific stores to general place

felixfontein:
Right now, metadata is handled quite strangely when flattened (by going through JSON), and all stores somehow have to deal with metadata. This PR cleans this up as follows:

* Use mapstructure to convert to/from metadata and coerce types.
* Convert metadata to/from `sops.TreeBranch`, and provide three options for handling flattening (none; keep `sops` toplevel map; flatten completely).
* Properly handle errors during flattening and unflattening. Right now you can panic SOPS and overwrite values by playing around with flattened metadata.
* Remove old public APIs for dealing with internal metadata and (un-)flattening.

While this is a breaking change, I think this is still OK:

* The APIs that are broken are internal ones. To from our promise, they're not public and it's not a breaking change.
* They should only be used by store implementations, which are all part of SOPS itself. This means that even though some other programs use parts of SOPS as a library that aren't public, it is unlikely that they directly use these parts of SOPS that are changed by this PR.

This is related to #1401, #1338, #1046, #1009, which were earlier attempts to introduce mapstructure.

### DIFF
--- a/stores/flatten.go
+++ b/stores/flatten.go
@@ -66,26 +66,6 @@ func tokenize(path string) []token {
 	return tokens
 }
 
-// DecodeNewLines replaces \\n with \n for all string values in the map.
-// Used by config stores that do not handle multi-line values (ini, dotenv).
-func DecodeNewLines(m map[string]interface{}) {
-	for k, v := range m {
-		if s, ok := v.(string); ok {
-			m[k] = strings.Replace(s, "\\n", "\n", -1)
-		}
-	}
-}
-
-// EncodeNewLines replaces \n with \\n for all string values in the map.
-// Used by config stores that do not handle multi-line values (ini, dotenv).
-func EncodeNewLines(m map[string]interface{}) {
-	for k, v := range m {
-		if s, ok := v.(string); ok {
-			m[k] = strings.Replace(s, "\n", "\\n", -1)
-		}
-	}
-}
-
 func descendMap(branch sops.TreeBranch, key string) (sops.TreeBranch, reflect.Value) {
 	for idx, elt := range branch {
 		if elt.Key == key {
--- a/stores/flatten.go
+++ b/stores/flatten.go
@@ -7,88 +7,11 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/go-viper/mapstructure/v2"
-
 	"github.com/getsops/sops/v3"
 )
 
 const mapSeparator = "__map_"
 const listSeparator = "__list_"
-
-// flattenAndMerge flattens the provided value and merges into the
-// into map using prefix
-func flattenAndMerge(into map[string]interface{}, prefix string, value interface{}) {
-	flattenedValue := flattenValue(value)
-	if flattenedValue, ok := flattenedValue.(map[string]interface{}); ok {
-		for flatK, flatV := range flattenedValue {
-			into[prefix+flatK] = flatV
-		}
-	} else {
-		into[prefix] = value
-	}
-}
-
-func flattenValue(value interface{}) interface{} {
-	var output interface{}
-	switch value := value.(type) {
-	case map[string]interface{}:
-		newMap := make(map[string]interface{})
-		for k, v := range value {
-			flattenAndMerge(newMap, mapSeparator+k, v)
-		}
-		output = newMap
-	case []interface{}:
-		newMap := make(map[string]interface{})
-		for i, v := range value {
-			flattenAndMerge(newMap, listSeparator+fmt.Sprintf("%d", i), v)
-		}
-		output = newMap
-	case []map[string]interface{}: // mapstructure also emits these
-		newMap := make(map[string]interface{})
-		for i, v := range value {
-			flattenAndMerge(newMap, listSeparator+fmt.Sprintf("%d", i), v)
-		}
-		output = newMap
-	default:
-		output = value
-	}
-	return output
-}
-
-// Flatten flattens a map with potentially nested maps into a flat
-// map. Only string keys are allowed on both the top-level map and
-// child maps.
-func Flatten(in map[string]interface{}) map[string]interface{} {
-	newMap := make(map[string]interface{})
-	for k, v := range in {
-		if flat, ok := flattenValue(v).(map[string]interface{}); ok {
-			for flatK, flatV := range flat {
-				newMap[k+flatK] = flatV
-			}
-		} else {
-			newMap[k] = v
-		}
-	}
-	return newMap
-}
-
-// FlattenMetadata flattens a Metadata struct into a flat map.
-func FlattenMetadata(md Metadata) (map[string]interface{}, error) {
-	var mdMap map[string]interface{}
-	config := mapstructure.DecoderConfig{
-		Result: &mdMap,
-	}
-	d, err := mapstructure.NewDecoder(&config)
-	if err != nil {
-		return nil, err
-	}
-	err = d.Decode(md)
-	if err != nil {
-		return nil, err
-	}
-	flat := Flatten(mdMap)
-	return flat, nil
-}
 
 type token interface{}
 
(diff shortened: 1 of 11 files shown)


stores/metadata.go
```diff
@@ -0,0 +1,287 @@
...
+	if tb, ok := metadata.(sops.TreeBranch); ok {
+		return tb, nil
+	}
+	return nil, fmt.Errorf("Internal error: unexpected metadata conversion result %T", metadata)
+}
+
+func SerializeMetadata(data sops.Tree, opts MetadataOpts) (sops.TreeBranches, error) {
+	md, err := metadataToTreeBranch(metadataFromInternal(data.Metadata))
+	if err != nil {
+		return nil, fmt.Errorf("Error while serializing metadata: %e", err)
```
hiddeco:
```suggestion
		return nil, fmt.Errorf("Error while serializing metadata: %w", err)
```

stores/metadata.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	if opts.Flatten != MetadataFlattenNone {
+		var prefix string
+		if opts.Flatten == MetadataFlattenFull {
+			prefix = SopsPrefix
+		}
+		md, err = flattenTreeBranch(md, prefix)
+		if err != nil {
+			return nil, fmt.Errorf("Error while flatting metadata: %e", err)
```
hiddeco:
```suggestion
			return nil, fmt.Errorf("Error while flattening metadata: %w", err)
```

NB: also notice _flattening_.

stores/metadata_test.go
```diff
@@ -0,0 +1,1001 @@
...
+								},
+								sops.TreeItem{
+									Key:   "key_name",
+									Value: "HC Vault key name",
+								},
+							},
+						},
+					},
+					sops.TreeItem{
+						Key: "agekey",
```
hiddeco: Should be


```suggestion
						Key: "age",
```

I believe?

stores/metadata_test.go
```diff
@@ -0,0 +1,1001 @@
...
+						},
+						sops.TreeItem{
+							Key:   "key_name",
+							Value: "HC Vault key name (inner)",
+						},
+					},
+				},
+			},
+			sops.TreeItem{
+				Key: "agekey",
```
hiddeco: Should be

```suggestion
				Key: "age",
```

I believe? Also not caught by tests, as I don't believe we check for it as expected output.

stores/stores.go
```diff
@@ -613,3 +604,23 @@ func ValToString(v interface{}) string {
 		return fmt.Sprintf("%v", v)
 	}
 }
+
+// DecodeNewLines replaces \\n with \n for all string values in the map.
+// Used by config stores that do not handle multi-line values (ini, dotenv).
+func DecodeNewLines(m map[string]interface{}) {
```
hiddeco, line 610: Does not appear to be used anymore? Same for `EncodeNewLines`. Also wonder if because this is no longer called, we are not introducing regressions.

stores/metadata.go
```diff
@@ -0,0 +1,287 @@
...
+	MetadataFlattenNone MetadataFlatten = iota
+	MetadataFlattenBelowTop
+	MetadataFlattenFull
+)
+
+type MetadataOpts struct {
+	Flatten MetadataFlatten
+}
+
+// SopsPrefix is the prefix for all metadatada entry keys
```
hiddeco:
```suggestion
// SopsPrefix is the prefix for all metadata entry keys.
```

stores/metadata.go
```diff
@@ -0,0 +1,287 @@
...
+						first = false
+						if tree, ok := branch[i].Value.(sops.TreeBranch); ok {
+							metadataTree = tree
+						} else {
+							return nil, sops.Metadata{}, fmt.Errorf("Found %v entry that is not a mapping", SopsMetadataKey)
+						}
+					}
+					branch = append(branch[:i], branch[i+1:]...)
+				} else {
+					i += 1
```
hiddeco:
```suggestion
					i++
```

stores/metadata.go
```diff
@@ -0,0 +1,287 @@
...
+				}
+			}
+			branches[bi] = branch
+		}
+	} else {
+		if len(branches) >= 1 {
+			branch := branches[0]
+			for i := 0; i < len(branch); i += 1 {
```
hiddeco:
```suggestion
			for i := 0; i < len(branch); i++ {
```

stores/flatten.go
```diff
@@ -130,162 +69,156 @@ func tokenize(path string) []token {
...
-		// The JSON unserializer treats every number as a float, so the above 'if'
-		// never applies in that situation.
-		if vFloat, ok := v.(float64); ok {
-			m["shamir_threshold"] = fmt.Sprintf("%.0f", vFloat)
+		var err error
+		destination, err = flattenDescendValue(item, fmt.Sprintf("%s%d", prefix, i), destination, destinationMap)
+		if err != nil {
+			return nil, err
 		}
+		i += 1
```
hiddeco:
```suggestion
		i++
```

stores/metadata_test.go
felixfontein: Fixed in 2513d1c738eace4c44d3a93abc13384b372c7f98.

stores/metadata_test.go
felixfontein: Fixed in 2513d1c738eace4c44d3a93abc13384b372c7f98.

stores/stores.go
felixfontein: In the previous versions of the INI and DotEnv stores, these function were used to process the metadata - despite the same transformation being made afterwards (store) resp. before (load) in the plain file variant. (As these transformations are idempotent, this never got noticed before.)

I guess we can simply remove them now; the plain loader/storer of these stores already handles newlines. WDYT?

felixfontein:
Rebased to resolve conflict.

sabre1041 (approved):
LGTM

felixfontein:
(Rebased against current `main`.)

hiddeco (approved):

stores/metadata.go
```diff
@@ -0,0 +1,305 @@
...
+		return tb, nil
+	}
+	return nil, fmt.Errorf("Internal error: unexpected metadata conversion result %T", metadata)
+}
+
+// SerializeMetadata embeds the metadata of `data` into its tree branches.
+// String values within the metadata are assumed to be single-line (RFC3339
+// timestamps, fingerprints, base64-encoded data keys); stores that do not
+// escape newlines (notably ini) will produce malformed output if this
+// assumption is violated.
```
felixfontein: This is not correct. Multi-line strings are acceptable and stores have to handle (and do handle) them.

The INI and DotEnv` stores uses `\n` sequences in the output.

I would simply remove this:
```suggestion
```

stores/metadata.go
```diff
@@ -0,0 +1,305 @@
...
+const (
+	// MetadataFlattenNone keeps the metadata under a single nested `sops`
+	// key in the first tree branch.
+	MetadataFlattenNone MetadataFlatten = iota
+	// MetadataFlattenBelowTop keeps the metadata under a top-level `sops`
+	// key in the first tree branch, with its values flattened beneath it.
+	MetadataFlattenBelowTop
+	// MetadataFlattenFull flattens the metadata fully into the first tree
+	// branch, prefixing each key with `SopsPrefix`. Only single-branch trees
+	// are supported (e.g. dotenv).
```
felixfontein: @hiddeco why introduce this new pre-condition for only full flattening? In the commit message you claimed that it was there implicitly, but I don't see how. For all three values in this enum, SOPS always inserts the metadata into all branches, but extracts it only from the first one. That was the behavior before this PR, and is still the behavior with tihs PR. (Except that fattening is only ever used for formats that allow exactly one branch.)

stores/metadata.go
felixfontein: For some reason GH screwed up my suggestion above. It should be:
```suggestion
```

felixfontein: (For example the PGP encrypted key is multi-line and part of the metadata.)

felixfontein:
@hiddeco @sabre1041 thanks a lot for reviewing and improving the PR!

PR #2154, opened 2026-04-21
sops exec-file: make sure to set GID to correct value instead of UID

felixfontein:
When using `sops exec-file`, currently the user's UID is used as the GID for the `setgid` and `setregid` syscalls. Use the user's GID instead.


hiddeco (approved):

felixfontein:

PR #2155, opened 2026-04-21
sops exec-file: improve error handling, reject non-local paths in `--filename`

felixfontein:
1. The current error handling is not great; errors are printed, but execution continues. Instead, fail on errors.
2. When `--filename` contains a path, right now the code does not check whether the path [is local](https://pkg.go.dev/path/filepath#IsLocal).

### DIFF
--- a/cmd/sops/subcommand/exec/exec_windows.go
+++ b/cmd/sops/subcommand/exec/exec_windows.go
@@ -1,6 +1,7 @@
 package exec
 
 import (
+	"fmt"
 	"os/exec"
 )
 
@@ -17,9 +18,8 @@ func WritePipe(pipe string, contents []byte) {
 	log.Fatal("fifos are not available on windows")
 }
 
-func GetPipe(dir, filename string) string {
-	log.Fatal("fifos are not available on windows")
-	return ""
+func GetPipe(dir, filename string) (string, error) {
+	return "", fmt.Errorf("fifos are not available on windows")
 }
 
 func SwitchUser(username string) {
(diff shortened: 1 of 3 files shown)


cmd/sops/subcommand/exec/exec_windows.go
```diff
@@ -17,9 +17,8 @@ func WritePipe(pipe string, contents []byte) {
 	log.Fatal("fifos are not available on windows")
 }
 
-func GetPipe(dir, filename string) string {
-	log.Fatal("fifos are not available on windows")
-	return ""
+func GetPipe(dir, filename string) (string, error) {
+	return fmt.Errorf("fifos are not available on windows")
```
hiddeco:
```suggestion
	return "", fmt.Errorf("fifos are not available on windows")
```

cmd/sops/subcommand/exec/exec_windows.go
hiddeco, line 5:
```suggestion
	"fmt"
	"os/exec"
```

cmd/sops/subcommand/exec/exec_windows.go
felixfontein: I'm pretty sure I compiled this before commiting it, but apparently I did not :(

felixfontein: Oh. It's in the Windows file. That's why I didn't notice... :facepalm:

hiddeco (approved):

PR #2163, opened 2026-04-29
age keystore: support stream reading

### LINKED ISSUE #1303: Passing age keys via stdin SOPS_AGE_KEY_FILE=/dev/stdin not working
Hi,

I want to load the age keys from a  password safe (keepassxc) and pass them to sops via stdin

```
keepassxc-cli show -a password db.kdbx mykey | SOPS_AGE_KEY_FILE=/dev/stdin sops -d secret.yaml 
```

but I'll get

```

Group 0: FAILED
  xxx: FAILED
      | age: no identity matched any of the recipients
  
  yyy: FAILED
    - | failed to load age identities: failed to parse
      | 'SOPS_AGE_KEY_FILE' age identities: no secret keys found

```

Decryption works fine when I write the output from `keepassxc-cli show -a password db.kdbx mykey` into a file first and use the file path in **SOPS_AGE_KEY_FILE**

### LINKED ISSUE #2162: SOPS_AGE_KEY_FILE not being able to read from file descriptors in SSS setups
Currently, I am building a system where I fetch one or more AGE keys from a remote storage and pass them to SOPS dynamically without touching the disk or polluting environment variables. To meet strict security requirements (preventing the key from leaking via /proc/<pid>/environ or persisting on disk), I am passing the keys via a file descriptor (e.g., /dev/fd/X or bash process substitution <(fetch_key)).

However, SOPS fails to process the key this way. Because file descriptors/pipes are read-once streams (non-seekable), if SOPS attempts to read the file multiple times during its execution, the stream is already exhausted after the first read, resulting in a failure.

I'd like for SOPS to support from read-once streams (like /dev/fd/* or named pipes). The ideal behavior would be for SOPS to read the contents of the key file into memory exactly once during initialization, and reuse that buffered in-memory value for all decryption operations, rather than attempting to reopen or read from the file path multiple times.

Some alternatives I've tried:
1. Using SOPS_AGE_KEY: Rejected due to strict security policies. Environment variables are visible to other processes via /proc/PPID/environ and can easily leak in crash dumps or child processes.
2. Writing to /tmp or /dev/shm: Rejected because writing secrets to the filesystem (even temporary or memory-backed ones) creates a risk of lingering files if the process crashes before cleanup.

I am willing to contribute and fix this! I plan to open a PR shortly to buffer the AGE key upon the first read, which should seamlessly add support for file descriptors. I'm opening this issue to track the problem and discuss the approach.

Dexmachi:
closes #2162 
closes #1303 

# Why

Currently, sops fails to read multiple age keys passed via process substitution or named pipes (like /dev/fd/X) because these are read-once streams without "seek" support, throwing an EOF when it attempts to read it more than once during `loadIdentities()` (which happens when there are multiple recipients or during parsing retries).

Passing keys via file descriptors is a fairly common security pattern to prevent  decrypted secrets from ever touching the disk or leaking into child processes via /prox/PPID/environ.

## How 

- Added a `fileStreamCache` (`sync.Map`) and a `readStreamSafe(path string) ([]byte, error)` helper function in age/keysource.go.
-  When loading identities, `readStreamSafe` checks if the path points to a stream with a `.HasPrefix`, checking for /dev/fd/ or /proc/
- if it _is_ a stream: It reads it into memory and caches the `[]byte` for subsequent calls within the same process.
- if it is _not_ a stream: It bypasses the cache and reads from the disk, as it was originally.
- a exportable `ClearFileStreamCache()` was created in order to zero out cached memory.

- [x] All unit tests passing.
- [x] Locally tested intended behavior, works as intended.


felixfontein:
Thanks for your contribution! Can you please make sure to sign-off your commits (otherwise this cannot be merged), and to remove all unrelated code changes (reformatting) from the PR? Thanks.

Dexmachi:
> Thanks for your contribution! Can you please make sure to sign-off your commits (otherwise this cannot be merged), and to remove all unrelated code changes (reformatting) from the PR? Thanks.

oh, damn, ofc, I am sorry

Dexmachi:
I'm kinda stupid, rolled back my changes only to forget to declare 1 variable and 1 dep lmao

(and the signoff)

age/keysource.go
```diff
@@ -399,14 +402,45 @@ func getUserConfigDir() (string, error) {
...
+		}
+		fileStreamCache.Delete(key)
+		return true
+	})
+}
+
+// reads a file from the given path, if it is a stream (e.g., /dev/fd/* or /proc/*)
+// it caches the content in memory to avoid issues with multiple reads from the same stream.
+func readStreamSafe(path string) ([]byte, error) {
+	isStream := strings.HasPrefix(path, "/dev/fd/") || strings.HasPrefix(path, "/proc/") || strings.HasPrefix(path, "/dev/stdin")
```
felixfontein: I'm wondering whether we should restrict this to all files; fifos can also appear in random other locations in the filesystem. (And on other OSes like Windows the above path prefixes aren't working anyway.)

@getsops/maintainers I'm wondering whether we should make this a more global thing, so that caching can be used in various parts of SOPS for all kind of config and key files that are potentially read multiple times (except files to encrypt/decrypt). WDYT?

Dexmachi: I'd say this would be the move, I just didn't implement this because it wasn't on the PR nor the issue's scope, I thought it might've made it harder to accept this PR.

Also, this is my first PR here, and I don't feel that confident on my Go skills--

I think I was able to solve the compatibility shebackle

felixfontein:
Please note that your PR has a conflict.

Dexmachi:
> Please note that your PR has a conflict.

Trying to resolve it right now

Dexmachi:
As this became stale, I'll be following this Idea:
> I'm wondering whether we should make this a more global thing, so that caching can be used in various parts of SOPS for all kind of config and key files that are potentially read multiple times (except files to encrypt/decrypt). WDYT?

and placing another PR on it.

PR #2164, opened 2026-04-29
HC Vault: add allowlist support for acceptable HC Vault URLs

felixfontein:
Default is `all`, which is the current behavior. When `SOPS_HC_VAULT_ALLOWLIST` is set to `none` or a comma-separated list of prefixes, none or only URLs starting with these prefixes are allowed.


hcvault/keysource.go
```diff
@@ -26,8 +26,72 @@ import (
...
+	All  bool
+	URIs []string
+}
+
+func (al *allowList) Allows(address string) bool {
+	if al.All {
+		return true
+	}
+	for _, uri := range al.URIs {
+		if strings.HasPrefix(address, uri) {
```
sabre1041, line 53: This will most likely not produce a match as address is the host of the Vault server (without a `/`) whereas the allowlist does have a trailing `/` [here](https://github.com/getsops/sops/pull/2164/changes#diff-c9407d2d59eb82b5820a2ff692f2c72984529d72fa4ec7bb9a025dae9a120910R76-R78)

felixfontein: Right! 2505e030a2666f49b527741b7917477004b9bf2c should protect against this.

sabre1041 (approved):
LGTM

Thanks for taking a look at adding the `/` to the vault address

felixfontein:

PR #2166, opened 2026-05-02
Create public encrypt pkg and references to it

### LINKED ISSUE #903: [request] encrypt package
Hi, 

Similarly to the decrypt package, would it be possible to do something similar for encryption? 
I would like to SOPS-encrypt files (ideally `io.Reader`s or similar) from a self-contained internal Go binary.

manute:
Creates the encrypt pkg and refactorize the code to change all the references to it


nachtjasmin:
Unfortunately, the given package interface is pretty rudimentary and requires a lot of manual plumbing by callers to work with the `stores` and all that. In comparison, the `decrypt` package provides much more integration and doesn't require callers to duplicate current behaviour from the current CLI. :/

I tried to integrate it but stopped after I realised that I'd duplicate the CLI at several points.

manute:
@nachtjasmin yeah I noticed that, thanks for the feedback.  This PR it is only to have it as public interface to not have to duplicate the file - so with the pkg of this PR you can implement a method like : 
```go
func EncryptWith(awsKmsArn, filePath string) ([]byte, error) {

	masterKey := kms.MasterKey{
		Arn: awsKmsArn,
	}

	opts := encrypt.EncryptOpts{
		Cipher: aes.NewCipher(),
		KeyServices: []keyservice.KeyServiceClient{
			keyservice.NewLocalClient(),
		},
		InputStore:  &yaml.Store{},
		OutputStore: &yaml.Store{},
		InputPath:   filePath,
		EncryptConfig: encrypt.EncryptConfig{
			KeyGroups: []sops.KeyGroup{
				[]keys.MasterKey{&masterKey},
			},
		},
	}

	return encrypt.Encrypt(opts)
}

```

so at least you have more control over it 👍🏽

PR #2169, opened 2026-05-03
hcvault: SOPS_VAULT_TOKEN_FILE environment variable

### LINKED ISSUE #2168: FR: Hashicorp Vault Token File Environment Variable`
Building on the context #2162 I'd Like to propose the addition of an environment variable for HashiCorp Vault Token files.

# Why
because relying on the default ~/.vault-token location limits flexibility in automated or strictly secure environments. This would allow SOPS to read Vault tokens from file descriptors and different paths, which would improve support for ephemeral secret injection and read-once tokens overall.

I am currently working on a PR to implement this and will link it here shortly

Dexmachi:
closes #2168 

Implements the `SOPS_VAULT_TOKEN_FILE` environment variable to decide which path to read the vault token from.

# Details
- If the env variable is set, SOPS will attempt to read only from the provided path, no fallbacks.
- If the env variable is set and SOPS fails to read it, it returns an explicit error.
- If the env variable is not set, it uses the current bahaviour, falling back to ~/.vault-token and returns no error if there is an error reading it.
- Supports ephemerally bound files (e.g., FDs)

- [x] All automated tests passing


felixfontein:
This PR has the same problem that #2163 fixes for age: the file is read more than once, and thus it won't work with fifos etc.

Dexmachi:
> This PR has the same problem that #2163 fixes for age: the file is read more than once, and thus it won't work with fifos etc.

To solve this issue without bringing changes from #2163, I had to essentially replicate the fix I made over there

I'd say against merging this PR until we decide how to go through with that PR in order to avoid duplicate logic. As such, I'll make this into a draft.

PR #2194, opened 2026-05-25
exec-file/exec-env: reset supplementary groups when changing user

felixfontein:


cmd/sops/subcommand/exec/exec_unix.go
```diff
@@ -51,6 +51,11 @@ func SwitchUser(username string) {
 	uid, _ := strconv.Atoi(user.Uid)
 	gid, _ := strconv.Atoi(user.Gid)
 
+	err = syscall.Setgroups([]int{gid})
```
sabre1041: Are we looking to map the supplemental groups for the specified user or reset it so that it only contains the primary group?

Mapping could be accomplished by something similar to the following

```
	groupIds, _ := user.GroupIds()
	intGroupIds := make([]int, len(groupIds))
	for i, gid := range groupIds {
		intGroupIds[i], _ = strconv.Atoi(gid)
	}

	err = syscall.Setgroups(intGroupIds)
	if err != nil {
		log.Fatal(err)
	}
```

felixfontein: Ah, I missed `user.GroupIds()`. Yeah, that makes sense. I've changed it slightly so that it falls back to `[]int{gid}` if `GroupIds()` returns an error.

sabre1041 (approved):
LGTM

felixfontein:

PR #2204, opened 2026-06-02
Restrict markdownlint to 0.15.0

felixfontein:
Since version 0.16.0 markdownline is more strict on line length limits, which causes a lot of issues in our .md documents. Pinning to the previous release 0.15.0 for now.

Ref: https://github.com/markdownlint/markdownlint/issues/573

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,11 +5,12 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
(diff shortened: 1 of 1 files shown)


felixfontein:
#2206 fixes most of the linting issues found by the latest version, except one case. Related issue: https://github.com/markdownlint/markdownlint/issues/573

felixfontein:
No longer needed, see https://github.com/getsops/sops/pull/2206#issuecomment-4618910918.

PR #2205, opened 2026-06-02
keystore: Generalized readStreamSafe

### LINKED ISSUE #1303: Passing age keys via stdin SOPS_AGE_KEY_FILE=/dev/stdin not working

### LINKED ISSUE #2162: SOPS_AGE_KEY_FILE not being able to read from file descriptors in SSS setups

Dexmachi:
closes https://github.com/getsops/sops/issues/2162
closes https://github.com/getsops/sops/issues/1303

This PR replaces my last attempt at these issues ( #2163 ).

On that PR, some issues regarding cache and memory safety have been brought up.

Additionally, the maintainers have been asked if generalizing that implementation would be good, and so I went to try and do that. (Sorry this took a while, I had some personal matters to attend to.)

This PR adds a new `fsio` package, and with it new `fsio.Read`, `fsio.Open` and `fsio.ClearCache`. These are stream safe and rely on caching credentials in order to mitigate read-once issues with pipes, FIFOs and such.

Tests regarding race conditions and read-once issues have also been added to this package, although some are unix only (mainly due to `syscall.Mkfifo` and FD use to test specific conditions regarding anonymous and named pipes). All tests are passing locally, and I've tested this version manually to check if the implementation works.

Also, this caching concerns _only_ stream reads, so this is purely a feature addition with no actual breaking changes to the current normal way to use the tool. This can easily be changed in the future (it's just removing an `if` statement), so it can extend to the rest of the tool.

> [!NOTE]
> The `fsio.ClearCache` function is technically a breaking change for users with library use-cases. If they use named pipes, anonymous pipes OR any other stream-based approach, they should use `defer fsio.ClearCache` so the cache's bytes get zeroed out from memory. This is _not required_, however it is an important safety consideration.


Dexmachi:
works :b 
image

Dexmachi:
as soon as #2232 gets merged, I'll update this branch

Dexmachi:
hey, uh

are there any updates about this @felixfontein?

felixfontein:
I'm still waiting on what the other maintainers think (https://github.com/getsops/sops/pull/2163#discussion_r3178641157). So far nobody commented here or in the original PR.

Dexmachi:
> I'm still waiting on what the other maintainers think ([#2163 (comment)](https://github.com/getsops/sops/pull/2163#discussion_r3178641157)). So far nobody commented here or in the original PR.

gotcha

PR #2206, opened 2026-06-03
Shorten .md lines

felixfontein:
This passes mdl 0.16, except for one line in the changelog. While it is possible to transform the document to shorten it, IMO this would not increase readability, and therefore I refuse to do it.

~~Needs #2204 to pass CI (but even without, passes it better than current `main`).~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,17 @@
 Improvements:
 
 * Dependency updates ([#2178](https://github.com/getsops/sops/pull/2178)).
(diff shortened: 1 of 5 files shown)


felixfontein:
Ok, with the latest mdl 0.17.0 this is now fine and doesn't require #2204..

felixfontein:
@getsops/maintainers I would be really glad if someone could review and approve this.

sabre1041 (approved):
LGTM

felixfontein:

PR #2210, opened 2026-06-05
Proposal/RFC: out-of-tree encryption backends interface

### LINKED ISSUE #2016: Proposal: out-of-tree encryption backends

Dexmachi:
closes #2016 

## What??

This is a common interface for SOPS that allows for Inter Process Communication via json and stdin/stdout.

## Why?

I've had my fair share of plugin-based architectures (long story short, had to make an decentralized plugin registry based on git repos) AND saw #2016, but didn't see many advancements in that thread regarding implementations. This is my shot at this interface and it is a very early-on PoC.

This is merely centralized proposal, and a place where other people can work on as a community to get it outta the ground.

## How?

Well, I took some inspiration (a LOT of it) from the aforementioned thread, and the most interesting idea, and the one that the people sounded to go towards the most was IPC proxy with stdin/stdouts.

This keeps the core tool entirely in charge of the contract (what the second subprocess needs to tend to in order to remain stable with the main tool), and is just another keystore, so no issues with backwards compatibility here.

So, let's get to the bottom of it shall we?

### The contract:
Firstly, we need to talk about contracts, and how a plugin will... well be a plugin, and how it is meant to work

All sops plugins must follow the `sops-plugin-*` naming convention.

Sops plugins should _IDEALLY_ be stateless and headless. They should only take a stdin and spit out the encrypted stdout. Note that this is currently not architecturally forced, as I don't really know how to force an external binary to follow specific conventions.

We follow a very easy IPC bridge, being a JSON one at that.
All IPC calls and outputs are done through the following contracts:

Sops -> Plugin:
encrypt:
```json
{
  "action": "encrypt",
  "instance_id": "prod-hsm-primary",
  "env_prefix": "SOPS_PLUGIN_PROD_HSM_PRIMARY_",
  "config": {},
  "plaintext": "dGhpcy1pcy1hLXNlY3JldC1kYXRhLWtleQ=="
}
```

decrypt:
```json
{
  "action": "decrypt",
  "instance_id": "prod-hsm-primary",
  "env_prefix": "SOPS_PLUGIN_PROD_HSM_PRIMARY_",
  "config": {},
  "ciphertext": "sops-plugin-encrypted-ciphertext-string-here"
}
```

Plugin -> Sops:
encryption successful:
```json
{
  "ciphertext": "kms_enc_987654321_xyz"
}
```

decryption successful:
```json
{
  "plaintext": "U2VncmVkbyBNdWl0byBJbXBvcnRhbnRl"
}
```

error:
```json
{
  "error": "Access denied to key alias/my-key in eu-west-1"
}
```

> [!NOTE]
> BOTH sops->plugin::encrypt AND plugin->sops::decrypt are made via BASE64 ENCODED plaintext.

## Core changes
There are no rewinds here. This plugin is just another keystore, nothing more than it.

other than that, we have a new `.sops.yaml` key!
```yaml
creation_rules:
   - plugins:
       - binary_name: customkms
          # the "config" part of the stdin
         config:
            region: "eu-west-1"
            key_id: "alias/my-key"
```

and 
```yaml
creation_rules:
- key_groups:
    - plugins:
        - binary_name: customkms
          # the "config" part of the stdin
          instance_id: dev-hsm
          timeout: "5s"
          config:
              region: "eu-west-1"
              key_id: "alias/my-key"
        - binary_name: customkms2
          # the "config" part of the stdin
          config:
              region: "eu-west-1"
              key_id: "alias/my-key"
    - age:
        - age1idk123sla456
```

should both work!

## Notes
The discovery currently is done via PATH, binary_name should be only the _proper plugin name_, not "sops-plugin-my_name". PLUGINS MUST NEVER USE STDOUT FOR THINGS OTHER THAN JSON STREAMS. If so, this will break. IF you need to print something out, use the stderr, and sops will handle it.

Also, note that the "config" part of the plugin is completely an utterly up for anyone to implement. There is no current forcing of behavior done in this regard.

## Limitations

I tried to follow devx's ADR questions on that thread, but there were some questions I've left out because this is currently a PoC.

Q1: about supporting legacy use-cases:
Yes, I do think it would be the way to go to support legacy configs. Ofc, this adds a new layer of maintenance, and could easily be taken out if necessary.

Q2: Is the config opaque or padronized?
The config in this PoC is _completely opaque_. This allows for more complex plugins to arise, once we get a good understanding about how plugins are being implemented, we can probably normalize this.

Q3: About retries being managed by sops or by the plugins:
I do think only letting the plugins handle that would be the best choice, as some plugins may or may not create these, creating a multiple retry cycle that may create significant overhead.

Q4: How should Timeouts be configured?
I'd say that an hybrid format works best here. There's the rule level and the key level, as well as a global fallback, so the hierarchy is Key-level config -> Rule-level config -> 10s fallback. This doesn't even touch the CLI's flags, and is put directly inside the .sops.yml config file.

Q5/6: about instance_ids:
This area definitely needs a thorough review to ensure the implementation follows the projects intents.
responses:
5: I don't think they should be mandatory _for the plugin implementation to use_. However, sending an environment variable string (env_prefix) is harmless and gives plugin devs the choice to use it or ignore it.

6: Yes, I believe deriving an environment variable prefix and passing it to the plugin is the way to go (mainly due to programmatic use-cases). The flow would be: SOPS parses the config -> derives an env-compatible prefix -> passes it in the JSON -> the plugin decides what to do with it.
(Again, I am not a voice of mastery over here, just someone trying to help.)

Q7: about keyservice:
I've implemented this to the best of my current understanding, but guidance and review here are highly appreciated. I would document this as experimental for now, but I did put in "PluginKey" instead of "ExperimentalGenericKeyV1" on the .proto file.

Q8: about error detailing:
So, in here we have two separate cases: expected errors and unexpected errors
1. expected errors should be passed to the stdout with an error code != 0. The stdout should be a json containing an "error" field, where the plugin creator sends in their own details, and sops spits out the binary name and the error passed down to it.
2. unexpected errors are treated as panics for plugin creators. Sops spits out the binary name, the instance_id, the exit code and the stderr of the plugin.

Also, IPC-based plugin architectures will always have some security issues (since the configuration is agnostic of binary). But hey, if AGE is doing it, why not?


Dexmachi:
Just added granular Timeouts!

now if a sops file can look a lil smth like this:

```yaml
creation_rules:
- path_regex: \.dev\.yaml$
    timeout: "3s"  # <~ rule fallback. If omitted, the default fallback is of 10s for all plugins without an explicit timeout.
    plugins:
      - binary_name: fastkms
        config:
          env: development
      - binary_name: local-vault
        config:
          address: "http://127.0.0.1:8200"

  - path_regex: \.prod\.yaml$
    # We don't have a fallback timeout here, so the default of 10s will apply to all plugins without an explicit timeout.
    plugins:
      - binary_name: slow-hsm
        timeout: "15s"     # <~ This plugin is slow, so we set a longer timeout than the default fallback.
        config:
          slot: 1

        # This is a faster plugin, so we can use the default fallback of 10s.
      - binary_name: fast-local
        config:
          mode: light
```

I know this is a bit invasive (new global key and all), but this makes timeouts more granular
we can maybe just get this out

Dexmachi:
I _think_ I've made this InstanceID correctly?

now it is completely explicit inside of each plugin instance:
```yaml
creation_rules:
- path_regex: \.dev\.yaml$
    plugins:
      - binary_name: fastkms
        instance_id: dev-hsm # <~ Will use SOPS_PLUGIN_DEV_HSM_(name up to the plugin) as the instance ID
        config:
          env: development
          
      - binary_name: local-vault # <~ will use SOPS_PLUGIN_LOCAL_VAULT_(name up to the plugin) as the instance ID
        config:
          address: "http://127.0.0.1:8200"
```

also, the stdin for the binary has changed to:

sops -> plugin bin
encryption:
```json
{
  "action": "encrypt",
  "instance_id": "prod-hsm-primary",
  "env_prefix": "SOPS_PLUGIN_PROD_HSM_PRIMARY_",
  "config": {},
  "plaintext": "dGhpcy1pcy1hLXNlY3JldC1kYXRhLWtleQ=="
}
```

decryption:
```json
{
  "action": "decrypt",
  "instance_id": "prod-hsm-primary",
  "env_prefix": "SOPS_PLUGIN_PROD_HSM_PRIMARY_",
  "config": {},
  "ciphertext": "sops-plugin-encrypted-ciphertext-string-here"
}
```

so now the plgin can know both the instance_id and easily parse the env_prefix (without having to parse stuff out)

Dexmachi:
While implementing this, I noticed the changes required significant coupling with core areas (e.g., keystores.go). The actual implementation and tests take up around 500 lines, but the total diff reached ~1000 lines because of this coupling, and I'm not even finished with integrating this with the rest of the project.

Is this tight coupling expected for new keystores  because of security and backward compatibility reasons, or is there a better approach or abstraction I should be using to reduce this footprint?

If this coupling is not intended, I would be happy to help with this architecture. The maintainers have said for some time now that the implementation of new keystores is quite a difficult process, and this PR itself was made because of the interest in the plugin interface to reduce this issue, but I personally do not think this is a good way to mitigate it, since every new addition to the core will always have this friction.

I'd love to hear about the direction this project is going to, and if a decoupled architecture is actually something wanted for it.

Edit:
After 2 or so days, I _think_, I've got a DIP compliant architecture working, although I have not tested it rigorously at all, I've limited myself to local manual tests and running `make test` to run the necessary tests. I'm still trying to understand how to do it better, but as of right now I have a +1.725 -1.240 diff count on top of the branch I'm proposing in this PR. If this is of interest, I can open up a draft PR here, or just link one from my fork. If it is not of interest, I'll drop any development on it ASAP. Note that this PR specifically does _not_ introduce this architecture change.

PR #2219, opened 2026-06-16
fix: handle pointers when serializing context

tlercher:
Starting with the change #2120 we've run into the issue that when saving a SOPS files the context values of a provider (in our case AWS KMS using `tenant`) gets mangled and a pointer value gets written into the encrypted file.

If such a invalid file is committed it cannot be decrypted anymore by `sops` since it lacks the proper `context` to do so.

i also took the liberty to add a method body to `TestSerializeMetadata` even if it just tests the AWS KMS case i've run into for now. -> Should this be a more specific test?


felixfontein (approved):
Thanks for reporting and fixing this!

PR #2220, opened 2026-06-16
fix(keyservice): return error for unsupported master key types

santosh7676:
## Summary

- Replace panic in `keyservice.KeyFromMasterKey` when encountering an unknown `keys.MasterKey` type with explicit error handling.
- Add `KeyFromMasterKeyOrError` and `ErrUnsupportedMasterKeyType` for callers that need to handle unsupported key types safely.
- Update core encrypt/decrypt paths in `sops.go` to propagate conversion errors instead of crashing the process.
- Add regression tests to ensure unsupported key types return an error and do not panic.

`KeyFromMasterKey` previously panicked on unknown master key types. That can crash SOPS during encrypt/decrypt operations instead of returning a recoverable error. This change makes failure behavior consistent with the rest of the key handling flow.

## Behavior change

- **Before:** unsupported `MasterKey` type -> process panic.
- **After:** unsupported `MasterKey` type -> returned error (`ErrUnsupportedMasterKeyType`) and normal error propagation in encrypt/decrypt paths.


felixfontein:
> `KeyFromMasterKey` previously panicked on unknown master key types. That can crash SOPS during encrypt/decrypt operations instead of returning a recoverable error.

This behavior is intentional, since there should be no unsupported master key types. If there are, this is a serious programming error, and a panic is the right reaction IMO.

santosh7676:
Thanks for the review @felixfontein,  that makes sense.
You're right that unsupported `MasterKey` types indicate an internal programming error rather than a user-facing failure, and panic is appropriate for that case. I'll close this PR.

PR #2221, opened 2026-06-18
docs: fix typo in exec-file --filename help text

s3onghyun:
The exec-file --filename help read "filename for the temporarily file"; corrected to "temporary", matching the command's own usage string.


felixfontein (approved):
Thanks for fixing this!

PR #2222, opened 2026-06-18
JSON store: preserve large integers that fit into int64

s3onghyun:
### Problem

The JSON store decodes with `json.Decoder` but never calls `UseNumber()`, so **every JSON number is decoded into a Go `float64`**. `float64` holds only ~15–17 significant decimal digits, so any integer above 2^53 (~9e15) is silently corrupted on re-emit:

```
input :  {"id": 1234567890123456789}
output:  {"id": 1234567890123456800}
```

This hits very common secret payloads — Discord/Twitter snowflake IDs, 64-bit identifiers, unix-nanosecond timestamps, large account/order numbers. In a secrets-management tool, a value silently changing across an encrypt/decrypt round-trip is serious.

### Fix

Call `dec.UseNumber()` on the decoder in `treeBranchFromJSON`. Numbers then arrive as `json.Number` (a string type) and are re-emitted verbatim. The recursive decode helpers share the one decoder, so a single call covers the whole tree. `stores.ValToString` already stringifies `json.Number` via `fmt.Stringer` for encryption/MAC, so no core-path changes are needed.

### Testing

- Added `TestLargeIntegerRoundtrip` (`stores/json/store_test.go`): a `1234567890123456789` value round-trips exactly. Fails before (mangled to `...800`), passes after.
- Updated existing tests that hard-coded the old lossy `float64` representation to the correct lossless `json.Number`.
- `go test ./stores/... .` green.

### DIFF
--- a/stores/json/store.go
+++ b/stores/json/store.go
@@ -123,9 +123,32 @@ func (store Store) sliceFromJSONDecoder(dec *json.Decoder) ([]interface{}, error
 			}
 			slice = append(slice, item)
 		} else {
-			slice = append(slice, t)
+			v, err := normalizeJSONNumber(t)
+			if err != nil {
+				return slice, err
+			}
+			slice = append(slice, v)
 		}
 	}
+}
+
+// normalizeJSONNumber converts a json.Number scalar (produced because the
+// decoder runs with UseNumber) into an int for integers within the int64 range
+// and a float64 otherwise. Non-number tokens are returned unchanged; a number
+// representable as neither returns the json.Number.Float64 error.
+func normalizeJSONNumber(t interface{}) (interface{}, error) {
+	n, ok := t.(json.Number)
+	if !ok {
+		return t, nil
+	}
+	if i, err := n.Int64(); err == nil {
+		return int(i), nil
+	}
+	f, err := n.Float64()
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
 }
 
 var errEndOfObject = fmt.Errorf("End of object")
@@ -163,7 +186,11 @@ func (store Store) treeItemFromJSONDecoder(dec *json.Decoder) (sops.TreeItem, er
 			item.Value = v
 		}
 	} else {
-		item.Value = value
+		v, err := normalizeJSONNumber(value)
+		if err != nil {
+			return item, err
+		}
+		item.Value = v
 	}
 	return item, nil
 
@@ -252,6 +279,10 @@ func (store Store) jsonFromTreeBranch(branch sops.TreeBranch) ([]byte, error) {
 
 func (store Store) treeBranchFromJSON(in []byte) (sops.TreeBranch, error) {
 	dec := json.NewDecoder(bytes.NewReader(in))
+	// Decode numbers as json.Number instead of the default float64, then
+	// normalize each to int/float64 (see normalizeJSONNumber). The default
+	// float64 silently loses precision for integers larger than 2^53.
+	dec.UseNumber()
 	value, err := dec.Token()
 	if err != nil {
 		return nil, err
@@ -261,7 +292,11 @@ func (store Store) treeBranchFromJSON(in []byte) (sops.TreeBranch, error) {
 			return nil, fmt.Errorf("SOPS only supports JSON files with a top-level object (starting with '{'), not arrays or other types. Got delimiter %s instead. To encrypt this file, wrap it in an object, e.g., {\"data\": [...]}", value)
 		}
 	} else {
-		return nil, fmt.Errorf("SOPS only supports JSON files with a top-level object (starting with '{'), not other JSON types. Got %#v of type %T instead", value, value)
+		v, nerr := normalizeJSONNumber(value)
+		if nerr != nil {
+			v = value
+		}
+		return nil, fmt.Errorf("SOPS only supports JSON files with a top-level object (starting with '{'), not other JSON types. Got %#v of type %T instead", v, v)
 	}
 	return store.treeBranchFromJSONDecoder(dec)
 }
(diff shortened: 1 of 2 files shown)


felixfontein:
I don't think this PR works, as you can use different stores for reading and writing, and no other store than JSON can handle `json.Number` values. Also can these actually be encrypted?

felixfontein:
I guess the answer is no, as several JSON related tests failed.

s3onghyun:
Good catch, thanks @felixfontein — you're right on both counts. `json.Number` leaks into the store-agnostic tree (so reading with JSON and writing with another store breaks) and the AES cipher can't encrypt it (it only handles string/int/float64/bool/time/comment), which is what the failing JSON tests showed.

I've reworked it: decode with `UseNumber` but immediately normalize each number to a concrete type the whole codebase already handles — `int` for integers and `float64` otherwise. This:

- keeps the tree free of `json.Number`, so cross-store read/write works;
- is encryptable (`int`/`float64` are handled by the cipher);
- matches how the YAML store already decodes integers (as `int`), so JSON and YAML are now consistent;
- still fixes the original precision loss (e.g. `1234567890123456789` no longer becomes `...800`).

I added an encrypt → decrypt round-trip assertion to the regression test to prove the decoded value is actually encryptable. `go test ./stores/... ./aes/... .` is green. PTAL.

stores/json/store.go
```diff
@@ -123,11 +123,31 @@ func (store Store) sliceFromJSONDecoder(dec *json.Decoder) ([]interface{}, error
...
+	if !ok {
+		return t
+	}
+	if i, err := n.Int64(); err == nil {
+		return int(i)
+	}
+	if f, err := n.Float64(); err == nil {
+		return f
+	}
+	return t
```
felixfontein: I don't think returning the `json.Number` object as a fallback is a good idea. Instead an error should be returned, namely the one from convering it to a `float64`.

s3onghyun:
Good point, thanks — you're right that falling back to the `json.Number` would re-introduce exactly the leak this is meant to avoid. Updated `normalizeJSONNumber` to return `(interface{}, error)` and propagate the `json.Number.Float64()` error when a value can be represented as neither `int` nor `float64`; both call sites now bubble it up. (In practice the decoder only ever yields syntactically-valid numbers so this path is unreachable, but returning the error is the correct, leak-free behavior.) `go test ./stores/... ./aes/... .` stays green.

s3onghyun:
Done. A couple of things I verified while at it, to be explicit about the boundaries:

- **Range**: integers within the int64 range now stay exact (covers all 64-bit signed IDs, snowflakes, etc.). Values *above* int64 fall back to float64 — bounded by the cipher's supported numeric types (`int` and `float64`). That's still strictly better than before, where every integer became a float64 and lost precision above 2^53. I documented this on `normalizeJSONNumber` and added `TestIntegerBoundaries` covering negative, MinInt64, MaxInt64, and zero round-tripping exactly as `int`.
- **Cross-store**: confirmed the decoded `int`/`float64` values emit cleanly through the YAML store (no `json.Number` reaches a non-JSON store), e.g. `1234567890123456789` (int) → YAML `1234567890123456789`.

`go test ./stores/... ./aes/... .` stays green. Open to a different approach for the >int64 range if you'd prefer (the cipher would need to grow a type to do better than float64 there).

s3onghyun:
Rebased onto the latest main and squashed the review iterations into a single commit. Tests stay green (`go test ./stores/... ./aes/... .`). Ready for another look whenever you have time — thanks for the thorough review, it made the change considerably more robust.

stores/json/store_test.go
```diff
@@ -120,19 +121,71 @@ func TestDecodeSimpleJSONObject(t *testing.T) {
...
+	value := branch[0].Value
+	assert.IsType(t, int(0), value)
+	cipher := aes.NewCipher()
+	key := []byte("0123456789abcdef0123456789abcdef")
+	enc, err := cipher.Encrypt(value, key, "")
+	assert.Nil(t, err)
+	assert.Contains(t, enc, "type:int")
+	dec, err := cipher.Decrypt(enc, key, "")
+	assert.Nil(t, err)
+	assert.Equal(t, value, dec)
```
felixfontein: I would remove this test, since it doesn't really test something related to JSON.
```suggestion
```

stores/json/store.go
```diff
@@ -123,11 +123,42 @@ func (store Store) sliceFromJSONDecoder(dec *json.Decoder) ([]interface{}, error
...
 
+// normalizeJSONNumber converts a json.Number scalar (produced because the
+// decoder runs with UseNumber) into a concrete type that every sops store and
+// the AES cipher can handle: an int for integers and a float64 otherwise.
+// Decoding integers as int (rather than the encoding/json default float64)
+// preserves large-integer precision and matches how the YAML store decodes
+// them. The AES cipher only supports int and float64 for numbers, so integers
+// within the int64 range are kept exact while values outside it fall back to
+// float64 — still an improvement over the previous behavior, where every
+// integer became a float64 and lost precision above 2^53.
```
felixfontein: I think this text would be fine for a commit message, but it doesn't really belong in the comment.
```suggestion
```

felixfontein (commented):
Thanks, this is definitely an improvement.

s3onghyun:
Thanks, @felixfontein — both applied:

- Removed the cipher encrypt/decrypt assertions from the test; it now only checks the JSON-store concern (large integers round-trip and decode to a concrete `int`, not a `json.Number`).
- Trimmed the doc comment on `normalizeJSONNumber` to just what it does; the rationale (precision above 2^53, parity with the YAML store) now lives in the commit message instead.

`go test ./stores/json/` passes.

felixfontein (approved):
Thanks a lot!

PR #2230, opened 2026-06-23
Fix issue when changing user in exec subcommands

sabre1041:
Fix exec subcommands (`exec-file` and `exec-env`) to set `HOME`, `USER`, and `LOGNAME` environment variables in the child process when the `--user` flag is used.

## Originating Issue
When using `--user` to run a child process as a different user, `SwitchUser()` changes the UID/GID but leaves the environment variables pointing at the original user. This causes the child process to inherit incorrect values for `HOME`,
  `USER`, and `LOGNAME`, which can break tools that rely on these variables (e.g., resolving `~` or determining the current user).

Example of issue:
```shell
sops exec-env --user=sops test.yml bash
bash: /root/.bashrc: Permission denied
```

  ## Changes
  - Added `UserEnv()` function in `exec_unix.go` that looks up the target user and returns the correct `HOME`, `USER`, and `LOGNAME` environment variables
  - Updated `ExecWithFile()` and `ExecWithEnv()` to capture user environment variables **before** switching the user, then inject them into the child process environment
  - User environment variables are included in both pristine and non-pristine modes when `--user` is set
  - Added a stub `UserEnv()` on Windows (`exec_windows.go`) that fatals, consistent with the existing `SwitchUser()` behavior
  - Added unit tests in `exec_unix_test.go` covering `UserEnv()` output, process environment isolation, and integration with both exec modes in pristine and non-pristine configurations

### DIFF
--- a/cmd/sops/subcommand/exec/exec_windows.go
+++ b/cmd/sops/subcommand/exec/exec_windows.go
@@ -22,6 +22,10 @@ func GetPipe(dir, filename string) (string, error) {
 	return "", fmt.Errorf("fifos are not available on windows")
 }
 
+func UserEnv(username string) []string {
+	return nil
+}
+
 func SwitchUser(username string) {
 	log.Fatal("user switching not available on windows")
 }
--- a/cmd/sops/subcommand/exec/exec.go
+++ b/cmd/sops/subcommand/exec/exec.go
@@ -57,7 +57,9 @@ func GetFile(dir, filename string) (*os.File, error) {
 }
 
 func ExecWithFile(opts ExecOpts) error {
+	var userEnv []string
 	if opts.User != "" {
+		userEnv = UserEnv(opts.User)
 		SwitchUser(opts.User)
 	}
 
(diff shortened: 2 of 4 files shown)


cmd/sops/subcommand/exec/exec_windows.go
```diff
@@ -22,6 +22,11 @@ func GetPipe(dir, filename string) (string, error) {
 	return "", fmt.Errorf("fifos are not available on windows")
 }
 
+func UserEnv(username string) []string {
+	log.Fatal("user switching not available on windows")
```
felixfontein: I'd probably remove the log output, since now the message is printed twice (as both `SwitchUser` and `UserEnv` are called directly after each other).

felixfontein (commented):
Thanks for working on this!

cmd/sops/subcommand/exec/exec_windows.go
sabre1041: removed per suggestion

felixfontein (approved):
Assuming you did run the `root` tests (I don't think they run in CI), this looks good!

PR #2234, opened 2026-06-28
feature: add init command

nooneknowspeter:
## Goal

Generate an example `.sops.yaml` in the current working directly; `pwd` unless specified using the `-d, --dir` flag

## Why

Helps new users get accustomed to the config file; following the docs can be tricky

Reduces time in setting up an initial config

I sort of had an interactive mode implementation;
- ask user how many creation rules
- ask for regex
- ascii checklist of key types
- user enters public keys/fingerprints after confirming selection

I scrapped that

Also had the idea to inputs keys via flags, for example, -pgp "pgp_fingerprint,..." -regex ".yaml$", 
but that would be cumbersome 

I aimed for simplicity and just made it output an example .sops.yaml instead

It is a useless feature especially when users get accustomed to the tool and understand it
but I thought it would be nice and I wish I had it when I first discovered the tool and started using it

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake . --impure
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 functional-tests/sops
 vendor/
 profile.out
+.direnv/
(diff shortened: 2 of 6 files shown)


Dexmachi:
smth cool to add to this is probably to make complete inits

for instance, setting age keys and other shticks like that

smth like `sops init --age age1pq1... --age age1pq2... --pgp ... --shamir 3` and setting everything up with `creation_rules.key_groups` (the newer version) already


ALSO it isn't useless, it can be very helpful for many cases, mainly because documentation can be hard to maintain

getting sops to autodocument itself should help a lot

nooneknowspeter:
Hey, hope all is well

Thanks for the feedback

That sounds like the second idea I had

I'll take key groups into consideration especially adding it to the example config file

I'll see what I can do when I get some time

PR #2227, opened 2026-06-20
Fix panic when expecting an encrypted string, but a non-string is encountered

felixfontein:
Also fixes a missing space in an error message, and improves some other places I found while looking at all unconditional string casts.

Fixes #2225.

PR #2245, opened 2026-07-06
fix: include decrypted sequence comments in the MAC (fixes #2243)

### LINKED ISSUE #2243: Regression in 3.13.2: MAC mismatch decrypting files written by ≤3.13.1 that contain a comment on a sequence item
### Summary

Files encrypted by sops **≤ 3.13.1** that contain a **comment attached to a YAML sequence item** fail MAC verification (`MAC mismatch ...`) when decrypted by **3.13.2**. 3.13.1 decrypts the exact same file without error. The data key and all values decrypt correctly — only the *recomputed* MAC differs, so sops rejects the file. This is a backward-incompatible change: files at rest that were fine become undecryptable after upgrading.

### Minimal reproduction

A 3-line YAML file whose only notable feature is a commented-out list item:

```yaml
# repro.yaml
items:
  # - commented@example.com
  - real@example.com
```

```console
$ sops-v3.13.1 --pgp <FP> -e repro.yaml > repro.enc.yaml   # any master key works (PGP/age/KMS)

$ sops-v3.13.1 -d repro.enc.yaml >/dev/null && echo OK
OK

$ sops-v3.13.2 -d repro.enc.yaml
MAC mismatch. File has DD26D0EC6D4AEE75...991FFA, computed 77C004ACBB1096...40469B4
```

The encrypted comment is stored as a sequence element `ENC[...,type:comment]`.

**Controls that still succeed under 3.13.2** (so the trigger is specifically a comment on a *sequence item*):

| file | result under 3.13.2 |
|---|---|
| `items:` + `# - x` + `- real` (comment on a **list item**) | ❌ MAC mismatch |
| `# c` + `foo: bar` (comment on a **mapping**) | ✅ OK |
| `items:` + `- real` (sequence, **no comment**) | ✅ OK |

### Environment

- Broken: sops **3.13.2** (official `sops-v3.13.2.linux.amd64`, and distro builds)
- Works: sops **3.13.1**
- File encrypted with 3.13.1 (`version: 3.13.1` in metadata), YAML store, default `unencrypted_suffix`, no `mac_only_encrypted`.

### Root cause

Regression introduced in **#2227** ("Fix panic when expecting an encrypted string, but a non-string is encountered", commits `4ed1e46bb` / `d1c031eaf`), first released in 3.13.2. The decrypt walk in `sops.go` changed the MAC-inclusion test from inspecting the *decrypted* value's type to inspecting the *input* node:

```diff
 if !tree.Metadata.MACOnlyEncrypted || encrypted {
-    if _, ok := v.(Comment); !ok {
+    if !isComment {
         bytes, err := ToBytes(v)
         // ... add to MAC
     }
 }
```

Under ≤3.13.1, an *encrypted* comment that decrypted successfully became a non-`Comment` value (`v`), so `v.(Comment)` was false and its decrypted bytes **were** folded into the MAC. Under 3.13.2, `isComment` is true for that same node, so it is **excluded** from the MAC. For sequence-item comments this changes the computed MAC relative to what ≤3.13.1 wrote, so previously-encrypted files no longer verify.

This is the same class of backward-incompatible MAC/serialization regression as #2188 (3.13.0 INI decrypt regression).

### Impact

Any repository with sops-encrypted YAML containing commented-out list entries (a common pattern, e.g. a commented-out allowlist/user entry) becomes undecryptable after upgrading to 3.13.2, on both `sops -d` and any in-process consumer (Terraform/Terragrunt, External Secrets, CI). Downgrading to 3.13.1 restores decryption.

cbcoutinho:
### What

Fixes a regression in 3.13.2 where sops fails to decrypt any YAML/JSON file
containing a **comment inside a sequence**, reporting `MAC mismatch`. This
affects files written by ≤3.13.1 **and** files 3.13.2 encrypts itself.

Fixes #2243.

### Root cause

A comment in a sequence is encrypted as an `ENC[...,type:comment]` element and
restored as a `sops.Comment` node when the file is loaded again.

#2227 changed the MAC-inclusion test in `Tree.Decrypt` from inspecting the
**decrypted result** to inspecting the **input node**:

```diff
-if _, ok := v.(Comment); !ok {
+if !isComment {
     bytes, err := ToBytes(v)
     hash.Write(bytes)
 }
```

A comment that decrypts successfully becomes a non-`Comment` value, so the old
check included it in the MAC. `Tree.Encrypt` still folds the comment in (at
encrypt time it is a non-`Comment` node), so the encrypt-side and decrypt-side
MACs disagree for any sequence containing a comment.

### Fix

Restore the pre-#2227 behaviour (test the decrypted value's type) while keeping
the non-string panic guard #2227 introduced. One-line semantic change.

### Reproduction

```yaml
# repro.yaml
items:
  # - commented@example.com
  - real@example.com
```

```console
$ sops -v3.13.1 -e repro.yaml > repro.enc.yaml
$ sops -v3.13.1 -d repro.enc.yaml   # OK
$ sops -v3.13.2 -d repro.enc.yaml   # MAC mismatch
```

A comment on a *mapping*, or a sequence with no comment, is unaffected.

### Test

Adds `TestMACWithCommentInSequence` (YAML store round-trip: load → encrypt →
emit → reload → decrypt, asserting the MACs match). It fails on `main` and
passes with this change. `go test ./... ` for the root, stores, decrypt and cmd
packages passes; `gofmt`/`go vet` clean.

---

_This PR was generated with the help of AI, and reviewed by a Human_


stores/yaml/mac_sequence_comment_test.go
felixfontein, line 1: This test is totally unrelated to the YAML store and should be in `sops_test.go` in the repo's root (without using the YAML store at all).

stores/yaml/mac_sequence_comment_test.go
```diff
@@ -0,0 +1,61 @@
...
+// a sops.Comment node. The MAC computed by Tree.Decrypt must include the
+// decrypted comment's value, exactly as Tree.Encrypt included it, otherwise the
+// two MACs disagree and decryption fails with "MAC mismatch" for every file
+// that contains a comment inside a sequence.
+func TestMACWithCommentInSequence(t *testing.T) {
+	store := Store{}
+	cipher := aes.NewCipher()
+	key := bytes.Repeat([]byte{'f'}, 32)
+
+	branches, err := store.LoadPlainFile([]byte("list:\n    # commented\n    - real\n"))
```
felixfontein: Instead of using the YAML store here, directly create `branches` using SOPS objects. There are plenty of examples for this in `sops_test.go`.

stores/yaml/mac_sequence_comment_test.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+	}
+
+	// Encrypt computes the MAC over the plaintext tree (the comment is a
+	// non-Comment node at this point, so it is folded into the MAC).
+	encMAC, err := tree.Encrypt(key, cipher)
+	assert.NoError(t, err)
+
+	// Round-trip through the store so the encrypted comment comes back as a
+	// sops.Comment node, exactly as it would when decrypting a file at rest.
```
felixfontein: Instead of using the YAML store, I would check that the tree has the right format, and decrypt the same tree. I don't see what influence serializing and deserializing the key has on this bug.

sops.go
```diff
@@ -626,8 +626,15 @@ func (tree Tree) Decrypt(key []byte, cipher Cipher) (string, error) {
...
-				// Only add to MAC if not a comment
-				if !isComment {
+				// Only add to MAC if the resulting value is not a comment.
+				// A comment that was successfully decrypted becomes a
+				// non-Comment value here, and it must be added to the MAC
+				// because Encrypt() included it (the plaintext comment is a
+				// non-Comment node at encrypt time). Testing the input's
+				// isComment instead would exclude successfully-decrypted
+				// comments from the MAC, producing a MAC mismatch on files
+				// that contain comments inside sequences (getsops/sops#2243).
```
felixfontein: This isn't true. An encrypted non-`Comment` that was actually a comment (according to the metadata of the encrypted string) becomes a `Comment` after decryption. An encrypted `Comment` that actually was a comment stays a `Comment`.

I'd rewrite this to:
```suggestion
				// Since non-Comment values can be decrypted to Comment
				// values if their metadata says they are comments, we cannot
				// use isComment here. This does happen for comments in slices.
```

felixfontein (commented):
Thanks for reporting this and creating a fix! There are a few issues with the test and the comment for the fix, though.

cbcoutinho:
Thanks for the review @felixfontein! Pushed 1aa891256 addressing all four points:

- **`sops.go` comment** — replaced with your suggested wording. You're right: the plaintext comment in a sequence *is* a `Comment` node, but `Encrypt` turns it into an ordinary (non-`Comment`) string element whose metadata marks it a comment; on `Decrypt` that string decrypts back to a `Comment`. Hence the MAC decision has to look at the decrypted result's type, not `isComment` on the input.
- **Test location** — moved out of the YAML store and into `sops_test.go` (`package sops`).
- **Build tree from SOPS objects** — the test now constructs `branches` directly with a `Comment` inside a slice, no store involved.
- **No serialize/deserialize** — it encrypts and then decrypts the *same* tree (`Encrypt` mutates in place), and additionally asserts the sequence comment became a non-`Comment` string after encryption, so the format is checked explicitly. A small `commentCipher` in the test models the AES cipher's comment-metadata round-trip without pulling in the `aes` package (which would be an import cycle in `package sops`).

Verified the test fails on the pre-fix code and passes with the fix.

sops_test.go
```diff
@@ -56,6 +56,86 @@ func (c encPrefixCipher) Decrypt(value string, key []byte, path string) (plainte
 	return v, nil
 }
 
+// commentCipher models the way the real (AES) cipher records, in the
+// ciphertext's metadata, whether a value was a comment. On Decrypt it uses that
+// metadata to reconstruct a Comment, which is what turns an encrypted
+// non-Comment slice element back into a Comment (see TestMACWithCommentInSequence).
```
felixfontein: Why not simply use the real AES cipher?

sops_test.go
```diff
@@ -56,6 +56,86 @@ func (c encPrefixCipher) Decrypt(value string, key []byte, path string) (plainte
...
+	// Encrypt computes the MAC and mutates the tree in place.
+	encMAC, err := tree.Encrypt(key, cipher)
+	assert.NoError(t, err)
+
+	// The comment inside the sequence must now be an encrypted string element,
+	// exactly as it would be in a file at rest — not a Comment node anymore.
+	seq := tree.Branches[0][0].Value.([]interface{})
+	if _, stillComment := seq[0].(Comment); stillComment {
+		t.Fatalf("expected the sequence comment to be encrypted to a non-Comment string, got %T", seq[0])
+	}
```
felixfontein: Why not use:
```suggestion
	_, stillComment := seq[0].(Comment)
    assert.False(stillComment)
```

sops_test.go
```diff
@@ -56,6 +56,86 @@ func (c encPrefixCipher) Decrypt(value string, key []byte, path string) (plainte
...
+	assert.IsType(t, "", seq[0])
+
+	// Decrypting the very same tree must reproduce the MAC Encrypt computed.
+	decMAC, err := tree.Decrypt(key, cipher)
+	assert.NoError(t, err)
+
+	assert.Equal(t, encMAC, decMAC)
```
felixfontein: I'd add another check here that the decrypted comment is a `Comment` with the right content.

sops.go
felixfontein: You didn't update this text.

cbcoutinho: Done in 23e3ad207 — replaced the block with your suggested wording ("Since non-Comment values can be decrypted to Comment values if their metadata says they are comments, we cannot use isComment here. This does happen for comments in slices.").

sops_test.go
cbcoutinho: Good call — switched to the real `aes.NewCipher()` in 23e3ad207. The reason it wasn't already: this test lives in `package sops`, and `aes` imports the root `sops` package, so importing `aes` from an internal test triggers "import cycle not allowed in test". I moved just this test into a new external test file (`mac_comment_test.go`, `package sops_test`), where importing both `sops` and `aes` is legal, and dropped the stub `commentCipher` entirely.

sops_test.go
cbcoutinho: Done in 23e3ad207 — now `_, stillComment := seq[0].(sops.Comment); assert.False(t, stillComment)`.

sops_test.go
cbcoutinho: Done in 23e3ad207 — after `Decrypt` the test now asserts `seq[0]` is a `sops.Comment` with `Value == "commented"`.

felixfontein (approved):

felixfontein:
@cbcoutinho thanks for fixing this!

PR #2247, opened 2026-07-09
age: optional caching of key command output

oivindoh:
Since sops loads identities once per recipient it attempts, executing `SOPS_AGE_KEY_CMD`/`SOPS_AGE_SSH_PRIVATE_KEY_CMD` each time when set, using slow clients like `bw` is painful when files have many recipients. In my testing, a file with 7 recipients where the matching key is last takes ~20 seconds to finish decrypting.

To fix this, add the opt-in `SOPS_AGE_KEY_CMD_CACHE` to make the command run at most once per process and reuse the result for each recipient. If unset or false, no behaviour changes.

Opt-in because a command may use `SOPS_AGE_RECIPIENT` to return a different key per recipient; caching can break that (a command that exits non-zero for unknown recipients survives, since errors are not cached). The flag is only for commands that ignore the recipient.

- Only successful outputs are cached.
- The lock is not held while the command runs, so concurrent misses behave as uncached.
- No invalidation: a long-lived keyservice serves the cached key until restart. The flag targets short-lived CLI use.

Tests cover once-per-process with the flag, once-per-recipient without, and errors not being cached.

### DIFF
--- a/age/keysource.go
+++ b/age/keysource.go
@@ -10,7 +10,9 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
+	"sync"
 
 	"filippo.io/age"
 	"filippo.io/age/agessh"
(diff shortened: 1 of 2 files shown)


McAndersDK:
Yes please.
our use case is that our command actually return all keys that are defined in a config file, so SOPS have all it need. (atleast if it fail to decrypt, the user cant provide any more private keys.) 
We also have a prompt to approve the fetch of the keys, and right now, if we set the SOPS_AGE_KEY_CMD we are getting prompted for each recipient in the encrypted file.

please @felixfontein @sabre1041

McAndersDK:
How do we get this going?
What are we missing?

PR #2257, opened 2026-07-16
fix(config): normalize paths before regex matching (fixes #2237)

### LINKED ISSUE #2237: creation_rules path_regex not matched against forward-slash paths on Windows
## Summary

On Windows, `creation_rules[].path_regex` is matched against the raw native-separator relative path (containing `\`), so regexes written with `/` — the natural style for cross-platform `.sops.yaml` shared between Windows/macOS/Linux devs — silently fail to match.

## Reproduction

`.sops.yaml`:
```yaml
creation_rules:
  - path_regex: secrets/.*\.env$
    age: age1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

On Windows:
```powershell
sops -e secrets\prod.env
# Error: config file not found, or has no creation rules, and no keys provided through command line
```

Same command on macOS/Linux against `secrets/prod.env` works.

## Root cause

`config/config.go` (v3.12.2):

- Line 582 (`configPathForFile`) strips the config-dir prefix using `filepath.Separator`, so on Windows the remainder keeps `\` separators.
- Line 595 runs `reg.MatchString(filePath)` against that raw native-separator string.
- Same pattern at line 522 for destination rules.

There is no `filepath.ToSlash` / `strings.ReplaceAll(path, "\\", "/")` before the regex match. Consequence: any `path_regex` using `/` as a separator (which is what every doc example and every cross-platform config uses) fails to match on Windows.

Users can work around it with `[/\\]`, but that's non-obvious and pollutes shared `.sops.yaml` files.

## Suggested fix

Normalize to forward slashes before matching, so the regex language stays platform-agnostic:

```go
filePath = filepath.ToSlash(filePath)
if reg.MatchString(filePath) { ... }
```

Applied at both call sites (`config.go:522` and `config.go:595`).

This is a one-line semantics change but strictly widens what matches on Windows — regexes written with `[/\\]` still match, regexes written with `/` start matching. No breakage for existing Linux/macOS users.

## Environment

- sops v3.12.2
- Windows 11
- PowerShell 7 and Git Bash both reproduce

tianrking:
## Summary

Fixes #2237. On Windows, creation and destination rule regexes were evaluated against native `\\`-separated paths, so shared `.sops.yaml` rules written with `/` did not match. Normalize the path only at each regex-matching boundary with `filepath.ToSlash`; file-system paths and other configuration behavior remain unchanged.

Adds a Windows-only regression test that uses native `filepath.Join` paths and verifies both creation rules and destination rules match `^secrets/.*\\.env$`.

## Validation

- `go test ./config -run '^TestPathRegexMatchesWindowsPaths$' -count=1`
- `go vet ./config`
- `go build ./cmd/sops`
- `git diff --check`

`go test ./...` is not clean in this local Windows environment because existing age/PGP tests observe preconfigured `SOPS_AGE_KEY`/SSH identities and incompatible GnuPG permissions; it also includes an existing Windows failure in `TestLoadConfigFileWithAmbiguousPath`. The new targeted regression test passes.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -510,6 +510,7 @@ func configFromRule(rule *creationRule, kmsEncryptionContext map[string]*string)
 func parseDestinationRuleForFile(conf *configFile, filePath string, kmsEncryptionContext map[string]*string) (*Config, error) {
 	var rule *creationRule
 	var dRule *destinationRule
+	filePath = filepath.ToSlash(filePath)
 
 	if len(conf.DestinationRules) > 0 {
 		for _, r := range conf.DestinationRules {
@@ -580,6 +581,7 @@ func parseCreationRuleForFile(conf *configFile, confPath, filePath string, kmsEn
 
 	// compare file path relative to path of config file
 	filePath = strings.TrimPrefix(filePath, configDir+string(filepath.Separator))
+	filePath = filepath.ToSlash(filePath)
 
 	var rule *creationRule
 
--- a/config/config_windows_test.go
+++ b/config/config_windows_test.go
@@ -1,0 +1,37 @@
+//go:build windows
+
+package config
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPathRegexMatchesWindowsPaths(t *testing.T) {
+	conf := parseConfigFile([]byte(`
+creation_rules:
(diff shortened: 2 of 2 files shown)


felixfontein (changes requested):
This will not get merged for the same reason as #1523.

PR #2282, opened 2026-08-21
comment_encryption: decouple comment encryption from value encryption

AnuchitO:
### What

opt-in `comment_encryption: plaintext|encrypted`. controls comment encryption independent of value encryption. today comment encryption is always tied to whichever of the 6 existing selectors is active — no way to keep comments (e.g. inline docs like `# required | default: db`) readable while the value under them is encrypted, or vice versa.

### How

- `Metadata.CommentEncryption` + one guard in `shouldBeEncrypted`, comment nodes only, when set. value-node logic and the unset/default case untouched (backward compat)
- combinable w/ the 4 value-only selectors (`encrypted_regex`/`unencrypted_regex`/`encrypted_suffix`/`unencrypted_suffix`); rejected w/ `encrypted_comment_regex`/`unencrypted_comment_regex` (both already decide comment encryption — combining is ambiguous). existing 6-selector mutual exclusion untouched
- wired: `config.go` (`.sops.yaml`), `stores.go` (persisted in the file's own `sops` metadata, for correct decrypt/rotate), `cmd/sops` (`--comment-encryption` flag, ignored w/ warning on single-value stores like binary)
- shared validation in `sops.ValidateCommentEncryption`, used by all 3 call sites instead of triplicating it

### Tests

- all 4 value/comment combinations, backward-compat guard (unset ⇒ byte-identical to today), config + file-metadata round-trip/validation, full 6-selector compatibility matrix
- e2e verified w/ real `age` keys across dotenv/YAML/INI/JSON/binary

### Note

found an unrelated pre-existing bug while testing (MAC mismatch decrypting a YAML/JSON sequence when `encrypted_comment_regex` actually encrypts a sequence comment — reproduces on `main` w/ zero involvement of this PR). filed separately as getsops/sops#2281, not touched here — flagging for visibility since it's in the same area of `sops.go`.

`comment_encryption` naming/values open to feedback.

### DIFF
--- a/cmd/sops/encrypt.go
+++ b/cmd/sops/encrypt.go
@@ -22,6 +22,7 @@ type encryptConfig struct {
 	EncryptedRegex          string
 	UnencryptedCommentRegex string
 	EncryptedCommentRegex   string
+	CommentEncryption       string
 	MACOnlyEncrypted        bool
 	KeyGroups               []sops.KeyGroup
 	GroupThreshold          int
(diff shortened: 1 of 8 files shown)


AnuchitO:
Closing to review before this goes to maintainers — will reopen or resubmit once I've gone through it myself.

Issue #2281, opened 2026-08-21
MAC mismatch decrypting a YAML/JSON sequence when encrypted_comment_regex actually encrypts a sequence comment


AnuchitO:
### Summary

`sops -d` fails with `MAC mismatch` decrypting a file where `encrypted_comment_regex` causes a comment inside a YAML/JSON sequence to actually get encrypted. Encrypt succeeds; decrypt of that same file always fails, deterministically — not a corruption/tampering scenario. No workaround found besides avoiding `encrypted_comment_regex`/`unencrypted_comment_regex` on a sequence.

Reproduces on current `main` (`e3f6086d`) / released `v3.13.3` — i.e. after #2245, so a different bug from #2243 (different trigger, same "MAC mismatch on a sequence comment" symptom).

### Minimal reproduction

Uses the repo's own test PGP key (`pgp/sops_functional_tests_key.asc`, imported by `make test`); any master key type triggers it equally (also reproduced with `age`).

```console
$ cat repro.yaml
items:
  # sops:enc
  # trigger
  - real

$ sops --pgp FBC7B9E2A4F9289AC0C1D4843D16CEE4A27381B4 --encrypted-comment-regex "sops:enc" -e repro.yaml > repro.enc.yaml
$? = 0

$ cat repro.enc.yaml
items:
    # sops:enc
    - ENC[AES256_GCM,data:s9IQR7yTK7I=,iv:f6uaAV8hHMLyPWBMlsv6m8n/HPg/JSKp5mymXYwBeEE=,tag:KMU3RXrD8pqrhLHkbHg8jw==,type:comment]
    - ENC[AES256_GCM,data:2QF+FQ==,iv:ddal9OOAyGXcGi9UB94pKo6OWZPlEmMFjFnML3D8SK4=,tag:e+Ibo0GgJ7o6sqCMKRvONA==,type:str]
sops:
    encrypted_comment_regex: sops:enc
    ...
    version: 3.13.3

$ sops -d repro.enc.yaml
MAC mismatch. File has 310070AB4092002AA33F550EB98680FA6662E7305FCE227843430ABA334FDFAD1FE0DD4A6484264B59F9B02A780433A552CE3D5C3DA643A77427463C06E47800, computed 8165B752245D56DE98621F3955061A191FA639EBF478EF872A919EAE842996C10824EB9A7878C26C57C7F331CD0FE14E03F35BC0E744CF45CFFA00C7C077CDE2
$? = 51
```

[...]

PR #2283, opened 2026-08-23
Add --(un)encrypted-comment-regex options to 'sops encrypt'

felixfontein:
While investigating #2281 I noticed that `sops encrypt` does not accept `--encrypted-comment-regex` and --unencrypted-comment-regex`, while `sops --encrypt` does. This makes `sops encrypt` also accept these options.


sabre1041 (approved):
LGTM

felixfontein:

PR #2284, opened 2026-08-23
Fix handling of comments in walkSlice()

### LINKED ISSUE #2281: MAC mismatch decrypting a YAML/JSON sequence when encrypted_comment_regex actually encrypts a sequence comment

felixfontein:
Fixes #2281.

### DIFF
--- a/functional-tests/src/lib.rs
+++ b/functional-tests/src/lib.rs
@@ -1738,4 +1738,47 @@ bar: |-
             "Unexpected decrypted content"
         );
     }
+
+    #[test]
+    fn test_slice_comment_round_trip() {
+        let file_path = prepare_temp_file(
+            "test_slice_comment_round_trip.yaml",
+            b"items:
+  # sops:enc
+  # trigger
+  - real",
+        );
+        assert!(
+            Command::new(SOPS_BINARY_PATH)
+                .arg("encrypt")
+                .arg("--encrypted-comment-regex")
+                .arg("sops:enc")
+                .arg("-i")
+                .arg(file_path.clone())
+                .output()
+                .expect("Error running sops")
+                .status
+                .success(),
+            "sops didn't exit successfully"
+        );
+        let output = Command::new(SOPS_BINARY_PATH)
+            .arg("decrypt")
+            .arg(file_path.clone())
+            .output()
+            .expect("Error running sops");
+        println!(
+            "stdout: {}, stderr: {}",
+            String::from_utf8_lossy(&output.stdout),
+            String::from_utf8_lossy(&output.stderr)
+        );
+        assert!(output.status.success(), "sops didn't exit successfully");
+        let data = &String::from_utf8_lossy(&output.stdout);
+        assert!(
+            data == "items:
+    # sops:enc
+    # trigger
+    - real
+"
+        );
+    }
 }
--- a/sops.go
+++ b/sops.go
@@ -380,6 +380,14 @@ func (branch TreeBranch) walkSlice(in []interface{}, path []string, commentsStac
 			return nil, err
 		}
 		in[i] = newV
+		// When encrypting comments in a slice, the Comment is converted to a non-Comment.
+		// When decrypting that non-Comment, the result is once again a Comment.
+		// Therefore, to figure out whether we clear the comment stack's top,
+		// we need to check whether both v and newV are not comments. Since we already checked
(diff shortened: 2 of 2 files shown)


sabre1041 (approved):
LGTM. Tested and verified MAC error no longer occurs

felixfontein:

PR #2285, opened 2026-08-23
feat: Added a Makefile target for producing a binary

sabre1041:
Currently, there is no way to produce a `sops` binary without executing the testing suites. This enhancement enables producing a `sops` binary without running the testing suites

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,9 @@ PROJECT_DIR         := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
 BIN_DIR             := $(PROJECT_DIR)/bin
 
 GO                  := GOPROXY=https://proxy.golang.org go
+GOPATH				:=$(shell go env GOPATH)
 GO_TEST_FLAGS       ?= -race -coverprofile=profile.out -covermode=atomic
+GO_BIN_DIR			:= $(GOPATH)/bin
 
 GITHUB_REPOSITORY   ?= github.com/getsops/sops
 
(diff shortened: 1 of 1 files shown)


felixfontein:
Right now, you can simply run `make install` and then run `~/go/bin/sops`. That's how I test things :)

felixfontein (approved):

sabre1041:
> Right now, you can simply run `make install` and then run `~/go/bin/sops`. That's how I test things :)

I updated the `install` target to first use the new `build` target and then copy the resulting binary to the go `bin` directory

Makefile
```diff
@@ -7,7 +7,9 @@ PROJECT_DIR         := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
 BIN_DIR             := $(PROJECT_DIR)/bin
 
 GO                  := GOPROXY=https://proxy.golang.org go
+GOPATH				:=$(shell go env GOPATH)
```
felixfontein, line 10: Shouldn't there be a space?
```suggestion
GOPATH				:= $(shell go env GOPATH)
```

Makefile
```diff
@@ -37,9 +39,14 @@ all: test vet generate install functional-tests
 .PHONY: origin-build
 origin-build: test vet generate install functional-tests-all
 
+.PHONY: build
+build:
+	$(GO) build -o $(BIN_DIR)/sops github.com/getsops/sops/v3/cmd/sops
+
 .PHONY: install
-install:
-	$(GO) install github.com/getsops/sops/v3/cmd/sops
```
felixfontein, line 42: Does `go install` do anything else than `go build` + copy the executable to `$(GOPATH)/bin`? I think not, but I don't know how `go install` works in detail.

felixfontein (approved):

Makefile
sabre1041: > 

thats basically what it does

PR #2301, opened 2026-09-18
feat: XML support

angyaljanos:
Adding support for  XML file format. Either full file can be encrypted or just a part of it with the comment. Please let me know if the tests arent enough.


angyaljanos:
@felixfontein can you review it?

